PR #17, opened 2017-12-22
CT-1423 Non-offender SAR  fields

stephenrichards:
This PR adds three field used for Non-offender SAR cases, and validates that they are present for SAR cases, and absent for all other case types.


app/models/case/sar/base_sar.rb
```diff
@@ -1,4 +1,6 @@
-class Case::SAR < Case
+class Case::SAR::BaseSAR < Case
```
misaka: Why not just `Case::SAR::Base` here?

app/models/case/sar/base_sar.rb
```diff
@@ -1,4 +1,6 @@
...
+
+  # self.abstract_class = true
```
misaka: Can get rid of this.

misaka (changes requested):

app/models/case/sar/base_sar.rb
stephenrichards: To be more specific.  If we go with just Base, then we're going to end up with a base Base case, Base SAR case, and and Base who knows what else.  I think it's better to have BaseCase (when we get round to doing that), BaseSAR, etc.

app/models/case/sar/base_sar.rb
stephenrichards: NO - actually, it should go back in.

stephenrichards: Ah for some reason it doesn't work when you take have it in. Feels like we should have some check to make sure it's never saved.  I'll think about that later.

app/models/case/sar/base_sar.rb
misaka: What is the problem with having more than one `Base` class? Rails has something like 12 of them (http://api.rubyonrails.org/?q=Base) so it's a common pattern and they must be able to use them without conflict there.

misaka: If anything, I'm more worried about the fact that `Case::FOI` is a class with the class `Case::FOI::ComplianceReview` in it ... that seems to be an anti-pattern to me although I'm not sure what the practical effects should be.

I think we should be aiming for this:

```
module Case
  class Base; end
  module FOI
    class Base < Case::Base; end
    class Standard < Case::FOI::Base; end
    class ComplianceReview < Case::FOI::Base; end
    class TimelinessReview < Case::FOI::Base; end
  end
  module SAR
    class Base < Case::Base; end
    class NonOffender < Case::SAR::Base; end
    class Offender < Case::SAR::Base; end
  end
end
```

misaka (approved):

PR #18, opened 2017-12-28
Ct 1455 dump safeguards

beccapearce:


lib/db/database_dumper.rb
```diff
@@ -34,6 +35,30 @@ def run
 
   private
 
+  def safeguard
+    puts 'Are you sure you need to do this data dump?'.yellow
+    puts ''
+    puts 'Could you solve the issue with any other option such as:'
```
misaka: What do you think about a little more pestering, something like this (in order of easiest to try to hardest):

```
  pester_user('is the issue covered with existing feature tests?', 'no')
  pester_user('can you track problem through Kibana?', 'no')
  pester_user('can you recreate the problem locally?', 'no')
  pester_user('have you tried to recreate the problem on staging with an anonymised dump?', 'yes')

...

def pester_user(query, expected_result)
  response = Readline.readline(query.green)
  unless expected_result.downcase.start_with? response.downcase.strip
    puts "some inspirational message"
    exit
  end
end
```

misaka (approved):

misaka (approved):

PR #21, opened 2018-01-03
CT-1459 Refactor Case to Case::Base

stephenrichards:


db/migrate/20171230113732_remove_default_type.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveDefaultType < ActiveRecord::Migration[5.0]
+  def change
+    change_column_default(:cases, :type, 'Case::FOI')
```
misaka: This may be a convenience for now, but I think we should probably remove this soonish.

misaka: Actually, we don't even need this do we?

misaka (changes requested):
Remove default value for of cases.type.

beccapearce (approved):

PR #50, opened 2018-01-25
CT-1525 Add Kilo Map report

stephenrichards:


beccapearce (approved):

misaka (approved):
A part of me died when I understood the format we of the spreadsheet we need to generate. :(

PR #54, opened 2018-01-29
CT-1391 log new sar case

misaka:
#53 Add SAR case creation:

### Move some controller actions around:

- `/cases/select_type`  to `/cases/new` : This page is now explicitly about choosing the type of correspondence to create a case for.
- `/cases/new` to `/cases/new/foi` : Dedicated page for creating FOI cases with correspondence type retrieved from URL path
- New SAR correspondence type gets the path `/cases/new/sar`

### Move new pages to subdirs

Pages for new FOI cases are moved into `app/views/cases/new/foi`, and `app/views/cases/new/sar` for SAR. These pages are different enough that they benefit from having their own structure until we parameterised them.

### New case page for SAR cases

With all the changes described in ticket [CT-1391](https://dsdmoj.atlassian.net/browse/CT-1391) and updates to the model, including:

- move more attributes from `Case::Base` into `Case::FOI::Standard` as more of them have been deemed unnecessary for SAR cases (e.g. delivery_method, requester_type)

### Work still needing to be done

- _DONE_ ~Hide/show on the new FOI and SAR pages is broken, presumably with the merging in of the GovUK Form Builder update. That change is meant to be backwards compatible, but doesn't seem to work, and I've even tried adding the nesting the content panel to hide/show and specifying the panel_id.~ 
- Fix admin page for SAR: the fields for SAR have changed significantly enough that more work is needed, it may be worth waiting for case attribute parameterisation though (defining the case attributes in the DB with automatically generated new/view/edit pages).
- Fix PaperTrail. We've moved more attributes into the JSON metadata but PaperTrail doesn't seem to track these correctly (previous versions seem to always get the current value). ~This may also be best waiting for attribute parameterisation.~ -- _Unfortunately using a cases properties table won't fix PaperTrail, as it doesn't fully support associations (yet)._
- Sorting out view and edit for SAR cases properly. There should be tickets for this, but it may also benefit from waiting for parameterised attributes.


lib/tasks/sar.rake
```diff
@@ -3,7 +3,7 @@ namespace :sar do
   task :create => :environment do
     category = get_sar_category
     Timecop.freeze Time.now - 24.hours do
-      kase = Case::SAR::NonOffender.new
+      kase = Case::SAR.new
       kase.name = Faker::Name.name
       kase.email = Faker::Internet.email(kase.name)
       kase.subject = Faker::Company.catch_phrase
```
stephenrichards: We don't need sar.rake anymore now we can create sar cases through the interface.

spec/features/cases/foi/case_approvals/disclosure_specialist_approval_spec.rb
```diff
@@ -72,19 +73,19 @@ def undo_de_escalate_case_as_disclosure_specialist(kase, case_list_item)
...
   end
 
-  def create_flagged_case_and_assign_to_team(days_back)
-    login_as manager
-    cases_page.load
-    cases_page.new_case_button.click
+  # def create_flagged_case_and_assign_to_team(days_back)
+  #   login_as manager
+  #   cases_page.load
+  #   cases_page.new_case_button.click
```
stephenrichards: let's remove if elsewhere now.

stephenrichards (changes requested):

lib/tasks/sar.rake
misaka: Hrm. We don't have the admin pages working yet, maybe this is worth keeping until that's done?

app/controllers/cases_controller.rb
```diff
@@ -57,7 +57,9 @@ def index
                .page(params[:page])
                .decorate
     @current_tab_name = 'all_cases'
-    @can_add_case = policy(Case::Base.new(category: Category.foi)).can_add_case?
+
+    # TODO: This needs to be true for any type of case, really.
+    @can_add_case = policy(Case::FOI::Standard).can_add_case?
```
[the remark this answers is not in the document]
misaka: Yes! Good spot, I'll change that here and in the other actions that do similar below.

spec/features/cases/foi/case_approvals/disclosure_specialist_approval_spec.rb
misaka: ✓

misaka: I've switched these features to use the feature steps that are more generally available.

stephenrichards (approved):

PR #58, opened 2018-02-01
CT-1511 Link SAR cases

stephenrichards:
SAR cases may only be linked by the manager of that case, and only to other
SAR cases


app/views/cases/show.html.slim
```diff
@@ -47,7 +47,7 @@ div id="case-#{@case.id}"
     = render partial: 'cases/case_status', locals: {case_details: @case}
 
   section.case-info
-    = render partial: 'cases/linked_cases', locals: { case_details: @case }
+    = render partial: 'cases/linked_cases', locals: { case_details: @case } if current_user.can_link_case?(@case)
```
misaka: I think you should be using the existing Pundit policy `new_case_link?` here to decide whether to show this button or not, and have the logic in there and not in `User.`

app/services/case_linking_service.rb
```diff
@@ -37,24 +37,42 @@ def call
...
 
   def validate_params
+    validate_case_number_present
+    validate_linked_case_not_this_case
+    validate_linked_case_exists
+    validate_sar_case_linking
+  end
+
+  def validate_sar_case_linking
+    if @case.is_a?(Case::SAR)
```
misaka: I think we should be doing this check in the `SARPolicy`, otherwise we'll have 3 places to configure this: `BasePolicy`, `User` and here.

app/models/user.rb
```diff
@@ -112,4 +112,9 @@ def has_live_cases?
   def multiple_team_member?
     team_roles.size > 1
   end
+
+  def can_link_case?(kase)
+    return true unless kase.is_a?(Case::SAR)
```
misaka: I think we should be doing this check in the `SARPolicy`, otherwise we'll have 3 places to configure this: `BasePolicy`, `User` and here.

misaka (changes requested):

app/models/user.rb
stephenrichards: OK, fair enough.

app/services/case_linking_service.rb
stephenrichards: Not sure what you mean here.

app/views/cases/show.html.slim
```diff
@@ -47,7 +47,7 @@ div id="case-#{@case.id}"
     = render partial: 'cases/case_status', locals: {case_details: @case}
 
   section.case-info
-    = render partial: 'cases/linked_cases', locals: { case_details: @case }
+    = render partial: 'cases/linked_cases', locals: { case_details: @case, permitted_events: @permitted_events }
```
misaka: Don't need `permitted_events` as checking the policy is handled inside the [helper used](https://github.com/ministryofjustice/correspondence_tool_staff/blob/d2d7f53d40168042daad79656662d1aa7e2e2c97/app/helpers/cases_helper.rb#L145) by `_linked_cases` partial.

misaka (changes requested):
Extraneous `permitted_events` to be removed from `app/views/cases/show.html.slim`.

misaka (approved):

PR #77, opened 2018-02-14
CT-1563 Enable de-activation of Business Unit

stephenrichards:
This PR enables the deactivation of Business Units which have no active users
and no open cases (but may have closed cases).

### DIFF
--- a/app/views/teams/_business_unit_detail.html.slim
+++ b/app/views/teams/_business_unit_detail.html.slim
@@ -66,7 +66,7 @@ p
                 class: 'button')
 hr
 
-- if @team.active?
+- if Pundit.policy(current_user, team).destroy?
   = link_to "Deactivate team", team_path(@team.id),
                        data: {:confirm => t('.destroy')},
                        method: :delete,
--- a/app/views/teams/_business_unit_detail.html.slim
+++ b/app/views/teams/_business_unit_detail.html.slim
@@ -64,3 +64,10 @@ p
                 new_team_user_path(team_id: team.id, role: team.role),
                 id: 'action--new-responder-user',
                 class: 'button')
+hr
+
+- if @team.active?
+  = link_to "Deactivate team", team_path(@team.id),
+                       data: {:confirm => t('.destroy')},
+                       method: :delete,
+                       id: 'deactivate-team-link'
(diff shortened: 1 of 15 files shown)


app/views/teams/_business_unit_detail.html.slim
```diff
@@ -64,3 +64,10 @@ p
                 new_team_user_path(team_id: team.id, role: team.role),
                 id: 'action--new-responder-user',
                 class: 'button')
+hr
+
+- if @team.active?
+  = link_to "Deactivate team", team_path(@team.id),
```
misaka: This should only show if the user is permitted to `destroy` the team via the `TeamPolicy`.

spec/controllers/teams_controller/destroy_spec.rb
```diff
@@ -62,12 +62,8 @@
           delete :destroy, params: params
         end
 
-        it 'displays a flash alert' do
-          expect(flash[:alert]).to eq "This Directorate still has active business units. Please deactivate all business units of the Directorate before deactivating the Directorate"
```
misaka: Doesn't `TeamsController#deletion_validation` set the flash message "Unable to delete team: team still has active children"? We should expect to see that.

misaka (changes requested):

spec/controllers/teams_controller/destroy_spec.rb
stephenrichards: Deactivation of Business Groups and Directorates is not implements and not part of this ticket.  A lot of the work has been done, but needs to be refactored so that checking the validity of deleting the record is an ActiveRecord validation, and the error message is displayed as a validation error, not a flash message.  This can all be done and tested in the ticket to deactivate business groups and directorates.

misaka: Ok. Feels like we should protect against deactivation of Business Groups or Directorates in the policy if we don't want to allow it, not just rely on the button not showing, which is what it looks like now.

misaka (approved):
For consideration: maybe add a check that the team being destroyed is a Business Unit in `TeamPolicy#destroy?`

PR #83, opened 2018-02-20
CT-1307 Allow users to remove links between cases

aliuk2012:
- the "Remove link" on the case detail page
- the remove_linked_case method to Case::Base model
- The controller action to remove the link.
- add destroy method to case linking service
- add new event to the state machine.
- trigger the event from CaseLinkingService#destroy

### DIFF
--- a/app/services/case_linking_service.rb
+++ b/app/services/case_linking_service.rb
@@ -34,6 +34,34 @@ class CaseLinkingService
     @result = :error
   end
 
+  def destroy
+    ActiveRecord::Base.transaction do
+      # find the linked case
+      @link_case = Case::Base.where(number: @link_case_number).first
+
+      # Destroy the links
+      @case.remove_linked_case(@link_case)
+
+      # Log event in both cases
+      @case.state_machine.remove_linked_case!(acting_user: @user,
+                                       acting_team: @user.teams_for_case(@case).first,
+                                       linked_case_id: @link_case.id)
+      @link_case.state_machine.remove_linked_case!(acting_user: @user,
+                                            acting_team: @user.teams_for_case(@case).first,
+                                            linked_case_id: @case.id)
+
+      @result = :ok
+
+    end
+
+    @result
+  rescue => err
+    Rails.logger.error err.to_s
+    Rails.logger.error err.backtrace.join("\n\t")
+    @error = err
+    @result = :error
+  end
+
   private
 
   def validate_params
--- a/app/services/case_linking_service.rb
+++ b/app/services/case_linking_service.rb
@@ -37,7 +37,7 @@ class CaseLinkingService
   def destroy
     ActiveRecord::Base.transaction do
       # find the linked case
-      @link_case = Case::Base.where(number: @link_case_number).first
+      @link_case = @case.linked_cases.find_by(number: @link_case_number).first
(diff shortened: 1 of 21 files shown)


app/services/case_linking_service.rb
```diff
@@ -34,6 +34,34 @@ def call
     @result = :error
   end
 
+  def destroy
+    ActiveRecord::Base.transaction do
+      # find the linked case
+      @link_case = Case::Base.where(number: @link_case_number).first
```
misaka: I think `@case.linked_cases.find_by(number: @link_case_number).first` would make more sense and be safer here.

spec/controllers/cases_controller/show_spec.rb
```diff
@@ -194,7 +195,7 @@
         let(:user) { another_responder }
 
         it 'permitted_events to be empty' do
```
misaka: Can we change the spec description to match the actual expectation?

spec/controllers/cases_controller/show_spec.rb
```diff
@@ -260,7 +261,7 @@
         let(:user) { create(:responder) }
 
         it 'permitted_events to be empty' do
```
misaka: Can we change the spec description to match the actual expectation?

spec/controllers/cases_controller/show_spec.rb
```diff
@@ -323,7 +324,7 @@
...
```
misaka: Can we change the spec description to match the actual expectation?

spec/controllers/cases_controller/show_spec.rb
```diff
@@ -371,7 +372,7 @@
         let(:user) { responder }
 
         it 'permitted_events to be empty' do
```
misaka: Can we change the spec description to match the actual expectation?

spec/controllers/cases_controller/show_spec.rb
```diff
@@ -384,7 +385,7 @@
         let(:user) { create(:responder) }
 
         it 'permitted_events to be empty' do
```
misaka: Can we change the spec description to match the actual expectation?

spec/views/cases/_linked_cases_html_slim_spec.rb
```diff
@@ -51,6 +56,12 @@ def disallow_case_policy(policy_name, kase)
         expect(row.case_type.text).to eq 'FOI '
         expect(row.request.text)
             .to eq "#{ linked_case.subject } #{ linked_case.name }"
+
+        # annoyingly had to add \n below because of how locales using html is handled
```
misaka: Just to mention, you could also do `row.remove_link.text.strip` to remove whitespace.

misaka (changes requested):

app/services/case_linking_service.rb
aliuk2012: 👍

misaka:
As discussed with @aliuk2012, the changes to the spec descriptions are actually highlighting a bigger issue ... that we need to refactor how we present action buttons and links in general. We'll save this for another ticket.

misaka (approved):

PR #85, opened 2018-02-22
CT-1148 passes type to redirect after team update

beccapearce:


app/controllers/teams_controller.rb
```diff
@@ -186,7 +186,7 @@ def post_update_redirect_destination
         areas_covered_by_team_path(@team)
       else
         flash[:notice] = 'Team details updated'
-        params[:team_type] == 'bg' ? teams_path : team_path(@team.parent_id)
+        @team.type == 'BusinessGroup' ? teams_path : team_path(@team.parent_id)
```
misaka: Is there a reason not to use `is_a?` as on line 185?

misaka (approved):

aliuk2012:
Shouldn't we have a test for this?

misaka (approved):

PR #87, opened 2018-02-23
CT-1583 Case closure copy change

stephenrichards:


misaka:
Unlike Al's branch this wasn't a parallel test run, so not sure that the failure isn't a flickering test. Re-running the tests.

Also, they took 49m to run?? Never seen it take so long.

misaka (approved):

aliuk2012 (dismissed):
Why all the copy changes?

aliuk2012 (approved):

PR #97, opened 2018-03-01
CT-1506 changes the validation for user has live cases

aliuk2012:
Currently there is an issue where a user who belongs to multiple teams
(for eg Team A and Team B). They need to be deactivate from Team A and
they have live cases assigned to them from Team B. The system will not
allow them to be deactivated from any of the teams becuase they have
live cases assigned to them.

This commit renames the existing User#has_live_cases? method to
User#has_live_cases_for_team? which allows us to pass in a team and
check if they have live cases for that team. If they don't then we can
safely say they can be deactivated from that team. It they dont have any
teams then their account will be closed


misaka (approved):

misaka:
Might get a conflict from the Stephen's commit ... in which case you may want to reset HEAD^ and rebase from develop once the other one is merged.

PR #118, opened 2018-03-27
CT-1651 add event into state machine config

stephenrichards:


misaka (approved):
I'm guessing that this is the kind of thing that the tests you're working on now would've caught?

PR #121, opened 2018-03-28
Ct 1642 state machine tests

beccapearce:


spec/state_machines/global_state_machine_spec.rb
```diff
@@ -0,0 +1,1387 @@
...
+  before(:all) do
+    DbHousekeeping.clean
+    @setup = StandardSetup.new
+  end
+
+  after(:all) { DbHousekeeping.clean }
+
+  describe :accept_approver_assignment do
+    it {
+      expect(1).to eq 1
```
misaka: Checking that the fundamental laws of computing still hold? ;)

spec/support/standard_setup.rb
```diff
@@ -0,0 +1,161 @@
...
+        responder:                        [ responder_user, responding_team ],
+        another_responder_in_same_team:   [ another_responder_in_same_team_user, responding_team ],
+        another_responder_in_diff_team:   [ another_responder_in_diff_team_user, another_responder_in_diff_team_user.responding_teams.first ],
+        press_officer:                    [ press_officer_user, press_office_team ],
+        private_officer:                  [ private_officer_user, private_office_team ]
+    }
+
+    @cases = {
+       std_unassigned_foi:        create(:case),
+       std_awresp_foi:            create(:assigned_case,
```
misaka: Took me a couple tries to realise that `aw` is `Awaiting` responder. Could we have a legend for what the shorthand means? I'd be happy to see proper shorthand here (anyone remember [Hungarian Notation](https://en.wikipedia.org/wiki/Hungarian_notation)??) so long as it's documented.

spec/support/standard_setup.rb
```diff
@@ -0,0 +1,161 @@
...
+       press_office_team:         find_or_create(:team_press_office),
+       private_office_team:       find_or_create(:team_private_office),
+       another_approving_team:    create(:approving_team)
+    }
+
+
+    @users = {
+      manager_user:                         find_or_create(:disclosure_bmt_user),
+      approver_user:                        find_or_create(:disclosure_specialist),
+      another_approver_user:                create(:approver, approving_team: another_approving_team),
```
misaka: Shouldn't we be testing for approver coworkers too? Similar to the `another_responder_in_same_team_user` below (which might also be expressed as `responder_coworker_user`).

spec/support/standard_setup.rb
```diff
@@ -0,0 +1,161 @@
...
+       responding_team:           create(:responding_team),
+       press_office_team:         find_or_create(:team_press_office),
+       private_office_team:       find_or_create(:team_private_office),
+       another_approving_team:    create(:approving_team)
+    }
+
+
+    @users = {
+      manager_user:                         find_or_create(:disclosure_bmt_user),
+      approver_user:                        find_or_create(:disclosure_specialist),
```
misaka: How about using `disclosure_bmt_user` instead of `manager_user` and `disclosure_specialist` instead of `approver_user` here since we're using what we're calling the "Organisational Context" here?

misaka (changes requested):
Some comments about the `StandardSetup`. Haven't finished reviewing completely but thought I'd publish these comments while I do some more work on deploys.

spec/state_machines/global_state_machine_spec.rb
stephenrichards: yep

spec/support/standard_setup.rb
stephenrichards: have added explanation into standard setup where they are defined

spec/support/standard_setup.rb
stephenrichards: done

misaka (changes requested):
Just looking for clarification on why we have press/private assignments pending.

misaka (approved):

PR #130, opened 2018-04-09
CT-1609 Update Search Index

stephenrichards:
Every time a Case::Base record gets updated, if any of the amended fields those
used in the search index, the dirty flag is set and a SearchIndexUpdater
job is scheduled to run after a delay of 10 seconds.

This also happens if an responding assignment is created or updated.

The search index updater job will collect all records in the database with the
dirty flag set, and call SearchIndex#update_document on them.

The reason for the 10 second delay is to prevent the job running multiple times
on the same record if the same record is updated several times during a transaction
(assignments to responding team will also cause the dirty flag to be set).


db/migrate/20180322183946_create_search_index.rb
```diff
@@ -0,0 +1,22 @@
+class CreateSearchIndex < ActiveRecord::Migration[5.0]
```
misaka: Did you mean to include this as part of this PR?

app/models/search_index.rb
```diff
@@ -0,0 +1,67 @@
+# == Schema Information
```
misaka: Did you mean to include this entire class as part of this PR?

spec/models/assignment_spec.rb
```diff
@@ -214,4 +214,124 @@ def create_pending_responding_assignment(kase, manager)
...
+            expect(kase.reload).to be_clean
+            @assignment.update(state: 'accepted')
+            expect(kase.reload).to be_clean
+          end
+        end
+
+
+        it 'queues the job' do
+          expect {
+            create :assignment, :responding, case_id: kase.id
```
misaka: This should test updating the job, otherwise it's exactly the same as the create test above. And I think we want a spec about queueing a job (or not) in the three contexts above.

spec/models/assignment_spec.rb
```diff
@@ -214,4 +214,124 @@ def create_pending_responding_assignment(kase, manager)
...
+      context 'new assignment' do
+        it 'marks the case as dirty' do
+          create :assignment, :responding, case_id: kase.id
+          expect(kase.reload).to be_dirty
+        end
+
+        it 'queues the job' do
+          expect {
+            create :assignment, :responding, case_id: kase.id
+          }.to have_enqueued_job(SearchIndexUpdaterJob).at_least(1)
```
misaka: Can we check that it enqueues the job with a 10s delay?

spec/models/case/base_spec.rb
```diff
@@ -1452,4 +1453,44 @@ def clear_case(kase, t, new_state, acting_team)
...
+
+
+  describe '#trigger_reindexing' do
+    context 'creating a new record' do
+      it 'sets the dirty flag' do
+        kase = build :case
+        expect(kase).not_to be_dirty
+        kase.save!
+        expect(kase).to be_dirty
+      end
```
misaka: Shouldn't we check that it enqueues a job in all the specs here?

misaka (changes requested):
I think we also need to add this to a run script (we said we'd run this in the reporting container, right?). And also a sidekiq configuration.

app/jobs/search_index_updater_job.rb
```diff
@@ -0,0 +1,13 @@
+class SearchIndexUpdaterJob < ApplicationJob
+
+  queue_as :search_index_updater
+
+  def perform
+    RavenContextProvider.set_context
+    kases = Case::Base.where(dirty: true)
```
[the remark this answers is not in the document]
stephenrichards: I thought you told me that sidekiq took care of not running the same thing twice?  After all, if it didn't, then everyone would be receiving three emails.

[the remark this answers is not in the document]
misaka: Ya, that's true, that should mitigate the issue. But didn't we discuss the possibility that more than one update occurs within a request, close together. For example in `CaseFlagForClearanceService`, we'd end up with multiple jobs on the job queue.

spec/models/assignment_spec.rb
stephenrichards: Done

spec/jobs/search_index_updater_job_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'rails_helper'
+
+describe PdfMakerJob, type: :job do
```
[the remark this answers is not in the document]
stephenrichards: done

app/jobs/search_index_updater_job.rb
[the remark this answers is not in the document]
stephenrichards: done

app/models/assignment.rb
```diff
@@ -55,6 +56,8 @@ class Assignment < ApplicationRecord
 
   attr_accessor :reasons_for_rejection
 
+  before_save :mark_case_as_dirty_for_responding_assignments
```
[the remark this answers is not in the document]
stephenrichards: I originally used before_save here because I wanted access to the fields that had changed, but I see you can get them in after_save as well.  There are issues with after commit - see https://apidock.com/rails/ActiveRecord/Transactions/ClassMethods/after_commit, although these bugs may have been addressed by now.  What do you think the implications of saving a case before the assignment might be?  I am reluctant to change it just because  - well, because what?   Am happy to change if there's a real issue, but not otherwise.

spec/models/assignment_spec.rb
stephenrichards: done

spec/models/assignment_spec.rb
stephenrichards: done

spec/models/case/base_spec.rb
stephenrichards: done

app/models/assignment.rb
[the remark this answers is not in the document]
misaka: It's just a smell for me ... what if something prevents the save from occuring by either stopping the chain or raising? What if we remove the 10 second delay and re-indexing stops working because of the race condition? If you think these concerns are too marginal to worry about fine, let's just roll with it as-is.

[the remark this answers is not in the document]
stephenrichards: Write a ticket to get it converted to after save or after commit if it bothers you that much, and convince Polly why we should spend time on it.  My view is that it doesn't really matter what order they're done in, as long as they all get done, or none get done.  After save has to be used if you're writing a new associated record and need the id of that record,  but in this case it isn't, and it seems to me better to make sure the record is marked as dirty before writiing the assignment.

misaka (approved):

PR #131, opened 2018-04-09
Ct 1641 Update Search index on change of Business Unit name

stephenrichards:


beccapearce (approved):
All looks fine except the extra commit

PR #133, opened 2018-04-11
CT-1611 search design

aliuk2012:
This commit updates the search page design:

- move the search submit button so that it appears next to the search field
- display number of cases found
- update the copy of the search field
- add copy for no results to help users try other search terms
- removes the report columns if not returns are found
- removes the flash message if no results found

Tech tidy up:
- remove unused css classes
- split the existing case index view into a separate search and index view this was done to help simplify the view and remove the logic used between the two views.

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -285,17 +285,14 @@ class CasesController < ApplicationController
 
   def search
     @query = params[:query]
-    @current_tab_name = 'search'
     if @query.present?
       @query.strip!
       @cases = policy_scope(Case::Base).search(@query).page(params[:page]).decorate
       if @cases.empty?
-        flash.now[:alert] = 'No cases found'
+        @cases = nil
       end
-    else
-      @cases = nil
     end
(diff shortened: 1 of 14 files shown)


app/controllers/cases_controller.rb
```diff
@@ -285,17 +285,14 @@ def confirm_respond
...
   def search
     @query = params[:query]
-    @current_tab_name = 'search'
     if @query.present?
       @query.strip!
       @cases = policy_scope(Case::Base).search(@query).page(params[:page]).decorate
       if @cases.empty?
-        flash.now[:alert] = 'No cases found'
+        @cases = nil
       end
```
stephenrichards: I will change this when I merge in my changes, but it's pointless to say 
``` if @casess.empty?
    @cases = nil
```

and then in the view write 

``` if !@cases.nil? ```

better to leave it empty in the controller, and test ```if @cases.any?``` in the view.  Much clearer what is actually going on that way.

stephenrichards (approved):

PR #137, opened 2018-04-12
Ct 1659 new password deactivated users emails

beccapearce:


app/mailers/action_notifications_mailer.rb
```diff
@@ -44,6 +44,18 @@ def ready_for_approver_review(assignment)
     mail(to: recipient.email)
   end
 
+  def account_not_active(record)
```
misaka: It sounds like this would be more appropriate in the `devise_mailer.rb` file ... this `action_notifier_mailer.rb` is more about notifying the user about case-related actions.

spec/controllers/passwords_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    let(:params){
+      {
+        user:{
+          email: deactivated_user.email
+        }
+      }
+    }
+
+    context 'user is deactivated' do
+      it 'redirect and shows message' do
```
misaka: Can we have something simple in here that differentiates it from the context below in terms of expectation? Otherwise it looks like we expect the same thing, so if this one also said "sends email" that'd help.

misaka (changes requested):

misaka (approved):

PR #149, opened 2018-04-20
CT-1550 Fix bug in papertrail versioning

stephenrichards:
Full notes in readme, but there were two issues with regenerating
an old version from a papertrail versions record.

One was thet JSONB fields were not properly recreated - this has been
fixed by writing a custom serializer

The socond was that gov_uk_date fields were not being recreated properly.
The readme gives a work-around tothe this issue.


beccapearce (commented):
The new tests aren't passing

beccapearce (approved):

PR #155, opened 2018-04-26
CT-1207 Status Filter

stephenrichards:


stephenrichards:
This has somehow got the type filter on the open cases page.  Need to take that out. until all the filters are ready to go on to that page.

app/views/cases/index.html.slim
```diff
@@ -8,7 +8,11 @@
     .column-full.button-holder
       = link_to t('common.new_case_button'), new_case_path, class: 'button'
 
-= render partial: 'state_filter_form' unless request.path == '/cases/my_open'
```
aliuk2012: We need to add this back in and remove the commented out code for the time being

aliuk2012 (approved):
Looks good from frontend POV

beccapearce (approved):

PR #156, opened 2018-04-27
CT-1508 search filters

misaka:
This provides the Case Type filter. Feature implemented:

- Case type filter frontend and backend
- Saving filters to the search_queries table
- Re-populating filter forms from saved query data.
- Making `StandardSetup` fixtures available to any tests.


Guardfile
```diff
@@ -10,14 +10,14 @@ rspec_options = {
...
-  watch(%r{(app|vendor)(/assets/\w+/(.+\.(scss|js|html|haml))).*}) { |m| "/assets/#{m[3]}" }
-end
+# guard 'livereload' do
+#   watch(%r{app/views/.+\.(erb|haml|slim)})
+#   watch(%r{app/helpers/.+\.rb})
+#   watch(%r{public/.+\.(css|js|html)})
+#   watch(%r{config/locales/.+\.yml})
+#   # Rails Assets Pipeline
+#   watch(%r{(app|vendor)(/assets/\w+/(.+\.(scss|js|html|haml))).*}) { |m| "/assets/#{m[3]}" }
+# end
```
aliuk2012: Why is this commented out?

spec/state_machines/global_state_machine_spec.rb
```diff
@@ -107,7 +140,7 @@
         [:press_officer, :full_unassigned_foi],
         [:press_officer, :full_awresp_foi],
         [:press_officer, :full_draft_foi],
-      )
+      ).debug
```
beccapearce: Do we need the debugs left in?

spec/state_machines/global_state_machine_spec.rb
```diff
@@ -124,7 +157,7 @@
         [:another_responder_in_same_team, :full_awresp_foi],
         [:another_responder_in_same_team, :trig_awresp_foi_accepted],
         [:another_responder_in_same_team, :full_awresp_foi_accepted]
-      )
+      ).debug
```
beccapearce: another debug

spec/state_machines/global_state_machine_spec.rb
```diff
@@ -304,7 +337,7 @@
         [:private_officer, :full_ppress_foi_accepted],
         [:private_officer, :full_pprivate_foi_accepted],
 
-      )
+      ).debug
```
beccapearce: de 🐛 Bug

Guardfile
misaka: I was getting errors about livereload not being installed or something. The right solution is probably to just get it installed locally, I'll put this back in.

aliuk2012 (changes requested):

spec/state_machines/global_state_machine_spec.rb
misaka: Don't need them and I'll take them out, but generally they shouldn't cause an issue as they don't activate if the file is being run at a terminal

app/views/cases/search_filters/_case_types.html.slim
```diff
@@ -0,0 +1,11 @@
+= form_for @query, url: search_cases_path, method: :get do |f|
+  = hidden_field_tag('search_query[parent_id]', @parent_id)
+  / = hidden_field_tag('search_query[filter_type]', 'case_type')
```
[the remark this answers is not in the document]
misaka: Oh, hrm, I hand't thought of that as a requirement. I don't mind adding that back in so that we don't have to deduce from the params which filter tab was used.

app/decorators/search_query_filter_decorator.rb
```diff
@@ -0,0 +1,35 @@
...
+    inherited_attribute_value(:search_text)
+  end
+
+  def filter_type
+    'case_type'
+  end
+
+  # def sensitivity
+  #   object.query['case_sensitivity']
+  # end
```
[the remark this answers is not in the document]
misaka: This decorator is actually no longer used. File removed.

app/decorators/search_query_filter_decorator.rb
```diff
@@ -0,0 +1,35 @@
...
+  def sensitivity_settings
+    {
+      'non-trigger' => 'Non-trigger',
+      'trigger'     => 'Trigger',
+    }
+  end
+
+  # def type
+  #   object.query['case_type']
+  # end
```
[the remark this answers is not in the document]
misaka: This decorator is actually no longer used. File removed.

app/views/cases/search_filters/_case_types.html.slim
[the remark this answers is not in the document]
stephenrichards: I'm not sure that's necessary.  They get the search results and all the tabs displayed in closed position, no?  So why do we need to know what tab they were last on?

stephenrichards (approved):

app/models/case/base.rb
```diff
@@ -103,10 +103,15 @@ def self.searchable_document_tsvector
...
 
+  # cases that have ever been flagged for approval
   scope :flagged_for_approval, ->(*teams) do
     joins(:assignments)
       .where(assignments: { team_id: teams.map(&:id), role: 'approving' })
   end
+  # cases that are currently considered as trigger
+  scope :trigger, -> { where(workflow: ['trigger', 'full_approval']) }
+  # cases that are NOT currently considered as trigger
+  scope :non_trigger, -> { where.not(id: trigger) }
```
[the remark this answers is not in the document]
misaka: I think this is efficient, or at least sufficiently so. This gets turned into a single SQL statement by the AREL and appears to be optimised by PostgreSQL. On my local database with 2800+ cases it takes 72.9ms to run.

aliuk2012 (approved):

PR #157, opened 2018-04-27
CT-1682 Add for_view_only scope 

stephenrichards:


spec/policies/base_policy_scope_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      end
+    end
+
+    context '#for_view_only' do
+      it 'returns all cases for managers' do
+        manager_scope = described_class.new(@manager, Case::Base.all).for_view_only
+        expect(manager_scope).to match_array(@existing_cases)
+      end
+
+      it 'for responders - returns only cases assigned to their team' do
```
misaka: I think this should should be: `for respondents - returns all cases`.

misaka (approved):
Just wrong wording in a spec, can be fixed later.

PR #182, opened 2018-05-03
CT-1671 reintroduce regex for filtering out business units

aliuk2012:


beccapearce (approved):

stephenrichards (dismissed):
We need to make this work using a search button to the right of the text box where we  enter the text to match the business unit name on. - as per Mia's design.

app/assets/javascripts/modules/AssignedToFilter.js
```diff
@@ -9,29 +9,39 @@ moj.Modules.AssignedToFilter = {
...
-              $elem.find(':checkbox').prop('checked', false);
-          }
-        })
+      .each(function (index, element) {
+        var $elem = $(this);
+        // Remove any character that is not alphanumeric
+        var labelText = $elem.text().replace(/[^a-z0-9+]+/gi, '');
+        // Remove any character that is not alphanumeric
+        var cleanedTerm = searchFor.replace(/[^a-z0-9+]+/gi, '');
+        var check = new RegExp(cleanedTerm, 'gi');
```
misaka: These two lines don't have to happen for every element in `$('.js-all-business-units').find('.multiple-choice')` does it?

app/assets/javascripts/modules/AssignedToFilter.js
```diff
@@ -9,29 +9,39 @@ moj.Modules.AssignedToFilter = {
...
+
+
+        // if checkbox label test contains any part of the search term
+        if (check.test(labelText)) {
+          $elem.show();
+        } else {
+          $elem.hide();
+          $elem.find(':checkbox').prop('checked', false);
```
misaka: Shouldn't all the business units that have been previously selected be replayed?

misaka: As per Slack discussion, we'll leave this for a later improvement and just get this out for now.

misaka (approved):

PR #192, opened 2018-05-09
CT-1689 Add Type filter to All Open Cases page

stephenrichards:


app/models/search_query.rb
```diff
@@ -79,10 +89,33 @@ def update_for_click(position)
...
+
+  private
+
+  def search_results
+    results = Case::BasePolicy::Scope.new(User.find(user_id), Case::Base.all).for_view_only
+    results.search(search_text)
+  end
+
+  def list_results
+    request = OpenStruct.new(path: list_path, params: YAML.load(list_params))
```
misaka: Instantiating the `GlobalNavManager` here doesn't feel right. It's already been instantiated correctly in the controller as `@global_nav_manager` so why don't we pass the result of `current_page_or_tab.cases.by_deadline`  all the way into here? It'd have to go through a few methods: `CasesController#open_cases` -> `CasesController#search_and_filter` -> `CaseSearchService#call` -> `SearchQuery#results` -> `SearchQuery#list_results` but none of these take parameters and it'd make sense to have a base AREL passed through to operate on.

app/models/search_query.rb
```diff
@@ -68,7 +71,14 @@ def update_for_click(position)
     save!
   end
 
+  def self.record_list(user, path, params)
+    self.create!(user_id: user.id,
+                 list_path: path,
+                 list_params: params.to_yaml,
```
misaka: Is there a reason we need YAML here and can't just use a hash (which I _think_ would just be converted to JSON)?

misaka (commented):
I have a couple suggestions which I think: 1. won't be a big change; 2. will help simplify by reducing duplication / unnecessary code. I just want to see if they're viable and worth doing.

app/models/search_query.rb
misaka: I see this is part of a larger possible strategy to move search logic from `GlobalNavManager` to `SearchQuery`. LGTM.

app/models/search_query.rb
misaka: @beccapearce has a feeling that this is required for something so 👍

misaka (approved):

PR #206, opened 2018-05-22
Ct 1394 close SAR as TMM

beccapearce:
Adds an end to end test for SARs and refactoring the closing method


spec/site_prism/support/case_type_element_selector.rb
```diff
@@ -0,0 +1,3 @@
+Capybara.add_selector(:multi_case_id) do
```
misaka: How about `case_form_element`?

misaka (approved):

PR #212, opened 2018-05-29
CT-1736 remove apt-get source "nodesource"

misaka:
The Ruby 2.3 image we use upgraded to Debian stretch which has node built-in so
this is no longer needed.


misaka:
Fixes the Jenkins build: https://ci.service.dsd.io/me/my-views/view/CT/job/ct-staff-build-docker/275/

aliuk2012 (approved):

PR #216, opened 2018-05-30
CT-1729 display clearance info immediately for SAR cases

misaka:
As with FOIs, the clearance info was not being displayed until the escalation
deadline had been met. This change ensures that it's displayed by always
returning false for the "within_esclation_deadline?" predicate.


misaka:
This also addresses CT-1728 using a different implementation.

stephenrichards (approved):

PR #218, opened 2018-05-31
CT-1403 Allow BMT users to edit open SAR cases

stephenrichards:


misaka (approved):
Nice one.

misaka (changes requested):
Actually, I just realised ... no feature tests?

beccapearce (changes requested):
Could I be a pain and get you to change it in the spec/features/user_journeys/sar/non_trigger_spec.rb too, the steps are there but commented out

beccapearce (approved):

misaka (approved):

PR #229, opened 2018-06-05
Ct 1492 sar cts admin

beccapearce:


app/controllers/admin/cases_controller.rb
```diff
@@ -30,55 +34,118 @@ def index
...
-    case_creator = CTS::Cases::Create.new(Rails.logger, case_model: Case::Base)
+    if params[:correspondence_type].present?
+      @correspondence_type = params[:correspondence_type]
+      self.send("prepare_new_#{@correspondence_type}".to_sym)
+    else
+      prepare_select_type
+    end
+  end
+
+  def prepare_new_foi
```
misaka: Should `prepare_new_foi` and `prepare_new_sar` be private?

app/controllers/admin/cases_controller.rb
```diff
@@ -30,55 +34,118 @@ def index
...
+    @selected_state = 'drafting'
+    @s3_direct_post = S3Uploader.s3_direct_post_for_case(@case, 'requests')
+    @case.reply_method = 'send_by_email'
+
     render :new
   end
 
   private
 
+  def prepare_select_type
```
misaka: What is being prepared here?

app/controllers/admin/cases_controller.rb
```diff
@@ -30,55 +34,118 @@ def index
   end
 
   def new
-    case_creator = CTS::Cases::Create.new(Rails.logger, case_model: Case::Base)
+    if params[:correspondence_type].present?
+      @correspondence_type = params[:correspondence_type]
+      self.send("prepare_new_#{@correspondence_type}".to_sym)
```
misaka: Not worth changing now, but just wanted to mention that I think that the `to_sym` here may be superfluous, so would be more succinct to just not bother. Also, it's possible that Ruby internally converts the sym back to a string, I don't know how `send` works.

db/structure.sql
```diff
@@ -1609,6 +1609,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20180508131152'),
 ('20180517140929'),
 ('20180522132456'),
+('20180522132457'),
```
misaka: There is no migration in this branch ... is there a migration missing from `master`?

spec/site_prism/page_objects/pages/admin/cases/foi_page.rb
```diff
@@ -0,0 +1,41 @@
+module PageObjects
+  module Pages
+    module Admin
+      module Cases
+        class FOIPage < PageObjects::Pages::Base
```
misaka: Should this be `NewFOIPage` so that it follows the path (`/admin/cases/new/foi`) and also matches what we have for non-admin FOI creation?

spec/site_prism/page_objects/pages/admin/cases/new_page.rb
```diff
@@ -3,37 +3,27 @@ module Pages
...
-          element :full_name, '#case_name'
-          element :email, '#case_email'
-          element :address, '#case_postal_address'
-          element :type_of_requester, :xpath,
-                  '//fieldset[contains(.,"Type of requester")]'
-          element :subject, '#case_subject'
-          element :full_request, '#case_message'
-          element :received_date, '#case_received_date'
-          element :created_at, '#case_created_at'
+          # element :submit_button, '.button'
```
misaka: Can be removed probably.

misaka (changes requested):
Couple small things like the migration in structure.sql and names of the new SitePrism pages.

app/controllers/admin/cases_controller.rb
beccapearce: probably!

app/controllers/admin/cases_controller.rb
beccapearce: Nothing this pattern was copied from cases_controller (the non-admin one) but happy to change make more sense!

db/structure.sql
beccapearce: Hmmmm no idea, this migration isn't in my branch either 👻spooky!! I'll remove it

spec/site_prism/page_objects/pages/admin/cases/foi_page.rb
```diff
@@ -0,0 +1,41 @@
+module PageObjects
+  module Pages
+    module Admin
+      module Cases
+        class NewFOIPage < PageObjects::Pages::Base
```
misaka: The filename should change for this too.

misaka (approved):

PR #245, opened 2018-06-11
CT-1751 Updates to Create SAR page

stephenrichards:
This PR re-organises some of the questions on the Create/Edit DAR pages
and changes some copy as a result of user research.  It also adds a new
field third_party_relationship which is mandatory if the SAR is raised by someone
other than the subject.

We've also taken the opportunity to properly put labels and error messages in the
translation file, rather than rely on decorator methods, which were not working
correctly for boolean fields on the edit page.


app/models/case/sar.rb
```diff
@@ -64,6 +64,11 @@ def within_escalation_deadline?
     false
   end
 
+  def third_party_display
+    third_party == true ? 'Da' : 'Nyet'
```
beccapearce: Does this need to be in Russian?

beccapearce (changes requested):

beccapearce (approved):

PR #249, opened 2018-06-13
Ct 1752 sar trigger journey

beccapearce:


stephenrichards (changes requested):
There is no validation that either Yes or No is selected in answer to the question "Flag for disclosure specialist?"

Not quite sure how far along the journey is covered with this ticket, but once accepted, there is no further action that the responder can take

stephenrichards (approved):

PR #252, opened 2018-06-15
CT-1564 Implement Business Unit performance report for SARs

stephenrichards:
This is implemented by moving the logic from the existing
R003BusinessUnitPeformanceReport into a new class BaseBusinessUnitPerformanceRepoert
and deriving the Business unit performce reports for both FOIs and SARs from that.  The
derived classes just differ in returning different names nad desciptions of the report, and
most importantly the case_scope, which determines which cases will be reported on

The specs for the new R103SarBusinessUnitPerformanceReport only exercise the new methods; the
code in the base class for analysing, totalling, calculating percentages in thoroughly testing in
the spec for the FOI report.

### DIFF
--- a/app/models/case/base.rb
+++ b/app/models/case/base.rb
@@ -423,6 +423,35 @@ class Case::Base < ApplicationRecord
     date_responded <= external_deadline
   end
 
+  # determines whether or not an individual BU responded to a case in time, measured
+  # from the date the case was assigned to the business unit to the time the case was marked as responded.
+  # Note that the time limit is different for trigger cases (the internal time limit) than for non trigger
+  # (the external time limit)
+  #
+  def business_unit_responded_in_time?
+    responding_transitions = transitions.where(event: 'respond')
+    if responding_transitions.any?
+      responding_team_assignment_date = transitions.where(event: 'assign_responder').last.created_at.to_date
+      responding_transition = responding_transitions.last
+      responding_date = responding_transition.created_at.to_date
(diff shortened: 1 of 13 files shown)


app/models/case/base.rb
```diff
@@ -423,6 +423,35 @@ def responded_in_time?
     date_responded <= external_deadline
   end
 
+  # determines whether or not an individual BU responded to a case in time, measured
+  # from the date the case was assigned to the business unit to the time the case was marked as responded.
+  # Note that the time limit is different for trigger cases (the internal time limit) than for non trigger
+  # (the external time limit)
+  #
+  def business_unit_responded_in_time?
+    responding_transitions = transitions.where(event: 'respond')
```
beccapearce: We don't have a 'respond' event for SARs anymore, the KILO just closes the case and enters their own date

app/models/case/base.rb
```diff
@@ -423,6 +423,35 @@ def responded_in_time?
...
+      internal_deadline = deadline_calculator
+                              .internal_deadline_for_date(correspondence_type, responding_team_assignment_date)
+      internal_deadline >= responding_date
+    else
+      raise ArgumentError.new("Cannot call ##{__method__} on a case without a response (Case #{number})")
+    end
+  end
+
+  def business_unit_already_late?
+    if transitions.where(event: 'respond').any?
```
beccapearce: Again we don't use the event 'respond' anymore

PR #261, opened 2018-06-18
CT-1762 allows request amends action on SARs

beccapearce:


app/policies/case/sar_policy.rb
```diff
@@ -74,6 +74,20 @@ def progress_for_clearance?
       check_user_is_a_responder_for_case
   end
 
+  def request_amends?
```
misaka: With this here, I think we still have `Case::BasePolicy#request_amends?`. As that's clearly about FOI cases now, I think we should move that to `Case::FOI::StandardPolicy#request_amends?` now.

misaka (changes requested):

spec/features/cases/sar/case_approval.rb/sar_approval_spec.rb
```diff
@@ -68,4 +68,28 @@ def undo_de_escalate_case_as_disclosure_specialist(kase, case_list_item)
     de_escalate_case_as_disclosure_specialist(kase, case_list_item)
     undo_de_escalate_case_as_disclosure_specialist(kase, case_list_item)
   end
+
```
misaka: No issues with the diff here, just ... `case_approval.rb` is the dir name? 😁 I'll change as part of CT-1755 which I'm basing on this branch. :)

misaka (approved):

PR #268, opened 2018-06-21
CT-1791 add filter for sar cases

misaka:


misaka:
Not merging until we can get clarity on whether this needs to be hidden depending on a feature flag, and whether this option should be visible depending on whether users have permissions to see SAR cases.

beccapearce (approved):

PR #274, opened 2018-06-22
CT-1775 Set up ICO classes

stephenrichards:
This work was mainly about laying down the groundwork for ICOs by adding the ICO correspondence type and the Case::ICO::FOI and Case::ICO::SAR models, adding a feature flag to disable ICOs in prod, etc.


config/state_machine/moj.yml
```diff
@@ -1211,3 +1212,186 @@ case_types:
               closed:
                 add_message_to_case:
                   if: Workflows::Predicates#user_is_assigned_disclosure_specialist?
+
+  ico:
+    name: Information Commissioner Office appeal
+    permitted_workflows:
+      - standard
```
beccapearce: shouldn't we make this a trigger only workflow so we can show the trigger badge based on workflow

spec/models/case/base_spec.rb
```diff
@@ -478,10 +478,6 @@
     it { should_not allow_value('foobar.com').for :email  }
   end
 
-  describe '#subject' do
```
beccapearce: should we be removing this? We're still performing this validation in the base model

config/state_machine/moj.yml
misaka: Good idea. I'll put this into the tickets about displaying ICO cases with the "Trigger" tag as it's also a code change and that'll be the right place to test it.

spec/models/case/base_spec.rb
misaka: Good catch, thanks!

beccapearce (approved):

PR #275, opened 2018-06-22
CT-1795 provide #internal_deadline_for_date method

stephenrichards:
Fix bug on dev where SARS reports couldn't be downloaded

### DIFF
--- a/app/services/deadline_calculator/calendar_days.rb
+++ b/app/services/deadline_calculator/calendar_days.rb
@@ -15,6 +15,11 @@ module DeadlineCalculator
       kase.received_date + kase.correspondence_type.internal_time_limit.days
     end
 
+    def internal_deadline_for_date(correspondence_type, date)
+      days = correspondence_type.internal_time_limit.days
+      date + days
+    end
+
     def external_deadline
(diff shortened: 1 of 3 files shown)


app/services/deadline_calculator/calendar_days.rb
```diff
@@ -15,6 +15,11 @@ def internal_deadline     # aka draft deadline
       kase.received_date + kase.correspondence_type.internal_time_limit.days
     end
 
+    def internal_deadline_for_date(correspondence_type, date)
+      days = correspondence_type.internal_time_limit.days
+      date + days
+    end
```
aliuk2012: Shouldn't there be a test for this? Also should we not have a test that recreates the existing bug and which passes once this method is introduced.

stephenrichards: To be  honest, I'm not sure why it is even running this on dev - it only gets used when trying to calculate the business unit performance columns, which we took out.  But point taken, I'll write a test and work out why it was being called in the first place on dev.

beccapearce (approved):

PR #277, opened 2018-06-24
CT-1334 Custom report design updates and SAR reports

aliuk2012:


db/migrate/20180621093054_update_cabinet_office_stats_to_be_custom.rb
```diff
@@ -0,0 +1,19 @@
+class UpdateCabinetOfficeStatsToBeCustom < ActiveRecord::Migration[5.0]
+  def up
+    rt = ReportType.find_by_abbr('R004')
+    if rt.nil?
+      ReportType.find_or_create_by!(abbr:'R004', full_name: 'Cabinet Office report', class_name: 'Stats::R004CabinetOfficeReport', custom_report: true, seq_id: 400)
```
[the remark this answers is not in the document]
misaka: Also, since this is a data migration, we should define a skeleton `ReportType` class in it with the minimum of machinery needed to make the `find_by_abbr` and `create!` work.

db/migrate/20180621093054_update_cabinet_office_stats_to_be_custom.rb
```diff
@@ -0,0 +1,19 @@
...
+    else
+      rt.update custom_report: true
+    end
+  end
+
+  def down
+    rt = ReportType.find_by_abbr('R004')
+    if rt.nil?
+      ReportType.find_or_create_by!(abbr:'R004', full_name: 'Cabinet Office report', class_name: 'Stats::R004CabinetOfficeReport', custom_report: false, seq_id: 400)
```
misaka: We probably shouldn't be creating R004 on the `down` migration, should we?

db/migrate/20180625160408_update_sa_rname.rb
```diff
@@ -0,0 +1,9 @@
+class UpdateSaRname < ActiveRecord::Migration[5.0]
```
misaka: The camel-casing here looks off, should it be `UpdateSARName`?

misaka (changes requested):

misaka (approved):

PR #280, opened 2018-06-25
CT-1793 Add SAR Monthly report

stephenrichards:
This is acheived by moving most of the code from the R005MonthlyPerformanceReport
into a BaseMonthlyPerformanceReport class, and deriving the original R005 report
for FOIs and the new R105 report for SARs from it.

Test for the new R105SarMonthlyPerformanceReport only exercise the methods in the
derived class, .title, .desdcription and #case_scope; all the methods in the base
class are fully exercised inthe spec for R005.


app/views/stats/index.html.slim
```diff
@@ -27,7 +27,7 @@ hr
             = "Action"
       tbody
         - @reports.each do |report|
-          - if FeatureSet.sars.disabled? && report.abbr == 'R103'
+          - if FeatureSet.sars.disabled? && report.abbr =~ /^R10\d$/   # the 100 series refer to SARS reports
```
misaka: Would it be safer to use `/^R1\d\d$/` here? Otherwise we have an assumption that we'll always have `0` in the middle and the failure caused by this is likely to be easily overlooked.

db/structure.sql
```diff
@@ -1610,6 +1610,8 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20180517140929'),
 ('20180522132456'),
 ('20180524132031'),
-('20180613141421');
+('20180613141421'),
+('20180620135756'),
+('20180622153909');
```
misaka: I don't see the migration `20180620135756` in this PR.

misaka (changes requested):

beccapearce (approved):

misaka (approved):

PR #291, opened 2018-06-28
CT-1753 change flash message after creating a case

beccapearce:
add a flash message after cases are created
update the flash message after assignment

### DIFF
--- a/spec/controllers/cases_controller_spec.rb
+++ b/spec/controllers/cases_controller_spec.rb
@@ -791,6 +791,11 @@ RSpec.describe CasesController, type: :controller do
           expect(created_case.received_date).to eq Time.zone.today
         end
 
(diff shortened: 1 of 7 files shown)


spec/controllers/cases_controller_spec.rb
```diff
@@ -791,6 +791,11 @@ def sar_closure_params(sar)
           expect(created_case.received_date).to eq Time.zone.today
         end
 
+        it 'diplays a flash message' do
+          post :create, params: params
+          expect(flash[:notice]).to eq "#{created_case.type_abbreviation} case created<br/>Case number: #{created_case.number}"
+        end
```
aliuk2012: I think you need a similar test for the assignments controller.

aliuk2012 (approved):

PR #293, opened 2018-06-29
CT-1327 add factories for ICO cases

misaka:
Just the factories.


app/models/case/ico/base.rb
```diff
@@ -18,4 +21,9 @@ def sent_by_email?
   def requires_flag_for_disclosure_specialists?
     false
   end
+
+  def set_deadlines
+    days = correspondence_type.internal_time_limit.business_days
```
stephenrichards: I think we need to look at this - I know the deadlines are being set based on user input, but we shouldn't be using business days in the model I don't think - we have the Deadline calculator to do that.

stephenrichards (approved):

PR #296, opened 2018-06-29
Replace Poltergeist/PhantomJS with Headless Chrome

aliuk2012:
Some issues identify with the existing tests
- `features/cases` calls `login_step` 18 times. `login_step` tests the full login as a user journey.
- `features/cases/filters/` need to be refactored. on avg takes over 1min to run 11 specs. A lot of DB setup/teardown is happening.


stephenrichards (approved):

README.md
```diff
@@ -87,16 +87,36 @@ $ xcode-select --install
...
+**Debugging:**
+
+To debug a spec that requires JavaScript, you need to set a environment variable called CHROME_DEBUG.
+It can be set to any value you like.
+
+Examples:
+
+```
+$ CHROME_DEBUG=1 bundle exec rspec
+```
```
misaka: I say cut out the other examples here and just leave this one. It's not really adding anything but noise ... there's no value in understanding what you can set `CHROME_DEBUG` to.

spec/features/admin/business_units_spec.rb
```diff
@@ -14,14 +15,12 @@
     expect(business_group_row).to be_present
     business_group_row.name.click
     expect(teams_show_page.heading)
-      .to have_text "You are viewing Business group #{bu.business_group.name}"
-
+      .to have_copy "You are viewing Business group #{bu.business_group.name}"
```
misaka: Doesn't adding `visible: true` as an argument to `have_copy` work?

aliuk2012: I'll give this a try.

aliuk2012: I tried  this 
```
expect(teams_show_page.heading)
      .to have_text("You are viewing Business group #{bu.business_group.name}",visible: true)
```

and got this error:
```
1) editing teams editing a business unit
     Failure/Error:
       expect(teams_show_page.heading)
         .to have_text("You are viewing Business group #{bu.business_group.name}",visible: true)
     
     ArgumentError:
       invalid keys :visible, should be one of :count, :minimum, :maximum, :between, :wait, :exact
     # ./spec/features/admin/business_units_spec.rb:17:in `block (2 levels) in <top (required)>'

```

aliuk2012: This commit introduces the new have_copy matcher: 2c553267f79d351cad5f1efaf6f17f0297f3341c to address the issue with hidden text

misaka: Sorry I got that wrong in my comment. What I should've said is you can add `visible: false` to the element matcher in the SitePrism object, for example in [PageObjects::Sections::Cases::TypeFilterPanelSection](https://github.com/ministryofjustice/correspondence_tool_staff/blob/master/spec/site_prism/page_objects/sections/cases/type_filter_panel_section.rb#L8). This is how I've handled this in the past.

misaka: Ok, this is slowly coming into focus now. :) So Chrome's web driver is matching differently now, not finding hidden text, and using `visible: false` on the matcher is only useful when the entire element is hidden. I think I'm up-to-speed now.

misaka (approved):

PR #304, opened 2018-07-04
Ct 1778 ICO filter

stephenrichards:


spec/factories/case/ico_fois.rb
```diff
@@ -4,18 +4,69 @@
...
       managing_team   { find_or_create :team_dacu }
     end
 
     current_state          'unassigned'
+    sequence(:name)        { |n| "#{identifier} name #{n}" }
     sequence(:subject)     { |n| "#{identifier} subject #{n}" }
     sequence(:message)     { |n| "#{identifier} message #{n}" }
     ico_reference_number   { generate :ico_foi_reference_number }
-    received_date          { 0.business_days.from_now }
+    # received_date          { 0.business_days.from_now }
```
misaka: We should probably use this version using business day calculation, vs the `received_date` below.

misaka (approved):

PR #310, opened 2018-07-10
Ct 1817 allow disclosure to take on ico for approval

beccapearce:


config/state_machine/moj.yml
```diff
@@ -1263,6 +1264,7 @@ case_types:
                 assign_to_new_team:
                 destroy_case:
                 edit_case:
+                flag_for_clearance:
```
misaka: What is this action being used for, since these cases are always flagged for clearance?

spec/state_machines/workflows/ico_permitted_events_spec/ico_appeal_trigger_spec.rb
```diff
@@ -0,0 +1,228 @@
+require 'rails_helper'
+
+describe ConfigurableStateMachine::Machine do
+  context 'trigger ico workflow' do
+
+  ##################### MANAGER FLAGGED ############################
```
misaka: Good stuff adding these missing tests in!!

misaka (approved):
Just need to check if the `flag_for_clearance` action actually needs to be there.

spec/state_machines/workflows/ico_permitted_events_spec/ico_appeal_trigger_spec.rb
beccapearce: I think we're also missing the ico_event_spec if you ever want to add it in :)

PR #311, opened 2018-07-10
CT-1776 add case link to ico

misaka:


app/models/linked_case.rb
```diff
@@ -0,0 +1,74 @@
+class LinkedCase < ApplicationRecord
+  self.inheritance_column = :_unused
```
stephenrichards: would have preferred to see the type column called `link_type`, then there would be no need to disable the STI

app/models/linked_case.rb
```diff
@@ -0,0 +1,74 @@
+class LinkedCase < ApplicationRecord
```
stephenrichards: Can you run `bundle exec annotate` so that we get a list of columns?

app/models/linked_case.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def create_reverse_link
+    self.class.skip_callback(:create, :after, :create_reverse_link)
+    self.class.create(case_id: linked_case_id, linked_case_id: self.case_id)
+  rescue ActiveRecord::RecordNotUnique
+    nil
+  ensure
+    self.class.set_callback(:create, :after, :create_reverse_link)
+  end
```
stephenrichards: Wouldn't it be easier to see if the link was already there, and if not, create it?  That would save having to skip the callback, rescue RecordNotUnique and reset the call back.

app/validators/case_link_type_validator.rb
```diff
@@ -0,0 +1,68 @@
...
+    },
+  }.with_indifferent_access
+
+  class << self
+    def classes_can_be_linked_with_type?(type:, klass:, linked_klass:)
+      if klass.to_s.in?(ALLOWED_LINKS_BY_TYPE[type])
+        linked_klass.to_s.in?(ALLOWED_LINKS_BY_TYPE[type][klass.to_s])
+      else
+        false
+      end
```
stephenrichards: again, would be a bit clearer if `type` were renamed `link_type` - had me scratching my head there for a moment working out what the difference between `type` and `klass` was

app/models/linked_case.rb
misaka: I wanted do have the option to enable STI without having to rename the column, etc, in the future, and I also thought it read well as 'link.type'.

app/models/linked_case.rb
misaka: Yes!

app/models/linked_case.rb
misaka: Hrm. I think I was trying to optimise to reduce DB access. Just to try it out, the code would look like:

```
def create_reverse_link
  unless self.class.exist(case_id: linked_case_id, linked_case_id: self.case_id)
    self.class.skip_callback(:create, :after, :create_reverse_link)
    self.class.create(case_id: linked_case_id, linked_case_id: self.case_id)
    self.class.set_callback(:create, :after, :create_reverse_link)
  end
end
```

A little bit tidier, but it isn't atomic and could cause introduce a race condition ... it should probably be protected with a rescue:

```
def create_reverse_link
  unless self.class.exist(case_id: linked_case_id, linked_case_id: self.case_id)
    begin
      self.class.skip_callback(:create, :after, :create_reverse_link)
      self.class.create(case_id: linked_case_id, linked_case_id: self.case_id)
    rescue ActiveRecord::RecordNotUnique
      nil
    ensure
      self.class.set_callback(:create, :after, :create_reverse_link)
    end
  end
end
```

What do you think?

app/validators/case_link_type_validator.rb
misaka: Hrm. Ya, I see why you made the earlier comment then. Is it worth renaming now?

app/models/linked_case.rb
stephenrichards: I was thinking more that the LinkCase Service should call something like `LinkedCases.link(case_a, case_b, link_type)`
 and that the LinkedCases.link method should take care of making both links in a transaction.  No need for after create hooks, which sometimes are more trouble than they are worth.

app/validators/case_link_type_validator.rb
stephenrichards: Let's do it as a separate thing.  Good to get this merged in ASAP I think.

stephenrichards (approved):

PR #321, opened 2018-07-17
Sectioned State machine config files

stephenrichards:


misaka (approved):
It's great having these split up. I do think that it'd be nice to have a more sophisticated approach that would be more intuitive for anyone that has to modify the structure (see my comments below), but I realise that that's a much bigger change.

PR #324, opened 2018-07-19
CT-1768 redesign report page

aliuk2012:
- Updates stats home page
- Moves audit csv to the admin dashboard
- moves the business unit map to the bottom of the team pages


db/migrate/20180621094208_add_correspondence_type_columns_to_report_types.rb
```diff
@@ -4,8 +4,6 @@ class AddCorrespondenceTypeColumnsToReportTypes < ActiveRecord::Migration[5.0]
   def up
     add_column :report_types, :foi, :boolean, default: false
     add_column :report_types, :sar, :boolean, default: false
-    ReportType.reset_column_information
```
beccapearce: Bit worried about removing the seeder bit in this migration

beccapearce (commented):
All looks good, but want to check about moving the seeder out of the migration, could be fine but want to double check :)

db/migrate/20180621094208_add_correspondence_type_columns_to_report_types.rb
stephenrichards: Adding a column to a table, and updating the seeds to populate that column cause problems when running rake db:reseed.  The way to do it is to maintain a strict separation of schema migrations and data migrations, so I've moved them all into data migrations.

Reseeding will run all the schema migrations, and then all the data migrations which will ensure everything is populated.  Ditto on production systems, we just run all the outstanding schema migrations first, followed by all the outstanding data migrations.

app/models/report_type.rb
```diff
@@ -2,26 +2,57 @@
...
 
 class ReportType < ApplicationRecord
+
+  VALID_DEFAULT_REPORTING_PERIODS = %w{
+                                        year_to_date
+                                        quarter_to_date
+                                        last_quarter
+                                        last_month
+                                      }
+
```
misaka: It'd be nice to have this as an `enum`. It appropriately descriptive and would give us scopes and validation for free.

app/services/stats/base_appeals_performance_report.rb
```diff
@@ -15,10 +15,10 @@ class BaseAppealsPerformanceReport < BaseReport
         responsible:                     ''
     }
 
-    def initialize(period_start= Time.now.beginning_of_year, period_end=Time.now)
+    def initialize(period_start = nil, period_end = nil)
```
misaka: `period_start` and `period_end` are not used in this method. And is there a reason to keep the commented-out assignments to `@period_start` and `@period_end` below?

stephenrichards (approved):

app/services/stats/base_appeals_performance_report.rb
misaka: How about prefixing these with underscore ('_') to denote that they're intentionally unused?

misaka (approved):

PR #326, opened 2018-07-24
CT-1829 ico create fixes

misaka:


app/views/cases/ico/_new_form.html.slim
```diff
@@ -1,3 +1,5 @@
+= f.text_field :ico_officer_name
```
aliuk2012: I thought you were going to play back this field on the case details page as part of this PR.

config/locales/en.yml
```diff
@@ -593,6 +593,10 @@ en:
       received_date: Date received at MOJ
       received_date_copy: "For example, 01 02 2018"
       case/ico:
+        external_deadline_copy: >-
+          We'll use this date to set the draft deadline, calculated as 20
+          working days from the final deadline.<br>
+          You will be able to change both dates later.
```
aliuk2012: Not sure I agree with this copy.  This sounds like the draft deadline is 20 workings after the final deadline. Perhaps we should remove this because I think we were going to allow users to input their own draft deadline.

spec/views/cases/new_html_slim_spec.rb
```diff
@@ -98,7 +105,7 @@
       expect(page).to have_external_deadline_month
       expect(page).to have_external_deadline_year
 
-      expect(page).to have_subject
+      expect(page).not_to have_subject
```
aliuk2012: could this be `expect(page).to have_no_subject` i believe using `have_no` is more performant than using `not_to`

app/views/cases/ico/_new_form.html.slim
misaka: Ya, but I guess the case details page wasn't ready yet, not fussed either way.

spec/views/cases/new_html_slim_spec.rb
misaka: Sure, will change it.

config/locales/en.yml
misaka: I'm sure Mia would be open to better wording of this. I think we went over it once together and I said the same thing, we can have a look tomorrow.

app/models/case/base.rb
```diff
@@ -137,7 +137,6 @@ def self.searchable_document_tsvector
   scope :internal_review_timeliness, -> { where(type: 'Case::FOI::TimelinessReview')}
   scope :deadline_within, -> (from_date, to_date) { where("properties->>'external_deadline' BETWEEN ? AND ?", from_date, to_date) }
   validates :current_state, presence: true, on: :update
-  validates :subject, presence: true, length: { maximum: 100 }
   validates :email, format: { with: /\A.+@.+\z/ }, if: -> { email.present? }
   validates_presence_of :received_date
```
beccapearce: Do we validate the ICO officer name?

app/views/cases/ico/_new_form.html.slim
misaka: I've added this now.

beccapearce (approved):

aliuk2012 (approved):

PR #327, opened 2018-07-24
See user teams and cases on admin page

misaka:
- Move dashboards into admin section
- Add tabs for admin sections: Cases, Users, Feedback, List Queries and Search
  Queries
- Add cases dashboard onto cases list page
- Add extra info to users listing: assigned teams and assigned cases

screenshot_2018-07-24 gov uk - the best place to find government services and information

screenshot_2018-07-24 admin case list - track-a-query


app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,9 @@
+class AdminController < ApplicationController
+  layout "admin"
+
+  private
```
beccapearce: This should be covered by the checks done in the routes with some tests

aliuk2012 (approved):
love the design and layout :-)

PR #331, opened 2018-07-27
CT-1768 Removes non-standard reports and renames some reports

aliuk2012:


db/data_migrations/20180727101528_update_standard_reports_and_rename_reports.rb
```diff
@@ -0,0 +1,32 @@
...
+            full_name: 'Business unit report'
+        },
+        'R105'  => {
+            standard_report: true,
+        }
+    }
+
+    attrs.each do |report_abbreviation, new_values|
+      rt = ReportType.find_by!(abbr: report_abbreviation)
+      rt.update(new_values)
```
misaka: Probably worth doing an `update!` here too?

misaka (approved):

PR #342, opened 2018-07-30
Ct 1832 ICO case details

aliuk2012:


app/views/cases/_case_status.html.slim
```diff
@@ -12,6 +12,12 @@
           = t('common.case.who_its_with')
         .case-status__data
           = case_details.who_its_with
+    - if case_details.type_abbreviation == 'ICO'
```
stephenrichards: I think we should start having methods on the class which we can interrogate, e.g. 
`case details.ico?`
It makes it a lot easier to read rather than testing the correspondence type abbreviation, and the correspondence type abbreviations don't match up exactly to the case type in any case.

We don't need to do this now, but worth thinking about for the future.

stephenrichards (approved):

app/views/cases/_case_status.html.slim
misaka: 👍

PR #343, opened 2018-07-30
Ct 1788 Create OverturnedICO::SAR

stephenrichards:


app/controllers/cases_controller.rb
```diff
@@ -152,45 +152,43 @@ def new
...
+      flash[:notice] = t('activerecord.errors.models.case.attributes.number.duplication')
+      render :new
     end
+  end
 
-  rescue ActiveRecord::RecordNotUnique
-    flash[:notice] =
-      t('activerecord.errors.models.case.attributes.number.duplication')
-    render :new
+  def new_overturned_ico
```
misaka: I see that the functionality required here is quite unique to Overturned ICO cases ... but I'm just a bit concerned about having case-specific methods at the controller level like this. I feel like anything case-specific should be in a service or a concern. Can we try to move this out at some point, not necessarily for this PR?

app/controllers/concerns/sar_cases_params.rb
```diff
@@ -3,6 +3,7 @@ module SARCasesParams
 
   def create_sar_params
     params.require(:case_sar).permit(
+      :correspondence_type,
```
misaka: Same comment as in `create_foi_params`.

app/models/case/base.rb
```diff
@@ -61,7 +61,7 @@ def self.searchable_document_tsvector
 
   attr_accessor :message_text
 
-  acts_as_gov_uk_date :received_date, :date_responded,
+  acts_as_gov_uk_date :received_date, :date_responded, :external_deadline,
```
misaka: You can move this into ICO Overturned base class, you just need to include _all_ the dates that `acts_as_gov_uk_date`, until we fix that to be cumulative.

app/models/case/ico/sar.rb
```diff
@@ -1,8 +1,17 @@
 class Case::ICO::SAR < Case::ICO::Base
 
+  has_paper_trail only: [
```
misaka: Won't this override the `has_paper_trail` in `Case::ICO::Base`? That one contains fields we want to track like `ico_reference_number` which should be in here, or better yet just use that?

app/models/case/overturned_ico/foi.rb
```diff
@@ -0,0 +1,7 @@
+class Case::OverturnedICO::FOI < Case::OverturnedICO::Base
+
+  def self.type_abbreviation
+    'OVERTURNED FOI '
```
misaka: Should this be `OVERTURNED_FOI`?

misaka (approved):

app/controllers/cases_controller.rb
stephenrichards: I did think about this.  The problem is I think that this is different from any other new action - it takes a parameter(the id of the ICO appeal from which it is created).  I'll have a play around and see if that causes any problems.

app/models/case/base.rb
stephenrichards: As you saw, removing it from Case::Base also seems to remove it from everything that derives from Case::Base.  Probably dependent on load order.  Would be fixed by fixing gov_uk_date_fields gem to allow multiple definitions.

app/models/case/overturned_ico/foi.rb
```diff
@@ -1,7 +1,7 @@
 class Case::OverturnedICO::FOI < Case::OverturnedICO::Base
 
   def self.type_abbreviation
-    'OVERTURNED FOI '
+    'OVERTURNED_FOI '
```
misaka: There's still a space here, I'm guessing this isn't getting in the way because we don't create overturned FOIs? We can fix this later when it becomes an issue.

misaka (approved):

PR #345, opened 2018-07-31
Add predicate methods for each case type

stephenrichards:


beccapearce (approved):
Cool, can we add them for ICO cases too?

PR #346, opened 2018-07-31
kanpsack

stephenrichards:


db/structure.sql
```diff
@@ -1634,8 +1634,10 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20180613141421'),
 ('20180620135756'),
 ('20180621094208'),
+('20180622153909'),
```
misaka: Should these be part of this PR?

misaka (changes requested):

PR #350, opened 2018-08-01
CT-1867 show response to ICO Appeal

aliuk2012:
- Moves the Response section position on the case details page depending on whether the case is an ICO or not.
- Uses translations for case to determine the section heading


config/locales/en.yml
```diff
@@ -646,6 +646,13 @@ en:
     select_type:
       heading: Choose correspondence type
       sub_heading: Create case
+    show:
+      case/ico:
+        attachments_heading: Appeal response
+      case/sar:
```
beccapearce: I think for this part we don't need to repeat Response for case/foi and case/sar I think if you do `    show:
      case/ico:
        attachments_heading: Appeal response
      case:
        attachments_heading: Response` it should work

aliuk2012: Thanks I didn't realise you could set a default.

aliuk2012: I didn;t realise you could use a default. the only problem is that you then have to add the default key to the ignore list

beccapearce (approved):

PR #353, opened 2018-08-01
CT-1783 Submit ICO

stephenrichards:


config/locales/en.yml
```diff
@@ -398,14 +410,6 @@ en:
...
     response_upload_error: Errors detected with uploaded files.
     response_upload_blank?: Please select the file(s) you used in your response.
-    respond_point_of_no_return_html: |
-      <p class="notice">
-        <i class="icon icon-important">
-          <span class="visually-hidden">Important</span>
-        </i>
-        <strong>You can't update a response after marking it as sent.</strong>
-      </p>
-
```
aliuk2012: Why was this removed?

config/locales/en.yml
```diff
@@ -657,14 +663,10 @@ en:
...
       heading: Request amends
     respond:
-      heading: Marking it as sent
-      reminders_html: |
-        <strong>Make sure you have:</strong>
-        <ul class="list list-bullet">
-          <li>cleared the response with the Deputy Director</li>
-          <li>uploaded the response and any supporting documents</li>
-          <li>sent the response to the person who made the request</li>
-        </ul>
```
aliuk2012: Why is the reminders key removed?

spec/features/cases/ico/respond_to_ico_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    cases_show_page.actions.mark_as_sent.click
+
+    cases_respond_page.back_link.click
+
+    expect(cases_show_page).to be_displayed(kase.id)
+
+  end
+
+end
```
aliuk2012: Do we need this feature test if we are including it in the user journey e2e tests?

aliuk2012 (changes requested):

config/locales/en.yml
beccapearce: it wasn't in the designs and we decided to make the FOI and ICO page the same

config/locales/en.yml
beccapearce: it wasn't in the designs and we decided to make the FOI and ICO page the same

app/views/cases/respond.html.slim
```diff
@@ -1,8 +1,13 @@
+= GovukElementsErrorsHelper.error_summary @case.object, @case.error_summary_message, ''
+
 - content_for :page_title do
   = t('page_title.mark_as_sent_page', case_number: @case.number)
 
 - content_for :heading
   = t('.heading')
+  br
```
misaka: Having these two in the heading looks a little bit odd to me:

screenshot_2018-08-03 marking response as sent - 180803001 - track-a-query

I don't see designs for it but it seems to me that other pages don't have the page heading and the case subject both so big.

misaka (changes requested):

app/views/cases/respond.html.slim
aliuk2012: @misaka I agree with you. Its something I've been speaking to Mia about, oddly its on this page and some of the "Close cases" page.

config/locales/en.yml
aliuk2012: 🤔  Still think this would be important for KILOs and FOI.

spec/features/cases/ico/respond_to_ico_spec.rb
aliuk2012: I think I would leave it down to the controller and model to throw and handle validation errors etc.

aliuk2012: if we later find a bug from an unhappy path we can add it and fix it

app/views/cases/respond.html.slim
beccapearce: Here are the designs: https://e25x69.axshare.com/#g=1&p=ico_appeal_case_sent_to_ico
If we want to change this I suggest doing it in another PR once you've agreed a change with Mia

app/state_machines/configurable_state_machine/machine.rb
```diff
@@ -82,7 +82,9 @@ def respond_to_missing?(method, include_private = false)
 
     def event_name(event)
       if events.include?(event.to_sym)
-        I18n.t("event.#{event}", default: event.to_s.humanize)
+        specific_key = "event.case/#{@kase.type_abbreviation.downcase}.#{event}"
+        default_key = "event.#{event}"
+        I18n.t(specific_key, default: I18n.t(default_key))
```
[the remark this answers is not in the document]
beccapearce: Totally agree!

config/locales/en.yml
beccapearce: I think that this is a question for design. Once Mia is back we can discuss it with her, but I think keeping it to her designs for now is the best thing, we can always add it back in

app/views/cases/respond.html.slim
misaka: @beccapearce Thanks for the link. As discussed in stand-up, let's just get this through and address this separately, if at all.

misaka (approved):

config/locales/en.yml
```diff
@@ -662,14 +668,10 @@ en:
...
+      heading: Mark as sent
+      respond_date: Add date this response was sent to ICO
```
aliuk2012: this copy is appearing for both FOI and SAR
image

config/locales/en.yml
aliuk2012: Mia agrees that we should keep "Make sure.." copy just for FOI.

image

Ignore the action buttons in this screenshot it should have one green and one link.

beccapearce: I think if we're going to change the designs on this ticket we should do it as part of another ticket

aliuk2012: I've just added this copy back to the page and added a spec. The reason is that this change to FOI wasn't actually defined in the original ticket but it did make the ticket a little easier to implement.

aliuk2012 (approved):

PR #355, opened 2018-08-03
CT-1788 Fixes as a result of rebase

stephenrichards:


app/controllers/stats_controller.rb
```diff
@@ -89,7 +89,10 @@ def report_is_current?(report)
                                 report.report_type.abbr
     scheduled_time = job_previous_run_time(job_config)
     report.created_at >= scheduled_time
-  rescue
+  rescue => err
+    puts ">>>>>>>>>> RESVUE #{__FILE__}:#{__LINE__} <<<<<<<<<<"
```
misaka: RESVUE? A special little something to go with this PR? 😄

app/views/cases/overturned_sar/_new_form.html.slim
```diff
@@ -36,12 +24,13 @@ hr
...
           form_hint_text: "We'll use this date to set the draft deadline, calculated as 20 working days from the final deadline."}
 
+
 p
   = "Only change the details below if the Requester's contact details and the "
   = "preferred way of receiving the information has changed."
 
 #delivery-method
-  = f.radio_button_fieldset :delivery_method, choices: [:send_by_email, :send_by_post] do |fieldset|
+  = f.radio_button_fieldset :reply_method, choices: [:send_by_email, :send_by_post] do |fieldset|
```
misaka: Can we sync up the id of the div here with the field name?

@aliuk2012 Do you reckon we should just remove it IF it's not even being used?

app/views/cases/overturned_foi/new.html.slim
```diff
@@ -0,0 +1 @@
+= render partial: 'cases/overturned_sar/new', locals: { kase: @case }
```
misaka: Is this correct? Do we need `overturned_foi/new`?

app/models/case/base.rb
```diff
@@ -8,7 +8,7 @@
 #  message              :text
 #  created_at           :datetime         not null
 #  updated_at           :datetime         not null
-#  received_date        :date
+#  recevied_date        :date
```
aliuk2012: typo???

app/views/cases/overturned_foi/new.html.slim
aliuk2012: @misaka I think this is correct we are doing overturned SARs not FOI

app/views/cases/overturned_sar/_new_form.html.slim
aliuk2012: @misha I've got a feeling its used for tests but I agree with you it should be sync up to the field that we are using.  BTW why are we using a new attr called reply_method? Isn't this the same information we are asking for on regular SARs couldn't we just reuse the same attributes as a regular SAR i.e delivery_method. i

app/views/cases/overturned_foi/new.html.slim
stephenrichards: We have two correspondence types: overturned_sar, and overturned_ico.  As the view path is determined by the correspondence type abbreviation we need two files, one int he overturned_foi directory, and one in the overturned_sar directory.  But they just include the correct partials.  the other way of doing it was to have a special method on correspondence type (or case) that returns the view path to use, and change everywhere we generate the virew path using the type abbreviation to use this new method.  I chose the path of least resistance.

stephenrichards: yep, `reply_method` was introduced for SARs, and its meaning is slightly different.

`delivery_method` refers to the way in which the case was sent to us, and can have values `sent_by_post` and `sent_by_email` (past tense).
`reply_method` refers to the way in which the response will be sent, and can have the values `send_by_post` and `send_by_email`

app/models/case/overturned_ico/foi.rb
```diff
@@ -1,7 +1,8 @@
 class Case::OverturnedICO::FOI < Case::OverturnedICO::Base
 
   def self.type_abbreviation
-    'OVERTURNED_FOI '
+    'OVERTURNED_FOI'
```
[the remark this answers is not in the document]
stephenrichards: When we do the Overturned FOIs.  This PR is for SARs only.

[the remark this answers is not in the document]
aliuk2012: But if we already have this method shouldn't we test it otherwise we risk just ignoring it. Having had a look through `spec/models` it looks like we only test  type_abbreviation for FOI and OVERTURNED_SAR. We are not tests original SAR or ICO Appeals.

app/services/case_create_service.rb
```diff
@@ -47,6 +47,7 @@ def create_params(correspondence_type)
 
   def create_case
     @case = @case_class.new(@permitted_params.merge(uploading_user: @user))
+    # @case.original_case = @case.original_ico_appeal.original_case
```
beccapearce: DO we still need this commented out code?

app/views/cases/overturned_foi/new.html.slim
misaka: After discussing it seems that this should be in shared.

aliuk2012 (approved):

PR #367, opened 2018-08-08
CT-1788 Prepopulate reply method and email/address

stephenrichards:


app/services/create_overturned_ico_case_service.rb
```diff
@@ -26,11 +26,12 @@ def call
                            @error = true
                        end
     if success?
-      original_case                                 = @original_ico_appeal
+      original_case                                 = @original_ico_appeal.original_case
       @overturned_ico_case                          = overturned_klass.new
-      @overturned_ico_case.subject                  = original_case.subject
+      # @overturned_ico_case.subject                  = original_case.subject
```
beccapearce: Commented out code

beccapearce (approved):

app/decorators/case/overturned_ico/sar_decorator.rb
```diff
@@ -4,5 +4,8 @@ def original_case_description
     "ICO appeal (SAR) #{object.original_ico_appeal.number}"
   end
 
+  def type_abbreviation
+    'Overturned SAR'
```
aliuk2012: Why is this in the decorator? All the other type_abbreviations are directly in the models. it also saves us calling decorate on the object.

aliuk2012: Theres no tests for this method

app/models/case/sar.rb
```diff
@@ -69,6 +69,10 @@ def within_escalation_deadline?
     false
   end
 
+  def sar?
+    true
+  end
```
aliuk2012: Shouldn't there be a test for this?

app/services/case_create_service.rb
```diff
@@ -71,7 +70,7 @@ def set_flagged_for_disclosure_errors
 
   def flag_for_disclosure_if_required
     @case.save!
-    @flash_notice = "#{@case.type_abbreviation} case created<br/>Case number: #{@case.number}".html_safe
+    @flash_notice = "#{@case.decorate.type_abbreviation} case created<br/>Case number: #{@case.number}".html_safe
```
aliuk2012: We dont need to use decorate if we move the type abbreviation into the model itself

app/services/create_overturned_ico_case_service.rb
```diff
@@ -26,11 +26,12 @@ def call
                            @error = true
                        end
     if success?
-      original_case                                 = @original_ico_appeal
+      original_case                                 = @original_ico_appeal.original_case
       @overturned_ico_case                          = overturned_klass.new
-      @overturned_ico_case.subject                  = original_case.subject
+      # @overturned_ic_case.subject                  = original_case.subject
```
aliuk2012: Are we not copying the subject from the original case?

app/decorators/case/overturned_ico/sar_decorator.rb
stephenrichards: it''s in the decorator because "Overturned SAR created" reads better in a flash message than "OVERTURNED_SAR created"

PR #368, opened 2018-08-08
CT-1780 edit ico appeals

misaka:
https://dsdmoj.atlassian.net/browse/CT-1780

Enable editing of ICO Appeals, both FOIs and SARs. Largely based on existing views and actions for FOIs and SARs, with a new approach to organising views and site prism pages to try to manage the developing complexity a bit better.


app/views/cases/show.html.slim
```diff
@@ -21,7 +21,7 @@
 
 = render partial: 'layouts/header'
 
-div id="case-#{@case.id}"
+div id="case-#{@case.id}" class="case"
```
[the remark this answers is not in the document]
misaka: Good idea.

spec/site_prism/page_objects/pages/cases/show_page.rb
```diff
@@ -55,6 +55,10 @@ class ShowPage < SitePrism::Page
                 PageObjects::Sections::Cases::WhatDoYouWantToDoSection,
                 '.what-do-you-want-to-do'
 
+        section :ico,
+                PageObjects::Sections::Cases::ICO::ShowSection,
+                '.case'
```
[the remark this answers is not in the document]
misaka: Well, I'm not sure it matters whether the class is `case` or `case-ico` ... obviously if we do the change you suggest in `show.html.slim` we should change it here. Or am I missing something? So far as I know this is only used to hook it up to tests, which is why I wasn't sure what you'd say about doing this in the first place.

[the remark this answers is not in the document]
aliuk2012: You are correct, it is used to hook up to the tests.
What it means is that when we come to hook up FOI/SAR we should be able to do section
```
section :foi,
                PageObjects::Sections::Cases::ICO::ShowSection,
                '.case-foi'

section :sar,
                PageObjects::Sections::Cases::ICO::ShowSection,
                '.case-sar'
```

app/decorators/case/overturned_ico/sar_decorator.rb
```diff
@@ -5,6 +5,9 @@ def original_case_description
   end
 
   def type_abbreviation
+    # This string is used when constructing paths or methods in other parts of
+    # the system. Ensure that it does not come from a user-supplied parameter,
+    # and does not contain special chars like slashes, etc.
```
stephenrichards: Not quite sure why this comment is under all the type abbreviation methods.  It's hard coded for each class, so there's no danger of it being overwritten.

app/views/cases/foi/_case_details.html.slim
```diff
@@ -1,10 +1,10 @@
 .grid-row
-  .column-full.case-details
+  - link_type_prefix = defined?(link_type) && link_type.present? ? "#{link_type}-" : ""
```
stephenrichards: This could perhaps be refactored out into a helper method as we're dong the same thing in at least one other place (not that I'm a fan of helper methods - but this seems a good case for one).

app/views/cases/ico/_case_details.html.slim
```diff
@@ -1,10 +1,10 @@
...
-  .column-full.case-details
+  - link_type_prefix = defined?(link_type) && link_type.present? ? "#{link_type}-" : ""
+  div class="column-full #{link_type_prefix}case-details"
     h2.request--heading
-      - if @case.type_abbreviation != case_details.type_abbreviation
-        = "Original case details"
-      - else
-        = "Case details"
+      / i18n-tasks-use t('common.case.case-details')
+      / i18n-tasks-use t('common.case.original-case-details')
```
stephenrichards: what do these two comment lines do?  Anything special?

app/views/cases/sar/_case_details.html.slim
```diff
@@ -1,10 +1,11 @@
 .grid-row
-  .column-full.case-details
+  - link_type_prefix = defined?(link_type) && link_type.present? ? "#{link_type}-" : ""
+  / div class="column-full case-details"
+  div class="column-full #{link_type_prefix}case-details"
```
stephenrichards: old code commented out but left in?

stephenrichards (approved):

app/decorators/case/overturned_ico/sar_decorator.rb
misaka: It's a warning in case we ever decide to change it from being hard-coded. [Code climate brought this up](https://codeclimate.com/github/ministryofjustice/correspondence_tool_staff/issues?status%5B%5D=&status%5B%5D=invalid&status%5B%5D=wontfix&engine_name%5B%5D=brakeman) in the views where `type_abbreviation` is used to generate paths to templates to render so as I've silenced the warning there I thought I'd better drop a note in to make sure we're careful about changing this.

app/views/cases/foi/_case_details.html.slim
misaka: Ya, good call, worth doing.

app/views/cases/ico/_case_details.html.slim
misaka: These signal to i18n-tasks that these translations should be considered used and won't show up in `i18n-tasks unused`. Seems like a better approach than adding an ignore for `common.*case-details` into `i18n-tasks.yml` as it tracks closer to the code where they're used.

app/views/cases/sar/_case_details.html.slim
misaka: Yup, already removed locally.

PR #371, opened 2018-08-08
CT-1784 ico close page

aliuk2012:


beccapearce (changes requested):
When I closed an appeal, the document I uploaded wasn't shown on the case details page, nor was the message I wrote

The ticket also describes the state being changed to closed- overturned or closed-upheld, but it only said closed

The button has the copy record ICO decision but in the ticket is says record ICO decision and close

beccapearce (approved):
Previous comments still apply but will be tackled in another ticket. Would be nice to have the admin pages work with closed cases too though.

PR #382, opened 2018-08-13
Ct 1846 display overturned spike

beccapearce:
updates the type abbreviation method to return a string that can be used to call the current team and user service

### DIFF
--- a/app/services/current_team_and_user_service.rb
+++ b/app/services/current_team_and_user_service.rb
@@ -20,6 +20,6 @@ class CurrentTeamAndUserService
   private
 
   def resolver_for_case(kase)
-    "CurrentTeamAndUser::#{kase.type_abbreviation.delete(" ")}::#{kase.workflow.camelize}".constantize.new(kase)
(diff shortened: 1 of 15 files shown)


app/services/current_team_and_user_service.rb
```diff
@@ -20,6 +20,6 @@ def user
   private
 
   def resolver_for_case(kase)
-    "CurrentTeamAndUser::#{kase.type_abbreviation}::#{kase.workflow.camelize}".constantize.new(kase)
+    "CurrentTeamAndUser::#{kase.type_abbreviation.delete("_")}::#{kase.workflow.camelize}".constantize.new(kase)
```
stephenrichards: I can't help feeling that perhaps we'd be better off having a method on each case type which defines the current team and user class to use.  Then we wouldn't need lots of empty CurrentTeamAndUser classes.

stephenrichards (changes requested):

stephenrichards (changes requested):
See my suggestion for refatoring the whole CurrentTeamAndUser thang

stephenrichards (approved):

PR #391, opened 2018-08-17
CT-1874 add translations that are missing

beccapearce:
Do we need to test all of these? I think it makes sense to, to avoid this issue coming up again, but don't want to add a load of tests if that's a silly idea!


aliuk2012 (approved):
Lets add the tests as part of a bigger piece of work that ensures we testing the translations and their copy

PR #393, opened 2018-08-21
Add ICO Appeal responded state to Admin case creation

aliuk2012:


spec/features/admin/cases/adding_cases_spec.rb
```diff
@@ -220,6 +220,14 @@
         kase = Case::ICO::FOI.first
         expect(BusinessUnit.dacu_disclosure).to be_in(kase.approving_teams)
       end
+
+      scenario 'creating a trigger ICO in awaiting_dispatch' do
+        create_ico(target_state: 'responded')
+        expect(admin_cases_page).to be_displayed
+        kase = Case::ICO::FOI.first
+        expect(BusinessUnit.dacu_disclosure).to be_in(kase.approving_teams)
```
beccapearce: Can we add a test for the state
expect(kase.current_state).to eq 'responded'

aliuk2012: done

beccapearce (approved):

PR #396, opened 2018-08-24
CT-1891 Update case details page display once ICO case has been closed

aliuk2012:
- [x] update the case status to show what the ico decision was
- [x] displays the ICO decision and date in a new section
- [x] displays any uploaded ICO documents in a new section


app/decorators/case/base_decorator.rb
```diff
@@ -103,7 +103,11 @@ def status
     if current_state != 'closed'
       translation_for_case(object, "state", object.current_state)
     else
-      I18n.t("state.#{current_state}_status")
+      if object.ico?
+        "Closed - #{ico_decision.downcase} by ICO"
```
aliuk2012: It doesn't look like our `translation_for_case` method allows us to pass variables to the locales file. Looking at https://intranet.justice.gov.uk/guidance/security/building-security-pass/, I might have misunderstood how this works but it looks like options is not being passed to `I18n.t`

app/views/cases/shared/_closed_case_details.html.slim
```diff
@@ -1,7 +1,8 @@
 tbody.response-details
   - if case_details.date_responded.present?
     tr.date-responded
-      th = t('common.case.date_responded')
+      / i18n-tasks-use t('common.case.date_responded')
+      th = translate_for_case(case_details, 'common.case', 'date_responded')
```
misaka: This translation isn't doesn't change depending on the class of case ... why use `translate_for_case` here?

app/decorators/case/ico/sar_decorator.rb
```diff
@@ -1,7 +1,7 @@
 class Case::ICO::SARDecorator < Case::ICO::BaseDecorator
 
   def pretty_type
-    'ICO appeal for SAR case'
+    'ICO appeal(SAR)'
```
misaka: The design calls for this to be:

> ICO Appeal (SAR)

Note capital 'Appeal' and the extra space. Might need to confirm with Mia whether the capital "A" is actually correct.

aliuk2012: Checking this with Mia. ATM it matches what we have for FOI i.e "ICO appeal(FOI)"

spec/decorators/case/ico/base_decorator_spec.rb
```diff
@@ -1,7 +1,19 @@
 require "rails_helper"
 
 describe Case::ICO::BaseDecorator do
+  let(:closed_ico_foi_case) { (create :closed_ico_foi_case).decorate }
+  let(:closed_ico_sar_case) { (create :closed_ico_sar_case).decorate }
```
misaka: This case type isn't used here. We discussed whether this is the right place to, essentially, test `Case::ICO::FOIDecorator` and `Case::ICO::SARDecorator` ... but it's not really clear what place is the best. :( I say we take the path of least resistance and remove this `let` from here for now.

spec/decorators/case/ico/sar_decorator_spec.rb
```diff
@@ -9,7 +9,7 @@
 
   describe '#type_printer' do
     it 'pretty prints Case' do
-      expect(ico_sar_case.decorate.pretty_type).to eq 'ICO appeal for SAR case'
+      expect(ico_sar_case.decorate.pretty_type).to eq 'ICO appeal(SAR)'
```
misaka: Another place where we should have a space in front of `(SAR)` (and maybe capitalise `Appeal`).

app/decorators/case/base_decorator.rb
misaka: We should fix the `translate_for_case` used here, since the one in lib does support it (I believe!).

misaka (approved):

PR #404, opened 2018-08-28
Update Factories to conform with FactoryBot's deprecation of static a…

stephenrichards:
…ssignments

FactoryBot is deprecaring static assignments of attributes - all assignments
must now be dynamic, i.e. in  a black.  This PR replaces all static assignments with
dynamic ones.


misaka (approved):
Nice small change!

PR #405, opened 2018-08-28
CT-1891 ICO Appeal factories

aliuk2012:
- [x] adds upheld/overturned ICO appeals factories
- [x] adds upheld/overturned ICO appeals admin created cases


app/validators/closed_case_validator.rb
```diff
@@ -100,14 +100,15 @@ def validate_date_ico_decision_received(rec)
...
+    elsif !rec.ico_decision.in?( Case::ICO::Base.ico_decisions.keys )
       rec.errors.add(:ico_decision, 'invalid')
     end
   end
 
   def validate_files_or_comment(rec)
     if rec.ico_decision == 'overturned' &&
-        rec.uploaded_ico_decision_files.blank? &&
+        (rec.uploaded_ico_decision_files.blank? ||
+            rec.attachments.ico_decisions.any?) &&
```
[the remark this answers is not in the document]
aliuk2012: Yes this was needed for the factories and creating valid overturned ICO Appeals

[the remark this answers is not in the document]
misaka: 👍

lib/cts/cases/create.rb
```diff
@@ -378,6 +372,37 @@ def create_responder(responding_team)
...
+    end
+
+    def transition_to_closed_for_ico(kase)
+      kase.prepare_for_close
+      ico_decision = options.fetch(:ico_decision, Case::ICO::Base.ico_decisions.keys.sample )
+      kase.update_attributes(date_ico_decision_received: Date.today, ico_decision: ico_decision)
+      if kase.overturned?
+        uploader = S3Uploader.new(kase, CTS::dacu_manager)
+        uploader.add_file_to_case('spec/fixtures/ico_decision.png', :ico_decision)
+        # kase.attachments << FactoryBot.build(:case_ico_decision)
```
misaka: We should be able to drop this line now.

spec/factories/case/ico_fois.rb
```diff
@@ -150,14 +150,25 @@
...
+      ico_decision         { "overturned" }
+      ico_decision_comment { Faker::NewGirl.quote }
+    end
+
+    after(:create) do |kase, evaluator|
+
+      if kase.ico_decision == 'overturned'
+        kase.attachments.push(*evaluator.attachments)
+        kase.save!
+      end
```
misaka: I would definitely move this into an `after(:create)` block in the `overturned_by_ico` trait, if that works of course.

spec/factories/case_attachments.rb
```diff
@@ -52,4 +52,12 @@
     preview_key { "#{SecureRandom.hex(16)}/request_previews/" +
                   "#{upload_group}/#{Faker::Internet.slug}.pdf" }
   end
+
+  factory :case_ico_decision, parent: :case_attachment do
+    type 'ico_decision'
```
misaka: Didn't we just merge a PR to convert all such attributes from static to dynamic? (by putting them in `{...}`) (see #404)

aliuk2012: Good spot!!

lib/cts/cases/create.rb
aliuk2012: 👍

misaka (approved):

PR #408, opened 2018-08-30
CT-1843 Automatically link cases from original and appeal

stephenrichards:


beccapearce (approved):
Looks good, think you and Misha need to discuss when you merge though

PR #424, opened 2018-09-06
CT-1895 Make Data subject full name searchable for SARs

stephenrichards:


spec/models/case/sar_spec.rb
```diff
@@ -217,4 +217,12 @@
...
     end
   end
+
+  describe '.searchable_fields_and_ranks' do
+    it 'includes subject full name' do
+      expect(Case::SAR.searchable_fields_and_ranks).to include({subject_full_name: 'B'})
+    end
+  end
+
+  describe ''
```
misaka: Can be removed.

misaka (changes requested):

misaka (approved):

PR #427, opened 2018-09-07
CT-1848 update case details page to support Overturned SARs

aliuk2012:
- [x] add pretty_type method to the decorator - Used in the case page title
- [x] delegates ico decision, date decision, ico decision comment and ico decision attachments
- [x] Display the ICO decision and uploaded documents
- [x] Display the Original & related cases 
- [x] Hide "Request" section

### DIFF
--- a/app/decorators/case/ico/base_decorator.rb
+++ b/app/decorators/case/ico/base_decorator.rb
@@ -2,7 +2,23 @@ class Case::ICO::BaseDecorator < Case::BaseDecorator
 
   attr_accessor :original_case_number, :related_case_number
 
-  def date_decision_received
+  def formatted_date_ico_decision_received
     I18n.l(object.date_ico_decision_received, format: :default)
   end
+
+  def ico_decision_summary
+    Slim::Template.new do
+      <<~EOHTML
+      p
+        strong
+          = "MoJ's decision has been #{object.ico_decision} by the ICO "
+        = "on #{ self.formatted_date_ico_decision_received }"
+      
+      - if #{object.ico_decision_comment.present?}
+        p
+          = " #{object.ico_decision_comment}"
+      EOHTML
+    end.render.html_safe
+  end
+
 end
--- a/app/decorators/case/ico/base_decorator.rb
+++ b/app/decorators/case/ico/base_decorator.rb
@@ -16,7 +16,7 @@ class Case::ICO::BaseDecorator < Case::BaseDecorator
       
       - if #{object.ico_decision_comment.present?}
         p
-          = " #{object.ico_decision_comment}"
+          = "#{object.ico_decision_comment}"
       EOHTML
     end.render.html_safe
   end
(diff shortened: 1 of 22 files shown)


app/decorators/case/ico/base_decorator.rb
```diff
@@ -2,7 +2,23 @@ class Case::ICO::BaseDecorator < Case::BaseDecorator
 
   attr_accessor :original_case_number, :related_case_number
 
-  def date_decision_received
+  def formatted_date_ico_decision_received
     I18n.l(object.date_ico_decision_received, format: :default)
```
stephenrichards: why does this method need to be called formatted-date_ico_decision_received, and not just date_ico_decision_recieved?  It is in a decorator, so the whole point is to have the same attribute names.

spec/support/stub_policy.rb
```diff
@@ -0,0 +1,20 @@
+# Utility method to help us setup policy.
+#
+# Without this approach, we will have multiple versions of this method each time
+# we need to stub out the policies
+
```
stephenrichards: Nice

stephenrichards (changes requested):

app/decorators/case/ico/base_decorator.rb
aliuk2012: I seem to remember running into difficulties when it came to working out targets because it cant use the formatted version to work out if its in target or not. So thats why it was originally `date_decision_received` but I think this method name is more descriptive and easier to read.

stephenrichards: I don't quite understand what you're saying, but normally, decorators use the same method name as the original attribute, and just add formatting or prettification.  However, looking at our decorators, we don't seem to have done that anywhere. Go figure!.  Will approve.

misaka: I know that @aliuk2012 and I were looking at this together. I think @stephenrichards's point is worth considering, even though I advocated for the `formatted_...` version, the Draper docs do seem to steer towards simply using the same method name as what's being decorated and it's true that to get the undecorated version we can just use `object`.

stephenrichards (approved):

app/decorators/case/overturned_ico/sar_decorator.rb
```diff
@@ -1,7 +1,15 @@
 class Case::OverturnedICO::SARDecorator < Case::OverturnedICO::BaseDecorator
 
+  def pretty_type
+    'ICO overturned (SAR)'
+  end
+
+  def requester_name_and_type
+    pretty_type
+  end
```
misaka: Something feels a little off about this change. So far as I can tell, this just affects the display of this case on the "New Cases" page, is that right? And do we really just want to replace the requester name and type with the case type?

I can't find a design for this, or any tests, which is why I'm left wondering.

misaka: @aliuk2012 points out that these came from the ICO Appeal decorators. I do think that this should be tidied up in another PR:

- add the case type to the New Cases page for other case types (i.e. FOI and SARs don't have the case type visible on the New cases page)
- fix the name of this method, if we don't change it, as we now have the case type in here
- add tests

PR #432, opened 2018-09-10
Ct 1855 overturned on closed case list

beccapearce:


misaka:
I see that this includes a commit from CT-1853, and changes to allow overturned SAR cases to be closed. I'm going to wait to review this until CT-1853 is merged so it's clearer what changes are in here.

beccapearce:
yep mentioned in slack that we should wait for ct-1853 before reviewing

stephenrichards (approved):

PR #441, opened 2018-09-13
CT-1929 Fix ICO SAR update closure details bug

stephenrichards:
Updating the closure reasons of a closed ICO appeal for SAR
cases crashed if no ICO decision files were uploaded.  This bug
has been fixed.


stephenrichards:
Deleted because of semaphore issues

PR #450, opened 2018-09-18
Ct 1852 Accept/Reject overturned SAR cases

aliuk2012:


spec/site_prism/page_objects/sections/cases/ico/ico_decision_section.rb
```diff
@@ -0,0 +1,17 @@
+module PageObjects
+  module Sections
+    module Cases
+      module ICO
+        class IcoDecisionSection < SitePrism::Section
```
misaka: Can we keep this consistent and use `ICODecisionSection`?

app/views/shared/components/_case_page_original_case_details.html.slim
```diff
@@ -1,4 +1,4 @@
-- if @case.ico?
+- if @case.ico? || @case.overturned_ico?
```
misaka: We're using different styles of detecting whether to render bits in views. In `edit.html.slim` we use:

```
  - if @case.respond_to?(:ico_decision) && @case.ico_decision.present?
    ...
```

Which is more in-line with the duck-typing philosophy in Ruby. Feels like we should choose one style and stick with it, what do you think? If we wanted we could use duck-typing here with:

```
  - if @case.respond_to?(:original_case) && @case.original_case.present?
    ...
```

spec/site_prism/page_objects/sections/cases/ico/ico_decision_section.rb
aliuk2012: Good point

app/views/shared/components/_case_page_original_case_details.html.slim
aliuk2012: I agree with you, I completely forgot we could inspect original_case for both case types.

beccapearce (approved):

misaka (approved):

PR #455, opened 2018-09-20
Ct 1899 allow teams ot foi

beccapearce:


db/data_migrations/20180920104841_populate_teams_for_overturned_fois.rb
```diff
@@ -0,0 +1,16 @@
...
+    foi_ct = CorrespondenceType.foi
+
+    teams = BusinessUnit.all
+    teams.each do |bu|
+      if bu.correspondence_types.include?(foi_ct)
+        unless bu.correspondence_types.include?(overturned_foi_ct)
+          bu.correspondence_types << overturned_foi_ct
+          bu.save!
+        end
+      end
```
misaka: I understood that what we were going to do is re-use the FOI correspondence type so that if a BU can handle FOIs, they can handle Overturned FOIs too.

misaka (approved):

PR #457, opened 2018-09-21
Ct 1900 ot sar authorise users

beccapearce:


db/data_migrations/20180920104841_populate_teams_for_overturned_fois.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    teams = BusinessUnit.all
+    teams.each do |bu|
+      if bu.correspondence_types.include?(foi_ct)
+        unless bu.correspondence_types.include?(overturned_foi_ct)
+          bu.correspondence_types << overturned_foi_ct
+          bu.save!
+        end
+      end
+    end
```
misaka: This is part of CT-1899 isn't it?

spec/policies/cases/overturned_ico/foi_policy_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+  let(:unassigned_case)       { create :overturned_ico_foi }
+
+  let(:manager)               { managing_team.managers.first }
+  let(:other_manager)         { other_managing_team.managers.first }
+  let(:responder)             { responding_team.responders.first }
+  let(:press_officer)         { find_or_create :press_officer }
+  let(:private_officer)       { find_or_create :private_officer }
+  let(:disclosure_approver)   { dacu_disclosure.approvers.first }
+
+
```
misaka: We discussed that a test that ensures that `Case::OverturnedICO::FOIPolicy` inherits from `Case::FOI::StandardPolicy` would mean we don't need as much testing throughout this test.

misaka (approved):

PR #459, opened 2018-09-21
CT-1932 admin for ot fois

beccapearce:


beccapearce:
Change prepare_new_overturned_sar and prepare_new_overturned_foi to use a method prepare_new_overturned(with_a_type_thing)
Change new_overturned_sar_case & new_overturned_foi_case to use new_overturned_case(LALALA)

app/controllers/admin/cases_controller.rb
```diff
@@ -59,77 +59,50 @@ def count_cases_created_on(date)
...
   def prepare_new_ico
-    @correspondence_type_key = params[:correspondence_type]
-    case_creator = CTS::Cases::Create.new(Rails.logger, type: 'Case::ICO::FOI' )
-    @case = case_creator.new_case
-    @case.responding_team = BusinessUnit.responding.responding_for_correspondence_type(CorrespondenceType.ico).active.sample
-    @target_states = available_target_states
-    @selected_state = 'drafting'
-    @case.ico_decision = Case::ICO::Base.ico_decisions.keys.sample
+    prepare_new_case('Case::ICO::FOI', CorrespondenceType.foi)
     @s3_direct_post = S3Uploader.s3_direct_post_for_case(@case, 'requests')
```
misaka: I think this can also be removed since it's done on line 108 in `prepare_new_case`. Which means ... is there any custom code left for different case types at all? Or can we have the controller call `prepare_new_case` with the type as an argument, and a method that returns the class `Case::ICO::FOI` and original type ...

app/controllers/admin/cases_controller.rb
```diff
@@ -43,9 +43,10 @@ def index
   end
 
   def new
+    # binding.pry
```
stephenrichards: !?!?!?

stephenrichards (approved):

app/controllers/admin/cases_controller.rb
beccapearce: ?????????

misaka (approved):

app/controllers/admin/cases_controller.rb
misaka: ‽‽‽‽‽

PR #467, opened 2018-09-24
CT-1938 prevent further clearance ot sar

beccapearce:


spec/state_machines/workflows/overturned_sar_permitted_events.rb/standard_spec.rb
```diff
@@ -15,44 +15,41 @@
...
-                                                                      :remove_linked_case,
-                                                                      :request_further_clearance]
+                                                                      :remove_linked_case]
         end
       end
 
       context 'awaiting responder' do
         it 'should show permitted events' do
-          k = create :awaiting_responder_sar
+          k = create :awaiting_responder_ot_ico_sar
```
aliuk2012: Why are all these tests being changed from a SAR to OT SAR? Shouldn't these be additional tests? I.e For SARs they can request further clearance but OT SARs they cant.

beccapearce: This is the overturned sar test, so all the cases should be overturned sars.

stephenrichards (approved):

PR #474, opened 2018-09-25
CT-1943 fix missing pundit translations

stephenrichards:


config/locales/en.yml
```diff
@@ -171,6 +171,8 @@ en:
     case/ico/foi_policy:
       can_close_case?: You are not authorised to close this case
       can_accept_or_reject_responder_assignment?: You are not allowed to accept or reject responder assignment
+    case/ico/sar_policy:
```
beccapearce: shouldn't we have this for overturned sar?

beccapearce (changes requested):
do we have translations for overturned SARs?

beccapearce (approved):

PR #475, opened 2018-09-26
CT-1910 add params for mark as sent

beccapearce:


spec/features/cases/overturned_foi/mark_as_sent_spec.rb
```diff
@@ -0,0 +1,73 @@
+require 'rails_helper'
```
stephenrichards: Sorry to say this Becca, but I think all these should really be written as controller tests, not feature tests.  They will run much more quickly, and I think we should reserve these headless browser tests for full journeys. What do you think?

stephenrichards (changes requested):

spec/features/cases/overturned_foi/mark_as_sent_spec.rb
beccapearce: Sounds goooood!

stephenrichards (approved):

PR #481, opened 2018-09-28
CT-1937 Fix Case Linking bug

stephenrichards:
This bug was caused by a user being both a responder and a manager.
Only a manager has permissions to link SAR cases (and OVT SAR cases),
but in certain circumstances, the 'User#.teams_for_case' returns a
responding team first, which causes the error.  This only seems to happen
in production environments where the database can return records in a
different order than on the local database.

The fix is to ensure only manging teams are selected as teh acting team
in the CaseLinkingService for SAR and OVT SAR cases.


app/services/case_linking_service.rb
```diff
@@ -11,37 +14,26 @@ def initialize(user, kase, link_case_number)
...
+        @case_link = LinkedCase.new(
+            linked_case_number: @link_case_number.to_s&.strip
+        )
+        create_links
+        @result
+      rescue CaseLinkingError => err
+        @case.errors.add(:linked_case_number, err.message)
+        @error = err
+        @result = :error
+      rescue => err
```
misaka: Can we let Sentry handle these so that we actually see the error? Otherwise it'll get lost in the logs if this ever happens in production.

misaka (changes requested):
If it's quick to do, let's let raised errors go to Sentry so we get to see them.

misaka (approved):

PR #483, opened 2018-09-30
CT-1901 create ico overturned foi

misaka:


app/controllers/concerns/overturned_ico_params.rb
```diff
@@ -1,16 +1,27 @@
...
+      :original_case_id,
+      :reply_method,
+      :email,
+      :postal_address,
+      :received_date_dd, :received_date_mm, :received_date_yyyy,
+      :external_deadline_dd, :external_deadline_mm, :external_deadline_yyyy,
+    )
   end
 
+  def create_ico_overturned_sar_params
```
aliuk2012: do we need to be case type specific with the method names? They seem to use the same params.

app/views/cases/overturned_shared/_new.html.slim
```diff
@@ -1,8 +1,9 @@
-= form_for @case, as: :case_overturned_sar, url: cases_path do |f|
+= form_for @case.object, as: "case_#{@correspondence_type_key}", url: cases_path do |f|
 
-  = hidden_field_tag 'correspondence_type', 'overturned_sar'
+  = hidden_field_tag 'correspondence_type', @correspondence_type_key
   = f.hidden_field :original_ico_appeal_id
-  = f.hidden_field :original_case_id, value: @case.original_ico_appeal.original_case_id
+  = f.hidden_field :original_case_id,
```
aliuk2012: Does this have to be on the page? I'm not seeing why we need it.

app/views/cases/overturned_shared/_new_form.html.slim
```diff
@@ -7,36 +7,36 @@ section
...
-  .column-two-thirds
-    .form-group
-      = f.gov_uk_date_field :received_date, { legend_text: "Date Information Commissioner's Office (ICO) decision was received",
-              form_hint_text: '',
-              today_button: {class: 'button-secondary'} }
-
+fieldset#received_date
+  = f.hidden_field :received_date_dd
+  = f.hidden_field :received_date_mm
+  = f.hidden_field :received_date_yyyy
```
aliuk2012: why do we need this on the page? Cant we set it in the controllers action?

app/views/cases/overturned_shared/_new_form.html.slim
```diff
@@ -7,36 +7,36 @@ section
...
 
 .form-group
-  = f.gov_uk_date_field :external_deadline, { legend_text: "Final deadline" ,
-          form_hint_text: "We'll use this date to set the draft deadline, calculated as 20 working days from the final deadline."}
+  = f.gov_uk_date_field :external_deadline,
+                        { legend_text: "Final deadline" ,
+                          form_hint_text: "For example, 01 02 2018" }
+  .form-hint
```
aliuk2012: could you add a `p` in front of the class name.

app/controllers/concerns/overturned_ico_params.rb
misaka: No. This must be left-over from when we thought they'd be more separate. I'll look at merging these.

app/views/cases/overturned_shared/_new.html.slim
misaka: These were just left-over from before, but you're right ... if we have the `original_ico_appeal`, we don't need and shouldn't have `original_case_id` so there's no risk of having them out-of-sync.

app/views/cases/overturned_shared/_new_form.html.slim
misaka: I just wanted to minimize changes and thought that since we handle it this way already ... will remove to clean up the form. As discussed, it doesn't make sense to allow editing of `received_date`.

app/views/cases/overturned_shared/_new_form.html.slim
misaka: Yup!

app/controllers/concerns/overturned_ico_params.rb
misaka: Ah, I see now. The issue is that one requires `case_overturned_sar` and the other `case_overturned_foi` as the type of case params (see line 17). I'll extract the param list to one place but I think that's the best we can reasonably do.

app/controllers/concerns/overturned_ico_params.rb
```diff
@@ -5,33 +5,45 @@ def create_ico_overturned_foi_params
...
+    )
   end
 
   private
 
   def create_overturned_ico_params_list
     [
       :original_ico_appeal_id,
-      :original_case_id,
+      # :original_case_id,
```
aliuk2012: remove?

app/controllers/concerns/overturned_ico_params.rb
```diff
@@ -5,33 +5,45 @@ def create_ico_overturned_foi_params
...
       :reply_method,
       :email,
       :postal_address,
-      :received_date_dd, :received_date_mm, :received_date_yyyy,
+      # :received_date_dd, :received_date_mm, :received_date_yyyy,
```
aliuk2012: remove?

aliuk2012 (approved):

config/locales/en.yml
```diff
@@ -399,6 +403,20 @@ en:
           upheld: Upheld by ICO
           overturned: Overturned by ICO
         ico_decision_comment: "Details of ICO's decision"
+      case_overturned_foi:
```
beccapearce: Can case_overturned_foi and case_overturned_sar inherit from each other? To stop the duplication

spec/services/case_create_service_spec.rb
```diff
@@ -25,189 +25,206 @@
...
+              'received_date_dd'        => received.day.to_s,
+              'received_date_mm'        => received.month.to_s,
+              'received_date_yyyy'      => received.year.to_s,
+              'external_deadline_dd'    => deadline.day.to_s,
+              'external_deadline_mm'    => deadline.month.to_s,
+              'external_deadline_yyyy'  => deadline.year.to_s,
+              'internal_deadline_dd'    => internal_deadline.day.to_s,
+              'internal_deadline_mm'    => internal_deadline.month.to_s,
+              'internal_deadline_yyyy'  => internal_deadline.year.to_s,
+              'message'                 => 'AAAAA'
```
beccapearce: I find this derivative

config/locales/en.yml
misaka: Yes!

spec/services/case_create_service_spec.rb
misaka: That's just the sound I make as I read your comment. With a few more exclamation marks, though.

beccapearce (approved):

PR #488, opened 2018-10-03
Order admin users page by name

beccapearce:
Super quick PR to help us find users from the admin page

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -3,7 +3,10 @@ class Admin::UsersController < AdminController
     if @team.present?
       @users = @team.users
     else
-      @users = User.all.page(params[:page]).per(50).decorate
+      @users = User.all
+                   .order(:full_name)
+                   .page(params[:page]).per(100)
+                   .decorate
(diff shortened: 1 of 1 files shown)


app/views/admin/users/index.html.slim
```diff
@@ -35,5 +35,3 @@
                   - user.cases.opened.each do |kase|
                     li
                       = link_to kase.number, case_path(kase.id)
-
-= paginate @users
```
aliuk2012: Cant we just leave the pagination and just increase the number of of results? Just thinking listing over 700+ users and all their cases on a single page might put the server under a bit of pressure

beccapearce: Sure! Not a problem! Might be nice to keep it fairly high though, I'm just thinking of using find on the page as a quick way to track down users.

How many do you think? 100? 200? 100000000000000 ;)

aliuk2012: 1million should just about do it :P 

Feature test?  ;)

aliuk2012 (approved):

PR #489, opened 2018-10-03
CT-1901 add "Back to case details" link to ICO OT create page

misaka:


spec/views/cases/overturned_shared/new_html_slim_spec.rb
```diff
@@ -97,4 +97,13 @@ def login_as(user)
             )
     end
   end
+
+  describe 'back to case details link' do
+    it 'exists on the page', js: true do
```
aliuk2012: Why do we need `js:true' ?

misaka: Removed.

aliuk2012 (approved):

PR #498, opened 2018-10-09
Bumps ruby 2.3.1 to ruby 2.5.1

aliuk2012:
# To do:
Fix the following failing specs:
```
rspec ./spec/state_machines/configurable_state_machine/config_validator_spec.rb:183 # ConfigValidator validation of a single case type permitted workflows is a Fixnum errors
rspec ./spec/state_machines/configurable_state_machine/config_validator_spec.rb:204 # ConfigValidator validation of a single case type validation of workflows workflows is not a hash errors
rspec ./spec/state_machines/configurable_state_machine/config_validator_spec.rb:254 # ConfigValidator validation of a single case type validation of workflows user roles is not a hash errors
```

# COMPLETED and working seeing comments below


aliuk2012:
The reason these 3 specs were failing was because the spec was expecting an error message saying the its not a valid Fixnum. Fixnums have been depreciated and removed since ruby 2.4 (see my commit)

aliuk2012:
Deployed and tested on `demo` if you would like to test it.

Once this PR is merged everyone will need to update their Ruby version and https://semaphoreci.com/correspondence-tool/correspondence_tool_staff-4/settings will needs to be updated too

stephenrichards (approved):

PR #501, opened 2018-10-09
CT-1962 Fix bug password reset bug

stephenrichards:
Users who had locked themselves out (by entering the wrong password too many times)
were receiving the de-activated account mail when requesting a new password.  This
is wrong, and that mail notification has now been removed.


spec/controllers/passwords_controller_spec.rb
```diff
@@ -17,10 +17,11 @@
     before(:each)  { @request.env["devise.mapping"] = Devise.mappings[:user] }
 
     context 'user is deactivated' do
-      it 'sends an email and shows message' do
+      it 'sends reset password mail and not account inactive mail' do
         mailer = double("mailer")
-        expect(DeviseMailer).to receive(:account_not_active).and_return(mailer)
-        expect(mailer).to receive(:deliver_later)
+        expect(DeviseMailer).not_to receive(:account_not_active)
```
aliuk2012: This method no longer exists

spec/controllers/passwords_controller_spec.rb
```diff
@@ -17,10 +17,11 @@
     before(:each)  { @request.env["devise.mapping"] = Devise.mappings[:user] }
 
     context 'user is deactivated' do
-      it 'sends an email and shows message' do
+      it 'sends reset password mail and not account inactive mail' do
```
aliuk2012: I dont think we need to reference 'inactive mail' as that no longer exists

aliuk2012 (changes requested):
Only a couple minor things need changing

aliuk2012 (approved):

PR #503, opened 2018-10-09
CT-1969 Create trigger overturned FOI

stephenrichards:


misaka (changes requested):
I don't see feature or view tests, I think we should have one or both.

spec/features/cases/overturned_sar/create_case_spec.rb
```diff
@@ -45,6 +45,50 @@
...
+    assignments_new_page.choose_business_group(responding_team.business_group)
+
+    # Select Business Unit
+    assignments_new_page.choose_business_unit(responding_team)
+  end
+
+  scenario 'creating an OVT FOI case', js: true do
+    ico_case = create(:closed_ico_foi_case)
+
+    # Move this to a user journey test once displaying works.
```
misaka: I don't think this comment is relevant anymore, is it?

spec/features/cases/overturned_sar/create_case_spec.rb
```diff
@@ -45,6 +45,50 @@
...
+    click_button 'Create case'
+
+    # Browse Business Group
+    assignments_new_page.choose_business_group(responding_team.business_group)
+
+    # Select Business Unit
+    assignments_new_page.choose_business_unit(responding_team)
+  end
+
+  scenario 'creating an OVT FOI case', js: true do
```
misaka: This file is for overturned SARs.

spec/features/cases/overturned_sar/create_case_spec.rb
```diff
@@ -45,6 +45,50 @@
...
+
+    # Move this to a user journey test once displaying works.
+    # create_and_assign_overturned_sar_case(user: manager,
+    #                                       ico_case: ico_case,
+    #                                       responding_team: responding_team)
+
+    cases_show_page.load(id: ico_case.id)
+    expect(cases_show_page).to be_displayed(id: ico_case.id)
+    # Replace the following-line with a click on the "New overturned case"
+    # button when available
```
misaka: I'm pretty sure we have a user journey that tests that the button comes up on the ICO page, etc, so we can probably get rid of the bits that load the ICO page first and related comments.

misaka (changes requested):
Feature test should be in a file specific to ICO OVT FOIs, and can have some stuff trimmed out.

spec/features/cases/overturned_sar/create_case_spec.rb
stephenrichards: The comment is relevant - but the test isn't - it is covered in the e2e user journey test `spec/features/user_journeys/overturned_foi/ovt_foi_non_trigger_e2e_spec.rb`.  Have removed this spec.

misaka (approved):

PR #514, opened 2018-10-15
CT-2007 WIP original mega-commit

misaka:
- cut down on massive duplication of all fixture data by creating some known
  singular fixtures such as foi_responder, sar_responder, foi_responding_team,
  sar_responding_team etc which allow all cases to have the same responder, etc,
  by default
- cut down duplication in cases and case transitions
- a lot of other fixes


misaka:
See comments on ticket if you're at all curious. But it ain't pretty.

stephenrichards (approved):

PR #528, opened 2018-10-19
CT-1931 Draft Deadline error message copy change

stephenrichards:


misaka (changes requested):
Will you amend the commit to have the right number? So it track correctly in the release, etc.

misaka (approved):

PR #529, opened 2018-10-19
CT-2013 State machine bugfix

stephenrichards:
users with multiple roles were not able to add messages if their first role
was prvented from doing so by a predicate returning false, even though
subsequent roles allowed them to.

This has been fixed by evaluating the configs for all roles, and returning true
if any of them resolve to true

Tests for this edgecase have been added

### DIFF
--- a/app/state_machines/configurable_state_machine/machine.rb
+++ b/app/state_machines/configurable_state_machine/machine.rb
@@ -48,8 +48,34 @@ module ConfigurableStateMachine
     #
 
     def can_trigger_event?(event_name:, metadata:, roles: nil)
-      config = config_for_event(event_name: event_name, metadata: metadata, roles: roles)
-      config.present?
+      configs = configs_for_event(event_name: event_name, metadata: metadata, roles: roles)
+      results = configs.map{ |config| event_triggerable_in_config?(config: config, user: @user ) }
+      results.include?(true)
+    end
+
+    def configs_for_event(event_name:, metadata:, roles:)
+      set_users_and_roles(metadata: metadata, roles: roles)
+      configs = []
+      @roles.each do |role|
+        role_config = @config.user_roles[role]
+        next if role_config.nil?
+        role_state_config =  role_config.states[@kase.current_state]
+        event_config = fetch_event_config(config: role_state_config, event: event_name)
+        configs << event_config
+      end
+      configs
+    end
+
+    def event_triggerable_in_config?(config:, user:)
+      if config.nil?
+        false
+      elsif config.if.nil?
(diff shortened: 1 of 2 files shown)


app/state_machines/configurable_state_machine/machine.rb
```diff
@@ -48,8 +48,34 @@ def permitted_events(current_user_or_id)
     #
 
     def can_trigger_event?(event_name:, metadata:, roles: nil)
-      config = config_for_event(event_name: event_name, metadata: metadata, roles: roles)
-      config.present?
+      configs = configs_for_event(event_name: event_name, metadata: metadata, roles: roles)
+      results = configs.map{ |config| event_triggerable_in_config?(config: config, user: @user ) }
```
misaka:
```suggestion
      configs.map{ |config| event_triggerable_in_config?(config: config, user: @user ) }.any?
```

app/state_machines/configurable_state_machine/machine.rb
```diff
@@ -48,8 +48,34 @@ def permitted_events(current_user_or_id)
...
+      results.include?(true)
+    end
```
misaka: :100:

app/state_machines/configurable_state_machine/machine.rb
```diff
@@ -48,8 +48,34 @@ def permitted_events(current_user_or_id)
...
+      elsif config.if.nil?
+        true
+      elsif predicate_is_true?(predicate: config.if, user: user)
+        true
+      else
+        false
+      end
     end
 
     def config_for_event(event_name:, metadata:, roles: nil)
```
misaka: Is this function useful as it stands? Is it being used anywhere else? If anything, it could be changed to return the first config that `event_triggerable_in_config` returns true for, that was the original intent for `config_for_event` but I think something may have got lost in a rebasing.

misaka: Just a thought:

```
def config_for_event(event_name:, metadata:, roles: nil)
  configs_for_event(event_name:event_name, metadata:metadata, roles:roles).detect { |c| event_triggerable_in_config?(config: c, user: @user) }
end
```

stephenrichards: yes, it's being used in the after hook matcher.

misaka (changes requested):
I say that if `config_for_event` isn't being used it can be removed, or if it is, or we just like the idea of it, it be changed to use `configs_for_event` as per the suggestions.

misaka (approved):

PR #555, opened 2018-10-29
CT-1990 Re-assign Trigger OVT SAR case to a new team

stephenrichards:


spec/state_machines/overturned_sar_event_spec.rb
```diff
@@ -16,7 +16,12 @@ def all_cases
         :ot_ico_sar_noff_unassigned,
         :ot_ico_sar_noff_awresp,
         :ot_ico_sar_noff_draft,
-        :ot_ico_sar_noff_closed
+        :ot_ico_sar_noff_closed,
+        :ot_ico_sar_noff_trig_awresp,
+        :ot_ico_sar_noff_trig_awresp_accepted,
+        :ot_ico_sar_noff_trig_draft,
+        :ot_ico_sar_noff_trig_draft_accepted,
+        :ot_ico_sar_noff_pdacu
```
beccapearce: I think we're missing a case in awaiting dispatch.

spec/state_machines/overturned_sar_event_spec.rb
```diff
@@ -95,6 +133,29 @@ def all_cases
...
+
+        [:another_disclosure_specialist, :ot_ico_sar_noff_pdacu],
+
+        [:responder, :ot_ico_sar_noff_pdacu],
+
+        [:another_responder_in_same_team, :ot_ico_sar_noff_pdacu],
+
+        [:another_responder_in_diff_team, :ot_ico_sar_noff_pdacu],
+
+        [:press_officer, :ot_ico_sar_noff_pdacu],
```
beccapearce: should press and private officers be able to link cases, since it's a SAR case I would expect that they wouldn't be able to view it

spec/state_machines/overturned_sar_event_spec.rb
```diff
@@ -123,6 +205,28 @@ def all_cases
...
```
beccapearce: same as above I wouldn't expect to see press and private here

beccapearce (changes requested):

beccapearce (approved):

PR #557, opened 2018-10-30
CT-1994 add notification for progress to disclosure allow anyone in responding team to progress

beccapearce:


app/controllers/cases_controller.rb
```diff
@@ -610,7 +610,7 @@ def progress_for_clearance
     authorize @case
 
     @case.state_machine.progress_for_clearance!(acting_user: current_user,
-                                                acting_team: @case.team_for_assigned_user(current_user, :responder),
+                                                acting_team: @case.team_for_unassigned_user(current_user, :responder),
```
misaka: We should have a test for this, no?

misaka (changes requested):
We should have a controller test.

misaka (approved):

PR #558, opened 2018-10-30
 Ct 2017 upload response awaiting dispatch trigger

beccapearce:
Since we introduced the new state machine it seems unnecessary to have two events for uploading responses (add_responses and add_response_to_flagged_case).

This PR  changes the system to only use add_responses.

This was prompted by a bug where users could not upload responses to trigger cases in awaiting dispatch.

### DIFF
--- a/app/services/response_uploader_service.rb
+++ b/app/services/response_uploader_service.rb
@@ -58,11 +58,6 @@ class ResponseUploaderService
                                            acting_team: @case.responding_team,
                                            filenames: filenames,
                                            message: @case.upload_comment)
-      # when 'upload-flagged'
-      #   @case.state_machine.add_responses!(acting_user: @current_user,
-      #                                      acting_team: @case.responding_team,
-      #                                      filenames: filenames,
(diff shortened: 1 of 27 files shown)


app/services/response_uploader_service.rb
```diff
@@ -53,16 +53,16 @@ def transition_state(response_attachments)
...
-      when 'upload-flagged'
-        @case.state_machine.add_response_to_flagged_case!(acting_user: @current_user,
-                                                          acting_team: @case.responding_team,
-                                                          filenames: filenames,
-                                                          message: @case.upload_comment)
+      # when 'upload-flagged'
+      #   @case.state_machine.add_responses!(acting_user: @current_user,
+      #                                      acting_team: @case.responding_team,
+      #                                      filenames: filenames,
+      #                                      message: @case.upload_comment)
```
misaka: Is this safe to remove?

misaka (approved):
There's some commented-out code that can be removed, but can be merged if necessary.

PR #559, opened 2018-10-30
ct-2020 hide update closure on trigger ot sar

beccapearce:


config/state_machine/configs/00_moj/20_sar/30_sar_trigger_workflow.yml
```diff
@@ -71,6 +71,7 @@
                 edit_case:
                   if: Workflows::Predicates#not_overturned?
                 update_closure:
+                  if: Workflows::Predicates#not_overturned?
```
misaka: Test? Test! Or ... make predicated by a feature flag!

misaka (changes requested):
We can make the predicate a feature flag, which would better communicate why the edit/update feature is disabled.

config/state_machine/configs/00_moj/20_sar/30_sar_trigger_workflow.yml
beccapearce: Feature flag is the way forward!

misaka (approved):

PR #570, opened 2018-11-05
Ct 2023 deactivate directorates

beccapearce:


app/controllers/teams_controller.rb
```diff
@@ -133,7 +133,8 @@ def destroy
       flash[:notice] = I18n.t('teams.destroyed')
       redirect_to(set_destination(@team))
     else
-      render action: :show
+      flash[:alert] = I18n.t('teams.error')
+      redirect_to team_path(@team)
```
stephenrichards: I think this message is being displayed whatever the error is, not just active children

config/locales/en.yml
```diff
@@ -780,11 +780,17 @@ en:
...
       new_team_member: New team member
+    directorate_detail:
+      destroy: Are you sure you want to deactivate this directorate?
     edit:
       edit: Edit %{team_type}
     new:
       new: New %{team_type}
     destroyed: This team has now been deactivated
+    error: To deactivate this team you need to first deactivate all members.
+    deactivate_info:
```
stephenrichards: See comment above: the message gets displayed for all errors, not just active children.

stephenrichards (changes requested):

stephenrichards (approved):

PR #573, opened 2018-11-07
CT-2021 Enable download of cases via CSV file

stephenrichards:


db/seeders/search_test_data_seeder.rb
```diff
@@ -161,9 +161,10 @@ def set_flagged
   end
 
   def select_responding_team
-    @responding_teams ||= BusinessUnit.responding.active.map { |team| team.id.to_s}
-
-    @responding_teams.to_a[ @case_count % @responding_teams.length]
+    # @responding_teams ||= BusinessUnit.responding.active.map { |team| team.id.to_s}
+    #
+    # @responding_teams.to_a[ @case_count % @responding_teams.length]
+    '16'
```
beccapearce: Can we uncomment the code and leave it to select a random Buisness unit

beccapearce (approved):

beccapearce (changes requested):
Just changes to the search data seeder that should be as they were

beccapearce (approved):

PR #575, opened 2018-11-09
Ct 2024 remove pit extension

beccapearce:


app/models/case/base.rb
```diff
@@ -660,6 +660,10 @@ def current_team_and_user_resolver
     CurrentTeamAndUser::Base.new(self)
   end
 
+  def extended_for_pit?
+    currently_extended_for_pit?
+  end
```
stephenrichards: shouldn't currently_extended_for_pit? be renamed as extended_for_pit? and replace this method?

app/models/case/base.rb
```diff
@@ -735,8 +739,15 @@ def update_deadlines
     end
   end
 
-  def extended_for_pit?
-    transitions.where(event: 'extend_for_pit').any?
+  def currently_extended_for_pit?
+    pit_related_transitions = transitions.where(event: ['extend_for_pit',
+                                                        'remove_pit_extension'])
+                                          .reorder(:created_at)
```
stephenrichards: reorder?

app/services/case_remove_pit_extension_service.rb
```diff
@@ -0,0 +1,34 @@
...
+      @case.update external_deadline: (find_original_final_deadline)
+      @case.reload
+      @result = :ok
+    end
+    @result
+  rescue => err
+    Rails.logger.error err.to_s
+    Rails.logger.error err.backtrace.join("\n\t")
+    @error = err
+    @result = :error
```
stephenrichards: This will log details of the error, but give no other indication of when it happens.  Better to raise err and the end of the rescue block.

app/services/case_remove_pit_extension_service.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  private
+
+  def find_original_final_deadline
+    first_pit_extension = @case.transitions.where(event: 'extend_for_pit')
+                                .reorder(:created_at)
```
stephenrichards: reorder?

stephenrichards (changes requested):

stephenrichards (approved):

PR #587, opened 2018-11-16
CT-2077 WIP: stream generated CSV for closed cases

misaka:


starswan:
This is a complex version of CT-2140 - which uses an Enumerator to achieve the same effect

starswan:
This has been replaced by CT-2140 a much simpler version. It was several N+1 query problem causing the issues (although CT-2140 uses a lazy enumerator to help)

PR #598, opened 2018-11-22
CT-2060 Upload during escalation period bugfix

stephenrichards:
REsponders should not be permitted to upload a response until 3
days (the escalation perios) has elapsed to give approvers enough
time to view the dase and take it on if they deem necessary.

There was a bug which allowed them to upload the response during
that escalation period.  This PR fixes that bug.

### DIFF
--- a/app/state_machines/workflows/predicates.rb
+++ b/app/state_machines/workflows/predicates.rb
@@ -122,7 +122,7 @@ class Workflows::Predicates
     @kase.extended_for_pit?
   end
 
-  def case_respondable?   # member of assigned team and outstide escalation period
+  def assigned_team_member_and_case_outside_escalation_period?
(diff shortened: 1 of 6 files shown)


app/state_machines/workflows/predicates.rb
```diff
@@ -122,6 +122,10 @@ def has_pit_extension?
     @kase.extended_for_pit?
   end
 
+  def case_respondable?   # member of assigned team and outstide escalation period
```
beccapearce: Could we name this something like assigned_team_member_outside_escalation_period? I know it's long but it makes the requirement much more clear

beccapearce (changes requested):
I think the method name could be more clear

beccapearce (approved):

PR #599, opened 2018-11-22
CT-2027 Recording BU responsible for lateness

stephenrichards:


app/controllers/cases_controller.rb
```diff
@@ -41,6 +41,7 @@ class CasesController < ApplicationController
                   :approve_response_interstitial,
                   :assign_to_new_team,
                   :close,
+                  :closure_outcomes,
```
misaka: This and the other actions here should be moved to `set_case`, and @case should be decorated as a separate step.

config/locales/en.yml
```diff
@@ -252,23 +254,26 @@ en:
         team_id: Assign to
         user_id: Choose a new team member for this case
       case_foi:
-        info_held_status_abbreviation: Was the information held?
+        info_held_status_abbreviation: Was the information held?1
```
misaka: Type here.

app/controllers/cases_controller.rb
```diff
@@ -340,6 +343,40 @@ def close
...
+    if !@case.update respond_date_params
+      render :close
+    else
+      @team_collection = CaseTeamCollection.new(@case)
+      @case.update(late_team_id: @case.responding_team.id)
+      render :closure_outcomes
+    end
+  end
+
+  def closure_outcomes
```
misaka: Is this action being used anywhere? There's no route for it and it doesn't work when visiting it via the browser.

app/views/cases/close.html.slim
```diff
@@ -18,6 +18,4 @@
 
 div class="case-#{@case.type_abbreviation.downcase}"
 
-  = render partial: "cases/#{@case.type_abbreviation.downcase}/close_form",
-           locals: { kase: @case,
-                     submit_button: t('.submit')}
+  = render partial: "cases/shared/date_responded_form", locals: { kase: @case, submit_button: t('.submit')}
```
misaka: Probably need to remove `_close_form.html.slim`.

spec/features/user_journeys/ico/trigger_e2e_journey_spec.rb
```diff
@@ -30,46 +30,48 @@
...
-                      user: disclosure_specialist,
-                      expected_status: 'Awaiting ICO decision',
-                      expected_to_be_with: 'Disclosure'
-
-    close_ico_appeal_case kase: kase,
-               user: manager,
-               timeliness: 'in time',
-               decision: 'upheld'
-  end
+  ################### temporarily commented out - re-isnstate before PR ####################
```
misaka: Don't forget this one!

app/services/mark_response_as_sent_service.rb
```diff
@@ -0,0 +1,45 @@
...
+    @kase = kase
+    @params = params
+    @user = user
+    @result = nil
+  end
+
+  def call
+    ActiveRecord::Base.transaction do
+      @kase.prepare_for_respond
+      @kase.ico? ? respond_to_ico : respond_to_non_ico
```
misaka: Aren't SAR cases like ICOs here, i.e. they get responded + closed in a single action?

misaka: And ... does this really need to be separated out in the functions below?

app/views/cases/close.html.slim
beccapearce: I think this is used for the edit closure reasons form, I'll see how easy it is to do editting in this ticket but it might be worth splitting out this work

app/services/mark_response_as_sent_service.rb
beccapearce: ICO cases I think are different, they go approved (awaiting_dispatch) to responded (awaiting_ico_decision), to closed

beccapearce: No sure about merging the methods, it makes sense but having tried it out, it is quite tricky to head you head around

config/locales/en.yml
```diff
@@ -701,15 +715,15 @@ en:
...
-        uploaded_request_files: Upload relevant files (optional)
+
     select_type:
       heading: Choose correspondence type
       sub_heading: Create case
+    shared:
+      date_responded_form:
+        close_date:  Date response sent
+        date_example: For example, 30 1 2017
+        date_ico_decision_received_date: DATE ICO DECISION RECEIVED
```
beccapearce: Change this to something better

beccapearce (approved):

PR #609, opened 2018-11-30
CT-2044 add YARD w/rake task

misaka:
Opted to use YARD instead of RDoc. It's output looks much prettier and it provides support for adding more semantic information to the documentation. See https://www.rubydoc.info/gems/yard/file/README.md#Feature_List


README.md
```diff
@@ -30,6 +30,16 @@ $ git clone git@github.com:ministryofjustice/correspondence_tool_staff.git
 $ cd correspondence_tool_staff
 ```
 
+#### Generating Documentation
```
beccapearce: can we add in the command to view documentation locally?

beccapearce (changes requested):

README.md
misaka: Added.

beccapearce (approved):

PR #612, opened 2018-12-03
Ct 2022 draft timeliness

beccapearce:


spec/controllers/cases_controller/execute_request_amends_spec.rb
```diff
@@ -5,6 +5,9 @@
   let(:private_officer)                { find_or_create :private_officer }
   let(:service)                        { instance_double(CaseRequestAmendsService,
                                                          call: true) }
+    let(:something)                        { instance_double(SetDraftTimelinessService,
```
misaka: Naming is hard ... ;)

app/services/case_request_amends_service.rb
```diff
@@ -2,10 +2,11 @@ class CaseRequestAmendsService
   attr_accessor :result
   attr_accessor :error
 
-  def initialize(user:, kase:, message:)
+  def initialize(user:, kase:, message:, compliance:)
```
misaka: Minor thing, but is there a reason this shouldn't be `compliant`?

app/services/response_uploader_service.rb
```diff
@@ -60,13 +60,18 @@ def transition_state(response_attachments)
...
+        SetDraftTimelinessService.new(kase: @case).call
+
       when 'upload-redraft'
         @case.state_machine.upload_response_and_return_for_redraft!(
                              acting_user: @current_user,
                              acting_team: @case.approving_teams.with_user(@current_user).first,
                              message: @case.upload_comment,
                              filenames: filenames
         )
+        if @compliance == 'yes'
```
misaka: `@compliance` isn't being set in this file anywhere. WIP?

beccapearce:
This needs to be made into a GOV UK date to enable editing later on

spec/controllers/cases_controller/approve_spec.rb
```diff
@@ -7,21 +7,41 @@
                                                  call: true,
                                                  result: :ok) }
 
-  describe 'PATCH execute_response_approval' do
+  describe 'approve' do
+    before do
+      sign_in approver
+      # allow(CaseApprovalService).to receive(:new).and_return(service)
```
mseedat-moj: TODO?

mseedat-moj (changes requested):
Initial run through regarding syntax with appropriate comments/suggestions. Issues to consider (general discussion):

1. should we use translations by default for all output strings e.g.  in cases_controller.rb?
2. cases_controller.rb is getting very long and not very 'Restful'. What should the goal be in terms of controller organisation?
3. General consistency for using line-breaks e.g. after `authorize` and to separate logical steps in a function
4. Spec/test organisation e.g. `draft_timeliness_spec.rb` be useful to organise `it...` into scenarios/contexts which describe which behaviour is expected in which context/why.

A really big piece of functionality, very thorough and well thought out.

app/controllers/cases_controller.rb
```diff
@@ -278,37 +285,161 @@ def destroy
...
+    authorize @case
+    @s3_direct_post = S3Uploader.s3_direct_post_for_case(@case, 'responses')
+    @approval_action = 'approve'
+    @case = @case.decorate
+  end
+
+  def upload_response_and_approve_action
+    authorize @case, :upload_response_and_approve?
+
+    service_params = {
```
starswan: This doesn't pass the 'compliant' flag

app/controllers/cases_controller.rb
```diff
@@ -21,7 +23,15 @@ class CasesController < ApplicationController
...
                   :remove_pit_extension,
                   :record_late_team,
-                  :update_closure
+                  :update_closure,
+                  :response_upload_and_approve,
+                  :response_upload_for_redraft,
+                  :upload_responses,
+                  :upload_responses_action,
+                  :upload_response_and_approve,
+                  :execute_upload_response_and_approve,
```
mseedat-moj: `:execute_upload_response_and_approve` should be further up the list to maintain the alphabetical ordering

app/controllers/cases_controller.rb
```diff
@@ -21,7 +23,15 @@ class CasesController < ApplicationController
...
+                  :upload_response_and_return_for_redraft,
+                  :upload_response_and_return_for_redraft_action,
```
mseedat-moj: Should this action name now be `execute_upload_response_and_return_for_redraft` or is it a special case of an action? Which is more sensible in terms of the existing code base ?

app/controllers/cases_controller.rb
```diff
@@ -278,37 +285,173 @@ def destroy
...
+        flash[:notice] = t('notices.case/ico.case_cleared')
+      else
+        flash[:notice] = t('notices.case_cleared', team: current_team.name,
+                                                   status: I18n.t("state.#{@case.current_state}").downcase)
+      end
+      redirect_to case_path(@case)
+    else
+      flash[:alert] = case_approval_service.error_message
+      @case = @case.decorate
+      render :approve
```
mseedat-moj: Is it possible to do `render :approve, alert: case_approval_service.error_message` or does that only work for `:notice` and `:error` ?

app/controllers/cases_controller.rb
```diff
@@ -278,37 +285,173 @@ def destroy
...
-      @case, current_user, bypass_params_manager, flash[:action_params]
+      kase: @case,
+      current_user: current_user,
+      action: 'upload',
+      uploaded_files: params[:uploaded_files],
+      upload_comment: params[:upload_comment],
+      is_compliant: false,
+      bypass_further_approval: false,
+      bypass_message: nil
     )
```
mseedat-moj: This class has a lot of parameters! I know it's something you've picked up on @starswan - any scope to either reduce or re-jig the code?

app/controllers/cases_controller.rb
```diff
@@ -278,37 +285,173 @@ def destroy
...
     rus.upload!
+
+    @s3_direct_post = S3Uploader.s3_direct_post_for_case(@case, 'responses')
+
+    case rus.result
+    when :blank
+      flash.now[:alert] = t('alerts.response_upload_blank?')
+      render :upload_responses
+    when :error
```
mseedat-moj: As above comment, possible to use shorthand i.e. `render :upload_responses, alert: ....` - also is the `.now` needed? if so a comment would help understanding.

app/controllers/cases_controller.rb
```diff
@@ -278,37 +285,173 @@ def destroy
...
+      action: 'upload-approve',
+      uploaded_files: params[:uploaded_files],
+      upload_comment: params[:upload_comment],
+      is_compliant: true,
+      bypass_message: nil,
+      bypass_further_approval: false
+    }
+    if params.key?(:bypass_approval)
+      service_params.merge!(
+        bypass_message: params[:bypass_approval][:bypass_message],
```
mseedat-moj: Is there is a possibility `params[:bypass_approval][:bypass_message]` could throw an exception if `params[:bypass_approval]` has a nil value? If so is it worth using something like `params.dig(:bypass_approval, :bypass_message)` ? Too defensive?

app/controllers/cases_controller.rb
```diff
@@ -278,37 +285,173 @@ def destroy
...
+      @s3_direct_post = S3Uploader.s3_direct_post_for_case(@case, 'responses')
+      flash.now[:alert] = t('alerts.response_upload_error')
+      render :upload_response_and_approve
+    when :ok
+      flash[:notice] = t('notices.response_uploaded')
+      set_permitted_events
+      redirect_to case_path @case
+    end
+  end
+  #rubocop:enable Metrics/MethodLength
```
mseedat-moj: Like the case statement, very neat.

app/controllers/cases_controller.rb
```diff
@@ -278,37 +285,173 @@ def destroy
...
+
+  def upload_response_and_return_for_redraft
+    authorize @case
     @s3_direct_post = S3Uploader.s3_direct_post_for_case(@case, 'responses')
```
mseedat-moj: Line break after `authorize @case`

app/controllers/cases_controller.rb
```diff
@@ -278,37 +285,173 @@ def destroy
...
+      is_compliant: params[:draft_compliant] == 'yes',
+      bypass_message: nil,
+      bypass_further_approval: false
+    )
+    rus.upload!
+
+    @case = @case.decorate
     case rus.result
     when :blank
+      @s3_direct_post = S3Uploader.s3_direct_post_for_case(@case, 'responses')
```
mseedat-moj: There's a few uses of the strings `'approve'` and `'redraft'` etc. I feel a bit nervous about them as the codebase grows and more developers join. Should be Constants/Enums or something a bit more robust IMO. Too big a job for this PR?

app/controllers/cases_controller.rb
```diff
@@ -554,40 +695,19 @@ def flag_for_clearance
...
     authorize @case
-    CaseRequestAmendsService.new(user: current_user, kase: @case, message: params[:case][:request_amends_comment]).call
-    if @case.type_abbreviation == 'SAR'
+    CaseRequestAmendsService.new(
+        user: current_user,
+        kase: @case,
+        message: params[:case][:request_amends_comment],
+        is_compliant: params[:case][:draft_compliant] == 'yes').call
+    if @case.sar?
       flash[:notice] = 'Information Officer has been notified a redraft is needed.'
```
mseedat-moj: Old code - but would prefer these flash strings to be moved into the translation file considering the rest of the function has changed anyway.

app/controllers/cases_controller.rb
```diff
@@ -799,22 +919,6 @@ def filter_params
     )
   end
 
-  def search_and_filter(full_list_of_cases = nil)
```
mseedat-moj: This was the unused dormant method right?

app/services/response_uploader_service.rb
```diff
@@ -48,10 +60,15 @@ def upload!
 
   private
 
+  #  when approving, always log compliance date because it must be
```
mseedat-moj: Extra space, `When...` has leading space.

mseedat-moj (changes requested):
Much improved some minor suggestions/questions.

app/controllers/cases_controller.rb
mseedat-moj: After discussion with @starswan it transpires that Line 313 does not actually have the desired effect. Should be `flash.now[:alert]`

app/controllers/cases_controller.rb
mseedat-moj: After discussion with @starswan the `.now` is needed to ensure the flash message is displayed in the current `render` as per documentation https://guides.rubyonrails.org/action_controller_overview.html#the-flash

app/controllers/cases_controller.rb
mseedat-moj: Out of scope for this PR but something to consider.

mseedat-moj (approved):

app/controllers/cases_controller.rb
starswan: I think you mean line 317.

app/controllers/cases_controller.rb
starswan: Yes this is the unused method that doesn't work

app/models/user.rb
```diff
@@ -174,7 +174,7 @@ def permitted_correspondence_types
 
   def other_teams_names(current_team)
     self.teams.delete(current_team)
-    self.teams.map(&:name).to_sentence
+    self.teams.order(:id).map(&:name).to_sentence
```
starswan: not clear why this change was made?

app/controllers/cases_controller.rb
mseedat-moj: OOS for this ticket

app/models/user.rb
mseedat-moj: Roll back this code as it is not clear why this was added (intent)

app/models/concerns/draft_timeliness.rb
```diff
@@ -0,0 +1,25 @@
+module DraftTimeliness
+  module ResponseAdded
+    extend ActiveSupport::Concern
+
+    def log_compliance_date
```
mseedat-moj: Method should have exclamation mark `!` in name i.e. `log_compliance_date!` to ensure user of the method knows this method raises an exception on failure. Applies to both instances of `log_compliance_date`

app/controllers/cases_controller.rb
```diff
@@ -31,10 +32,9 @@ class CasesController < ApplicationController
                   :update_closure,
                   :response_upload_for_redraft,
```
mseedat-moj: Alphabetical order of method names is not maintained, please amend!

PR #618, opened 2018-12-27
Ct 2076 add draft timeliness to csv

beccapearce:
To be merged after CT-2022


app/services/csv_exporter.rb
```diff
@@ -47,6 +48,7 @@ def to_csv #rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength
           @kase.internal_deadline&.strftime('%F'),
           @kase.external_deadline&.strftime('%F'),
           @kase.date_responded.present? ? @kase.date_responded.strftime('%F') : nil,
+          @kase.date_draft_compliant.present? ? @kase.date_draft_compliant.strftime('%F') : nil,
```
starswan: might be out of scope, but wondering if we shouldn't be using a named to_s overload on dates rather than calling strftime('%F') everywhere?

starswan (commented):
Looks good apart form maybe DRYing up the strftime('%F') bits - but could be considered out of scope.

app/services/csv_exporter.rb
```diff
@@ -35,18 +35,19 @@ def initialize(kase)
     @kase = kase
   end
 
-  def to_csv #rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength
+  def to_csv #rubocop:disable Metrics/MethodLength
```
starswan: Given we are adding a field to a CSV, shouldn't there be a matching change in the CSV header?

starswan (changes requested):
Sorry just spotted something else.

starswan (approved):

PR #661, opened 2019-01-30
Ct 2109 extend sar deadline

mseedat-moj:
### Summary
Allow user to extend the deadline to respond to a SAR for up to 30 days at a time

### Ticket
Implements both add and remove SAR extension deadline due to ticket confusion.

https://dsdmoj.atlassian.net/browse/CT-2109
https://dsdmoj.atlassian.net/browse/CT-2116

### Notes

-  `extend_for_pit` implementation change can be found in PR #677 
- Assumes extension is done using Calendar days rather than Business days
- Feature test for Approver user extending a SAR deadline is outstanding (skipped test)
- Note the alteration in tests related to case history as the use of `simple_format` exposes
   line break characters stored in the database that were previously not retrieved

### Screenshots
screen shot 2019-02-14 at 01 37 47

screen shot 2019-02-14 at 01 37 58


app/services/case_extend_deadline_for_sar_service.rb
```diff
@@ -0,0 +1,70 @@
...
+    @case = kase
+    @extension_deadline = extension_deadline
+    @reason = reason
+    @result = :incomplete
+  end
+
+  def call
+    ActiveRecord::Base.transaction do
+      if validate_params
+        # TODO (Mohammed Seedat): State Machine rules for SAR Extensions
```
beccapearce: Think this should be the same as for pit extensions, so managers and approvers can extend in the state drafting for both standard and trigger cases and in the state awaiting dispatch for trigger cases

app/policies/case/sar_policy.rb
```diff
@@ -89,6 +89,11 @@ def can_request_further_clearance?
     !self.case.flagged?
   end
 
+  # TODO (Mohammed Seedat): Replace with rules for authorization
+  def extend_deadline_for_sar?
+    true
```
beccapearce: Personally I like to have this logic in the state machine so here we would have 
```
clear_failed_checks
    check_can_trigger_event(:extend_deadline_for_sar)
```
which will check the permissions set in the state machine

app/views/cases/extend_deadline_for_sar.html.slim
```diff
@@ -0,0 +1,51 @@
+- content_for :page_title do
```
beccapearce: Might be worth adding in a view test for this page

beccapearce (commented):
Good start! Once we add the event to the state machine, we can look at adding it to the transition decorator so It'll show up in the case history, and adding it to out event tests.

app/models/case/sar.rb
```diff
@@ -115,6 +116,47 @@ def sar?
...
+  def extendable?
+    max_allowed_deadline_date > external_deadline
+  end
+
+  # Original deadlines are not stored due to the way Polymorphic
+  # Cases are created. Therefore retrieving from the earliest
+  # case transition history is necessary otherwise assume current
+  # external_date is the final deadline.
+  def initial_deadline
+    sar_extensions = self.transitions
```
beccapearce: is it worth adding a scope for this on case transitions or is that overkill?
just thinking we could have some kind of extended transitions scope that could collect all the sar extensions and pit extensions, that way we could make it a little more consistent

spec/services/case_extend_sar_deadline_service_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+        it { is_expected.to eq :validation_error }
+        it {
+          expect(sar_case.errors[:reason_for_extending])
+            .to eq ["can't be blank"]
+        }
+      end
+
+      context 'exception' do
+        let(:service) { new_service(manager, sar_case, 3) }
```
beccapearce: Could we call out the number here as a separate variable?

`let(:days_to_extend_by) { 3 }
let(:service) { new_service(manager, sar_case, days_to_extend_by) }`

or see other comment about named variables

spec/services/case_extend_sar_deadline_service_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+        subject! { new_service(manager, sar_case, 1, reason: '').call }
+
+        it { is_expected.to eq :validation_error }
+        it {
+          expect(sar_case.errors[:reason_for_extending])
+            .to eq ["can't be blank"]
+        }
+      end
+
+      context 'exception' do
```
beccapearce: what is the exception here?

spec/services/case_extend_sar_deadline_service_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+        context 'is within limit' do
+          subject! { new_service(manager, sar_case, max_extension).call }
+
+          it { is_expected.to eq :ok }
+        end
+      end
+
+      context 'reason' do
+        subject! { new_service(manager, sar_case, 1, reason: '').call }
```
beccapearce: What is new service? could we rename this?

spec/services/case_extend_sar_deadline_service_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+          expect(sar_case.errors[:reason_for_extending])
+            .to eq ["can't be blank"]
+        }
+      end
+
+      context 'exception' do
+        let(:service) { new_service(manager, sar_case, 3) }
+
+        it 'rolls-back changes' do
+          allow(sar_case).to receive(:update!).and_throw(RuntimeError)
```
beccapearce: Not sure what's causing the error here. Is it something we could have more meaningful output of to help us track it down if it breaks in production?

spec/services/case_extend_sar_deadline_service_spec.rb
```diff
@@ -0,0 +1,172 @@
...
+
+  context '#new_extension_deadline' do
+    context 'calculates extended date with calendar days' do
+      it 'uses the original date before first extension' do
+        service = new_service(manager, sar_case, 3)
+
+        expect(service.send(:new_extension_deadline, 3))
+          .to eq(initial_deadline + 3.days)
+
+        service.call # required to create new deadline
```
beccapearce: What do you mean by this comment? Also not sure of the setup up here, why are we sending the methods to the service shouldn't we create the service with the variables?

spec/services/case_remove_sar_deadline_extension_service_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+          result = new_service(manager, sar_case).call
+
+          expect(result).to eq :ok
+          expect(sar_case.external_deadline).to eq initial_deadline
+          expect(sar_case.deadline_extended?).to be false
+        end
+      end
+    end
+
+    context 'exception' do
```
beccapearce: what exception is this?

spec/site_prism/page_objects/pages/cases/extend_sar_deadline_page.rb
```diff
@@ -0,0 +1,30 @@
...
+          PageObjects::Sections::PrimaryNavigationSection,
+          '.global-nav'
+
+        section :page_heading,
+          PageObjects::Sections::PageHeadingSection,
+          '.page-heading'
+
+        element :copy, '.action-copy'
+
+        # Form
```
beccapearce: Is this just a note? I think if it's necessary to put it in then it should be a section?

spec/state_machines/workflows/sar_permitted_events/sar_standard_spec.rb
```diff
@@ -43,6 +43,7 @@
                                                                       :assign_to_new_team,
                                                                       :destroy_case,
                                                                       :edit_case,
+                                                                      :extend_sar_deadline,
```
beccapearce: If this is in the standard spec shouldn't it also be in the trigger spec?

spec/views/cases/show_html_slim_spec.rb
```diff
@@ -310,4 +313,62 @@ def setup_policies(policies)
...
     end
   end
+
+  describe 'SAR case' do
+    context 'before it is extended' do
+      let(:sar) { create(:approved_sar).decorate }
+
+      before do
+        assign(:case, sar)
+        assign(:permitted_events, [:extend_sar_deadline])
```
beccapearce: does this need to be in both the permitted events and filtered permitted events?

spec/views/cases/show_html_slim_spec.rb
```diff
@@ -310,4 +313,62 @@ def setup_policies(policies)
       end
     end
   end
+
+  describe 'SAR case' do
```
beccapearce: what is this test actually testing for?

beccapearce (changes requested):
Obviously some tests still to fix. 
Just an overview of what I spotted since it seems there's more work to do on it to get the tests passing

spec/views/cases/show_html_slim_spec.rb
mseedat-moj: Yes in `cases/show.html.slim` the action bar only shows if `@filtered_permitted_events` has values and then only outputs buttons if `@permitted_events` is also populated.

spec/views/cases/show_html_slim_spec.rb
mseedat-moj: I read it in conjunction with the context so that the first context would read `SAR case before it is extended`. Is there a style guide for writing specs/preference?

app/models/case/sar.rb
```diff
@@ -115,6 +116,47 @@ def sar?
...
+
+    if sar_extensions.any?
+      sar_extensions.first.original_final_deadline
+    else
+      external_deadline
+    end
+  end
+
+  def extend_deadline!(new_deadline)
+    self.reload
```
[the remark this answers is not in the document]
mseedat-moj: I'm basing this on the existing code in `CaseExtendForPITService` where the reload is carried out before the update. I assumed this was done out of a bad experience with inconsistent objects?

app/models/user.rb
```diff
@@ -61,6 +61,12 @@ class User < ApplicationRecord
   }
   scope :active_users, -> { where(deleted_at: nil) }
 
+  ROLE_WEIGHTINGS = {
```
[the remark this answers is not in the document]
mseedat-moj: Let's discuss how this could/should work

app/services/case_extend_sar_deadline_service.rb
```diff
@@ -0,0 +1,88 @@
+class CaseExtendSARDeadlineService
+  attr_reader :result, :error
+
+  def initialize(user, kase, extension_days, reason)
```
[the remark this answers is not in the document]
mseedat-moj: I've copied the code from `CaseExtendForPITService` - is named parameters something we want to  make the default for method signatures?

app/services/case_extend_sar_deadline_service.rb
```diff
@@ -0,0 +1,88 @@
...
+    Rails.logger.error err.to_s
+    Rails.logger.error err.backtrace.join("\n\t")
+    @error = err
+    @result = :error
+  end
+
+  private
+
+  def new_extension_deadline(extend_by)
+    current_deadline =  @case.deadline_extended? ?
```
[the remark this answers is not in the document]
mseedat-moj: It was a single line (100 chars) but I've tried to be more strict with the 80 character per line limit! Rejigged.

spec/services/case_extend_sar_deadline_service_spec.rb
mseedat-moj: This was taken from tests for `CaseExtendForPITService` which tries to active the failure of the ActiveRecord transaction.

spec/services/case_extend_sar_deadline_service_spec.rb
mseedat-moj: I'm using `service.send(:new_extension_deadline, 3)` because `new_extension_deadline` is a private method.

app/controllers/cases_controller.rb
```diff
@@ -642,6 +645,54 @@ def remove_pit_extension
...
+  end
+
+  def remove_sar_deadline_extension
+    authorize @case, :remove_sar_deadline_extension?
+
+    service = CaseRemoveSARDeadlineExtensionService.new(current_user, @case)
+    result = service.call
+
+    if result == :ok
+      flash[:notice] = 'Deadline extension removed'
```
mseedat-moj: Replace with Translation

app/controllers/cases_controller.rb
```diff
@@ -642,6 +645,54 @@ def remove_pit_extension
...
+      redirect_to case_path(@case.id)
+    else
+      flash[:alert] = 'Unable to remove SAR deadline extension'
```
mseedat-moj: Replace with translation

app/models/case/sar.rb
```diff
@@ -59,7 +59,8 @@ def self.searchable_fields_and_ranks
                  third_party: :boolean,
                  third_party_relationship: :string,
                  reply_method: :string,
-                 late_team_id: :integer
+                 late_team_id: :integer,
+                 deadline_extended: [:boolean, default: false]
```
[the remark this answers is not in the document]
beccapearce: TO DISCUSS

app/models/case/sar.rb
beccapearce: TO DISCUSS

spec/views/cases/show_html_slim_spec.rb
beccapearce: No style guide, just should be understandable, we check for that in peer review. I was just saying I fing it a little unclear I feel the describe should be 'extend SAR deadline link' or something, that seems like the thing we are testing here

spec/services/case_extend_sar_deadline_service_spec.rb
beccapearce: ok, could we improve it?

app/services/case_extend_sar_deadline_service.rb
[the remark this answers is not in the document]
beccapearce: Yes, I want it to be multiline I haven't asked for it to be single line, 
`if @case.deadline_extended?`
etc.

app/services/case_extend_sar_deadline_service.rb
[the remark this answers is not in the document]
beccapearce: I think in all the tests there are loads of magic numbers, we could either put them in as named variables here so they make sense elsewhere or call them out as a different variable in the test.
I;m trying to avoid something like service.new(3) what is 3? I'd rather service.new(days_to_extend_by: 3) or days_to extend by = 3 service.new(days to extend by)

app/models/case/sar.rb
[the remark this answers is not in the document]
beccapearce: I don't know you can have a look through the old commits, they might have more info as to the reasoning. Can we see if it's needed, or if it works the other way round?

spec/services/case_extend_sar_deadline_service_spec.rb
beccapearce: but isn't it called from the initialise? shouldn't we create the service with the variables?

app/decorators/case_transition_decorator.rb
```diff
@@ -36,6 +36,7 @@ def details
          'add_response_to_flagged_case',
          'approve_and_bypass',
          'extend_for_pit',
+         'extend_sar_deadline',
```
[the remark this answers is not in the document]
mseedat-moj: There isn't one for `remove_pit_extension` so not sure what the rule is.

app/models/case/sar.rb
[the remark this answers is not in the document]
mseedat-moj: Recommend a separate piece of work if this is to be carried forward

app/models/case/sar.rb
```diff
@@ -115,6 +116,47 @@ def sar?
     true
   end
 
+  def extendable?
+    max_allowed_deadline_date > external_deadline
+  end
+
+  # Original deadlines are not stored due to the way Polymorphic
```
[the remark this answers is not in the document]
mseedat-moj: Removed.

app/models/case/sar.rb
mseedat-moj: Perhaps one to Roll in with any changes to `extend_for_pit` changes.

app/models/case/sar.rb
```diff
@@ -115,6 +116,47 @@ def sar?
...
+                            .order(:id)
+
+    if sar_extensions.any?
+      sar_extensions.first.original_final_deadline
+    else
+      external_deadline
+    end
+  end
+
+  def extend_deadline!(new_deadline)
```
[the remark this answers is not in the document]
mseedat-moj: Decided to leave these in the `Case::SAR` model as I believe the mechanism for actual extension/removal should be encapsulated within the `Case::SAR` model to maintain consistency.

app/models/case/sar.rb
```diff
@@ -115,6 +116,47 @@ def sar?
...
+    self.reload
+    self.update!(
+      external_deadline: new_deadline,
+      deadline_extended: true
+    )
+  end
+
+  def reset_deadline!
```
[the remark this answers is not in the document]
mseedat-moj: As comment for `extend_deadline!`

app/models/user.rb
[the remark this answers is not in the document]
mseedat-moj: Well spotted - didn't even use the weightings for the `extend_sar_deadline` work! Will amend!

app/state_machines/workflows/predicates.rb
```diff
@@ -117,6 +117,18 @@ def has_pit_extension?
     @kase.extended_for_pit?
   end
 
+  def has_sar_deadline_extension?
+    @kase.try(:deadline_extended?)
+  end
+
+  def can_extend_sar_deadline?
```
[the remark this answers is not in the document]
mseedat-moj: Will need some help with this

app/state_machines/workflows/predicates.rb
```diff
@@ -117,6 +117,18 @@ def has_pit_extension?
...
+    @kase.try(:extendable?)
+  end
+
+  def approver_can_remove_sar_deadline?
```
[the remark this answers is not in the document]
mseedat-moj: As above, need some help with best naming scheme/role name to use

config/state_machine/configs/00_moj/20_sar/20_sar_standard_workflow.yml
```diff
@@ -122,5 +126,9 @@
                   switch_workflow: trigger
 
               drafting:
+                extend_sar_deadline:
+                  if: Workflows::Predicates#user_is_assigned_disclosure_specialist?
                 flag_for_clearance:
                   switch_workflow: trigger
+                remove_sar_deadline_extension:
+                  if: Workflows::Predicates#approver_can_remove_sar_deadline?
```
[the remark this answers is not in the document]
mseedat-moj: As with predicates.rb comments require some assistance re: naming

config/state_machine/configs/00_moj/20_sar/20_sar_standard_workflow.yml
```diff
@@ -122,5 +126,9 @@
                   switch_workflow: trigger
 
               drafting:
+                extend_sar_deadline:
+                  if: Workflows::Predicates#user_is_assigned_disclosure_specialist?
```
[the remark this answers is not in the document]
mseedat-moj: As with predicates.rb comments require some assistance re: naming/role

app/state_machines/workflows/predicates.rb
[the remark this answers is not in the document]
beccapearce: so extendable calls `max_allowed_deadline_date > external_deadline` 
so I guess deadline_not_extended_over_max_allowed_deadline?
or deadline_does_not_exceed_max_deadline?,
Have a think a think about how to communicate this criteria in the method name

app/state_machines/workflows/predicates.rb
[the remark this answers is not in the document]
beccapearce: I suggested case_extended_and_user_assigned_disclosure_specialist or something?
Like above lets communicate what the method does in the method name

config/state_machine/configs/00_moj/20_sar/20_sar_standard_workflow.yml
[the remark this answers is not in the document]
beccapearce: This isn't a naming issue

app/models/case/sar.rb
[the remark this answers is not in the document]
beccapearce: We discussed this when we started the work. I said we could add a field as long as we did the same for FOIs. Still would like the two to be consistent.

app/models/case/sar.rb
beccapearce: Don't think we actually discussed this. It was just something to think about.

app/models/case/sar.rb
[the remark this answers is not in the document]
beccapearce: Thinking we should move the calling of the state machine into this method so we aren't calling two extend_sar methods in the service

app/decorators/case_transition_decorator.rb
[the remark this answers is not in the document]
mseedat-moj: No custom reason for removing a SAR deadline extension so state not required in this case statement

app/models/user.rb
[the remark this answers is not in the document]
mseedat-moj: I've left the additional changes to `User.rb` in place and additionally parameterised `role_weightings` in User#self.sort_teams_by_roles`.

This is to prevent code duplication for `User#team_for_case` used in  `CaseExtendSARDeadlineService`, `CaseRemoveSARDeadlineExtensionService` & `MessagesController` at which point it makes sense to move the weighting/role sorting code in to the `User` model as well for consistency.

config/state_machine/configs/00_moj/20_sar/20_sar_standard_workflow.yml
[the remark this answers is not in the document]
mseedat-moj: Renamed to `case_extended_and_user_assigned_disclosure_specialist?` as per recommendation for `predicates.rb` review

spec/services/case_extend_sar_deadline_service_spec.rb
mseedat-moj: Renamed with more descriptive context name

spec/services/case_extend_sar_deadline_service_spec.rb
mseedat-moj: Renamed with more descriptive spec name

spec/services/case_remove_sar_deadline_extension_service_spec.rb
mseedat-moj: Renamed with more descriptive context name

spec/views/cases/show_html_slim_spec.rb
mseedat-moj: Renamed spec name to be more descriptive

spec/models/case/sar_spec.rb
```diff
@@ -254,4 +254,98 @@
...
+
+    describe '#extendable?' do
+      it 'is true if external_deadline is less than max possible deadline' do
+        max_statutory_deadline = subject.initial_deadline + max_extension.days
+
+        expect(subject.extendable?).to eq true
+        expect(subject.external_deadline).to be < max_statutory_deadline
+      end
+
+      it 'is false when already extended equal or beyond satutory limit' do
```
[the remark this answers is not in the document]
mseedat-moj: It's trying to test the edge case of the new deadline being set to the maximum allowable statutory limit and therefore ensuring the result of `deadline_extendable?` is `false`

spec/factories/case/sars.rb
```diff
@@ -211,6 +211,16 @@
     end
   end
 
+  factory :extended_deadline_sar, parent: :accepted_sar do
```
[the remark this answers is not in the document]
mseedat-moj: The factories in `case/sars.rb` all seem to have separate factories for different states so continued that pattern.

app/models/case/sar.rb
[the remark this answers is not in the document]
mseedat-moj: Implemented in separate PR (for merge into this one) #677

app/controllers/cases_controller.rb
```diff
@@ -14,13 +14,16 @@ class CasesController < ApplicationController
                   :edit,
                   :edit_closure,
                   :extend_for_pit,
+                  :extend_sar_deadline,
                   :execute_extend_for_pit,
+                  :execute_extend_sar_deadline,
```
starswan: think this should be extend_sar_deadline_action for consistency - there are more xxx_action methods than execute_xxx methods

app/services/case_extend_sar_deadline_service.rb
```diff
@@ -0,0 +1,89 @@
...
+    else
+      @case.initial_deadline + extend_by.days
+    end
+  end
+
+  def message
+    "#{@reason}\nDeadline extended by #{@extension_days} days"
+  end
+
+  def validate_params
```
starswan: It's not completely obvious to me that this method returns a falsey value when there is an error, which is required by its caller.

app/services/case_remove_sar_deadline_extension_service.rb
```diff
@@ -0,0 +1,28 @@
+class CaseRemoveSARDeadlineExtensionService
+  attr_reader :result, :error
+
+  def initialize(user, kase)
+    @user = user
+    @case = kase
+    @result = :incomplete
+  end
+
+  def call
```
starswan: Not really sure about the boiler-plate error handling. Most services catch specific errors, this one (and the one above) catch everything (even syntax errors)

app/services/case_extend_sar_deadline_service.rb
```diff
@@ -0,0 +1,89 @@
...
+    end
+
+    unless extension_date_valid?
+      @result = :validation_error
+    end
+
+    @result = :ok if @case.errors.empty?
+  end
+
+  def extension_date_valid?
```
starswan: Both the case and this service know about Settings.sar_extension_limit and what it means (the error message here expresses the same intent as coded in Case#max_allowed_deadline_date) Maybe this is a justification for adding a validation to the case model?

app/services/case_extend_sar_deadline_service.rb
```diff
@@ -0,0 +1,89 @@
...
+    extension_limit = @case.max_allowed_deadline_date
+
+    valid = false
+
+    if @extension_days.blank? || @extension_deadline.blank?
```
starswan: given that it is calculated, I don't think extension_deadline can ever be blank?

app/services/case_remove_sar_deadline_extension_service.rb
starswan: Think the transaction handling could do with documenting and/or splitting out (probably not in this story though). I think there is a general desire to not show the standard crash page when things go wrong, but to rollback the transaction and be polite.

app/services/case_remove_sar_deadline_extension_service.rb
```diff
@@ -0,0 +1,28 @@
...
+      @case.reset_deadline!
+      @result = :ok
+    end
+    @result
+  rescue => err
+    Rails.logger.error err.to_s
+    Rails.logger.error err.backtrace.join("\n\t")
+    @error = err
+    @result = :error
+    raise err
```
starswan: Given my comment above, I think this might be a bug...?

config/locales/en.yml
```diff
@@ -157,9 +157,6 @@ en:
               <<: *report_dates
               before_start_date: "can't be after Date to"
 
-  page_title:
```
beccapearce: Does this need to be removed?

app/state_machines/workflows/predicates.rb
```diff
@@ -114,7 +114,19 @@ def overturned_editing_enabled_and_responder_in_assigned_team?
...
-    @kase.extended_for_pit?
+    @kase.has_pit_extension?
+  end
+
+  def has_sar_deadline_extension?
+    @kase.try(:deadline_extended?)
+  end
+
+  def deadline_does_not_exceed_max_deadline?
+    @kase.try(:deadline_extendable?)
```
starswan: Not sure about this try? In what circumstances might the @kase variable be nil?

db/migrate/20190220183951_add_pit_deadline_extended_property_to_cases.rb
```diff
@@ -0,0 +1,17 @@
...
+        .order(created_at: :desc)
+
+      if pit_transitions.any? && pit_transitions.last.event == 'extend_for_pit'
+        kase.has_pit_extension = true
+        kase.save!
+      end
+    end
+  end
+
+  def down
```
beccapearce: We should raise something here if there is no down method so we know that this is an irreversible migration

spec/factories/case/sars.rb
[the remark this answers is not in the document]
beccapearce: The other factories use traits and since the extension can happen in different states it makes sense to have it as a trait.
There is a pattern for this in the foi factories

spec/models/case/sar_spec.rb
```diff
@@ -254,4 +254,98 @@
     end
   end
 
+  describe 'deadline' do
+    subject             { create :sar_case }
```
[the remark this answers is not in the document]
beccapearce: Could we rename subject to reflect what it is.

spec/models/case/sar_spec.rb
```diff
@@ -254,4 +254,98 @@
...
+    describe '#max_allowed_deadline_date' do
+      it 'is 60 calendar days after the initial_deadline' do
+        max_statutory_deadline = subject.initial_deadline + 60.days
+
+        expect(subject.max_allowed_deadline_date).to eq max_statutory_deadline
+      end
+    end
+
+    describe '#extend_deadline!' do
+      subject!                { create :approved_sar }
```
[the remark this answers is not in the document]
beccapearce: can we rename subject to reflect what it is

spec/models/case/sar_spec.rb
```diff
@@ -254,4 +254,98 @@
...
+      it 'sets new external_deadline' do
+        expect { subject.extend_deadline!(new_deadline) }.to \
+          change(subject, :external_deadline)
+          .from(initial_deadline)
+          .to(new_deadline)
+      end
+    end
+
+    describe '#reset_deadline!' do
+      subject!                  { create :extended_deadline_sar }
```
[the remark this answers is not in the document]
beccapearce: can we rename subject to reflect what it is

spec/decorators/case_extend_sar_deadline_decorator_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  context '#allow_extension_period_selection?' do
+    let(:new_sar_case)       { create(:sar_being_drafted) }
+    let(:extended_sar_case)  { create(:extended_deadline_sar) }
+
+    it 'should be true for a new SAR' do
+      decorated_case = CaseExtendSARDeadlineDecorator.decorate new_sar_case
+      expect(decorated_case.allow_extension_period_selection?).to be true
+    end
+
+    it 'should be false for a SAR that is already extended' do
```
starswan: Unsure about this test. If the SAR has been extended by 30 days, shouldn't it be possible to extend it by another 30 days?

spec/services/case_extend_sar_deadline_service_spec.rb
```diff
@@ -0,0 +1,185 @@
+require "rails_helper"
+
+describe CaseExtendSARDeadlineService do
+  let(:team_disclosure_bmt) { find_or_create :team_disclosure_bmt }
+  let(:manager)             { find_or_create :disclosure_bmt_user }
+  let!(:sar_case)           { create(:approved_sar) }
```
beccapearce: is the ! needed?

spec/services/case_extend_sar_deadline_service_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+  let!(:initial_deadline)   { sar_case.initial_deadline }
+  let!(:max_extension)      { Settings.sar_extension_limit.to_i }
```
beccapearce: Is the ! needed?

app/services/case_extend_sar_deadline_service.rb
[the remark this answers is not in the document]
beccapearce: Named variables?

spec/models/case/sar_spec.rb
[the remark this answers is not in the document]
beccapearce: This would be better with a timecop.

spec/models/case/sar_spec.rb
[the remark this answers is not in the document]
beccapearce: can we rename subject to reflect what it is

spec/models/case/sar_spec.rb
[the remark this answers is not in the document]
beccapearce: can we rename subject to reflect what it is

spec/decorators/case_extend_sar_deadline_decorator_spec.rb
mseedat-moj: Discussed with @starswan - test makes sense when read in conjunction with the context.

spec/support/features/interactions.rb
```diff
@@ -249,5 +249,31 @@ def upload_response_with_action_param(kase, user, action)
       allow(uploader).to receive(:remove_leftover_upload_files)
       rus.upload!
     end
+
+    def extend_sar_deadline_for(kase, num_days, reason: 'The reason for extending')
```
beccapearce: In keeping the the rest of this file worth having named variables

beccapearce:
A fair few comments have been marked as resolved but not addressed, Might be worth double checking Mo

app/state_machines/workflows/predicates.rb
mseedat-moj: Good question @starswan - use of `try` as a defensive solution as `deadline_extendable?` is only available for SAR cases. Note added as per discussion.

config/locales/en.yml
mseedat-moj: Yes all the other page titles are in `page_titles.yml` so I think the left over `marking_response_as_sent` was a mistake.

app/services/case_remove_sar_deadline_extension_service.rb
mseedat-moj: Discussed with @starswan - leave as is for now as the handling of errors is a system wide issue requiring a broader discussion

app/services/case_remove_sar_deadline_extension_service.rb
mseedat-moj: Discussed with @starswan - leave as is for now as the handling of errors is a system wide issue requiring a broader discussion

app/services/case_extend_sar_deadline_service.rb
mseedat-moj: Discussed with @starswan - leave as is for now as the handling of validations is a system wide issue requiring a broader discussion

app/services/case_extend_sar_deadline_service.rb
mseedat-moj: Very good point. Have re-written this code however have not back ported the same to extend_for_pit_service as that is well beyond the scope of this ticket.

app/services/case_extend_sar_deadline_service.rb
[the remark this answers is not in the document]
mseedat-moj: I've now changed the service to request named variables rather than only ensuring tests use descriptive variable names/named parameters.

db/migrate/20190220183951_add_pit_deadline_extended_property_to_cases.rb
mseedat-moj: Decided against raising an exception because this migration creates the denormalisation as a JSON property, so if the database were to be rolled back the case_transitions would still exist, so there's nothing to roll back in theory.

spec/factories/case/sars.rb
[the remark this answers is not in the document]
mseedat-moj: I'm leaving this as is as I followed the current implementation in factories/cases/sars.rb which has a factory per state.

spec/models/case/sar_spec.rb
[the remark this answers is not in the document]
mseedat-moj: I've gone through the tests again after discussion with @starswan and decided to use `let` as it makes more sense in most of the cases where I've used subject.

spec/models/case/sar_spec.rb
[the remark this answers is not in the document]
mseedat-moj: Re-arranged the test execution steps which makes the test intention easier to understand

spec/models/case/sar_spec.rb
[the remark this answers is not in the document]
mseedat-moj: I've gone through the tests again after discussion with @starswan and decided to use `let` as it makes more sense in most of the cases where I've used subject.

spec/models/case/sar_spec.rb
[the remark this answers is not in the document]
mseedat-moj: I've gone through the tests again after discussion with @starswan and decided to use `let` as it makes more sense in most of the cases where I've used subject.

spec/services/case_extend_sar_deadline_service_spec.rb
mseedat-moj: Went through the tests and removed the `!` where possible.

spec/services/case_extend_sar_deadline_service_spec.rb
mseedat-moj: Went through the tests and removed the `!` where possible.

app/models/case/sar.rb
mseedat-moj: I would wait until the model has more methods requiring some kind of case_transition retrieval and extract at that point.

app/models/case/sar.rb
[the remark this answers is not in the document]
mseedat-moj: Out of scope for this ticket. Discussed with @beccapearce as a general issue regarding system software design.

spec/factories/case/sars.rb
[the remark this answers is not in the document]
beccapearce: Exactly we have a factory per state. Extended sar is not a state so we should have a trait for it rather than a factory such as in factories/cases/fois.rb.

db/migrate/20190220183951_add_pit_deadline_extended_property_to_cases.rb
beccapearce: Should probably be a data migration then.

beccapearce (approved):

PR #665, opened 2019-02-05
ct-2061 set scope on refusal reasons to prevent sar reasons showing in foi closure

beccapearce:
…n foi closure


mseedat-moj (approved):
Sensible changes. Approved

PR #680, opened 2019-02-25
CT-2117 Display late responder name in stats

starswan:
display business unit responsible for late response in CSV downloads

### DIFF
--- a/app/helpers/cases_helper.rb
+++ b/app/helpers/cases_helper.rb
@@ -263,18 +263,6 @@ module CasesHelper
     end
   end
 
-  def case_late_team_name(kase)
-    if kase.responded_late?
-      if kase.late_team.nil?
-        'Unspecified'
-      else
-        kase.late_team.name
-      end
-    else
-      'N/A'
-    end
-  end
-
   def manager_updating_close_details_on_old_case?(user, kase)
     # the only reason a manager can't the closure details of a closed case is because
     # it is an "old style" closure, i.e. it was closed before we implemented the new
--- a/app/helpers/cases_helper.rb
+++ b/app/helpers/cases_helper.rb
@@ -263,6 +263,18 @@ module CasesHelper
     end
   end
 
+  def case_late_team_name(kase)
+    if kase.responded_late?
+      if kase.late_team.nil?
+        'Unspecified'
+      else
+        kase.late_team.name
+      end
+    else
+      'N/A'
+    end
+  end
+
   def manager_updating_close_details_on_old_case?(user, kase)
     # the only reason a manager can't the closure details of a closed case is because
     # it is an "old style" closure, i.e. it was closed before we implemented the new
--- a/app/services/csv_exporter.rb
+++ b/app/services/csv_exporter.rb
@@ -66,7 +66,7 @@ class CSVExporter
           @kase.respond_to?(:reply_method) ? @kase.reply_method : nil,
           @kase.respond_to?(:subject_type) ? @kase.subject_type : nil,
           @kase.respond_to?(:subject_full_name) ? @kase.subject_full_name : nil,
-          case_late_team_name(@kase)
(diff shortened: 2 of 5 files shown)


app/services/csv_exporter.rb
```diff
@@ -62,11 +64,12 @@ def to_csv
...
+          @kase.respond_to?(:subject_full_name) ? @kase.subject_full_name : nil,
+          case_late_team_name(@kase)
       ]
-    rescue => err
-      raise CSVExporterError.new("Error encountered formatting case id #{@kase.id} as CSV:\nOriginal error: #{err.class} #{err.message}")
-    end
+    # rescue => err
+    #   raise CSVExporterError.new("Error encountered formatting case id #{@kase.id} as CSV:\nOriginal error: #{err.class} #{err.message}")
+    # end
   end
```
mseedat-moj: Exception commented out - should this be deleted or uncommented?

spec/services/csv_exporter_spec.rb
```diff
@@ -4,33 +4,40 @@
 
   # let(:responding_team)   { create :responding_team, name: 'Transport for London' }
   # let(:responder)         { create :responder, full_name: 'Joe Smith', responding_teams: [ responding_team] }
+  let(:late_team)   { create :responding_team, name: 'Transport for London' }
   let(:foi_case)          { create :closed_case, :fully_refused_exempt_s40,
```
mseedat-moj: Look nicer if the `let` were all aligned! And the above existing commented out `let` - worth deleting them?

spec/services/csv_exporter_spec.rb
```diff
@@ -4,33 +4,40 @@
...
                              'FOI Responding Team',
                              'foi responding user',
-                             '2018-08-30',
-                             '2018-09-13',
-                             '2018-09-27',
-                             '2018-09-25',
+                             '2018-02-01',
+                             '2018-02-15',
+                             '2018-03-01',
+                             '2018-05-01',
```
mseedat-moj: What's the reason for changing these dates? Easier to comprehend?

mseedat-moj (changes requested):
Some questions, generally looks good,

app/helpers/cases_helper.rb
```diff
@@ -263,6 +263,18 @@ def case_details_for_link_type(link_type)
...
+      if kase.late_team.nil?
+        'Unspecified'
+      else
+        kase.late_team.name
+      end
+    else
+      'N/A'
+    end
+  end
+
```
mseedat-moj: Agree with decision to move the `late_team_name` code - is there a decorator this function could belong to rather than a helper as that seems to be the way.

starswan: This was in a decorator. It seems that the decorator 'pattern' breaks down when there are multiple (different) display formats for the same model.

spec/services/csv_exporter_spec.rb
starswan: The original test asserted things from the factory. Now the test has to set them explicitly, the expectations change accordingly.

app/services/csv_exporter.rb
```diff
@@ -28,14 +29,16 @@ class CSVExporter
...
+      'SAR Subject full name',
+      'Business unit responsible for late response'
   ]
 
   def initialize(kase)
     @kase = kase
   end
 
   #rubocop:disable Metrics/CyclomaticComplexity
+  #rubocop:disable Metrics/MethodLength
```
beccapearce: could we put these rubocop:disable methods on the to_csv method
`def to_csv #rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength`
That way we don't have to disable and enable them

mseedat-moj (approved):

mseedat-moj (approved):

PR #682, opened 2019-02-27
CT-2138 fix places where Timecop.freeze isn't rolled back during tests

starswan:
Running all the tests fails sporadically due to some tests failing to use correct Timecop.freeze patterns - either 
a) using a block or 
b) before { Timecop.freeze(t) } 
     after { Timecop.return }

This PR fixes (I think) all tests that exhibit this behaviour


beccapearce:
YASSSSS! So glad someone's found this I've been trying to spot it for ages!

beccapearce (approved):

PR #683, opened 2019-02-28
Ct 2131 code coverage optional

starswan:
Make code coverage optional (via a COVERAGE variable) as it speeds up local builds. 
Added 'rails' profile to split out models, controllers etc and remove specs from coverage.
Added extra sections to report for validators, services, policies and decorators.


beccapearce (approved):
Looks good!
Could we add something to the readme about how to run with coverage

PR #684, opened 2019-02-28
Ct 2075 edit draft timeliness

beccapearce:
awaiting CT-2022 to be merged


spec/factories/case/ico_fois.rb
```diff
@@ -131,32 +131,33 @@
...
              acting_team: evaluator.approving_team,
              acting_user: evaluator.approver
 
       kase.approver_assignments.each { |a| a.update approved: true }
+      kase.update!(date_draft_compliant: evaluator.date_draft_compliant)
       kase.reload
     end
   end
 
   factory :responded_ico_foi_case, parent: :approved_ico_foi_case do
```
starswan: Unsure of the reason for the transient + after_create loop - maybe a comment would help?

starswan (changes requested):
Looks good other than maybe a comment needed in the factories.

starswan (approved):

PR #686, opened 2019-03-01
CT-2113 humanise outputs for csv

beccapearce:
Remove draft deadline for non-trigger cases - A draft deadline is showing for non-trigger cases, this should be blank.

Improve 'Request status' column - Values in this column should match the text in the service e.g. 'Needs reassigning', 'To be accepted' etc.

Improve 'Workflow' column - This column should be called 'Trigger' and should have the value 'Yes' if true

References to DACU should be removed - Instances of DACU should be replaced with "Disclosure Team"

Improve 'Third party' column - Should have the value 'Yes' if true

Improve 'Requester type' column - Replace '_' with spaces, e.g. 'member_of_the_public' → member of the public

Make column headings consistent - 'Info_held' → 'Info held' and 'External_deadline' → 'External deadline'

humanise reply method

### DIFF
--- a/app/services/csv_exporter.rb
+++ b/app/services/csv_exporter.rb
@@ -11,14 +11,14 @@ class CSVExporter
       'Responder',
       'Date received',
       'Internal deadline',
-      'External_deadline',
+      'External deadline',
       'Date responded',
       'Date compliant draft uploaded',
-      'Workflow',
+      'Trigger',
       'Name',
       'Requester type',
       'Message',
-      'Info_held',
+      'Info held',
       'Outcome',
       'Refusal reason',
       'Exemptions',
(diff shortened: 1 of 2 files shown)


app/services/csv_exporter.rb
```diff
@@ -74,4 +74,8 @@ def to_csv
   def dequote_and_truncate(text)
     text.tr('"', '').tr("'", '')[0..4000]
   end
+
+  def humanize_boolean(boolean)
+    boolean ? 'Yes' : nil
```
starswan: should this not be 'Yes'/'No' ?

app/services/csv_exporter.rb
```diff
@@ -39,30 +39,30 @@ def initialize(kase)
...
-          @kase.respond_to?(:reply_method) ? @kase.reply_method : nil,
-          @kase.respond_to?(:subject_type) ? @kase.subject_type : nil,
-          @kase.respond_to?(:subject_full_name) ? @kase.subject_full_name : nil
+        @kase.number,
+        @kase.decorate.pretty_type,
+        I18n.t("state.#{@kase.current_state}").downcase,
+        @kase.responding_team&.name,
+        @kase.responder&.full_name,
+        @kase.received_date&.strftime('%F'),
+        @kase.flagged? ? @kase.internal_deadline&.strftime('%F') : nil,
```
starswan: should this not be '' (blank) rather than nil in a CSV?

spec/services/csv_exporter_spec.rb
```diff
@@ -28,12 +26,12 @@
                              'FOI Responding Team',
                              'foi responding user',
                              '2018-08-30',
-                             '2018-09-13',
+                             nil,
```
starswan: I think we should be changing the test data so that these columns retain the same values. Or adding more tests so that all the cases are covered?

starswan (changes requested):
'' vs nils in CSV output...? don't know which is correct

app/services/csv_exporter.rb
beccapearce: Nil is what is currently being used such as
`@kase.respond_to?(:third_party) ? @kase.third_party : nil`
is there a need to change it to " "?

app/services/csv_exporter.rb
beccapearce: Looking at what we have currently we tend to leave the cell blank if it's no. The ticket too just said yes if true. Together I guessed that nil was the best choice

starswan (approved):

mseedat-moj:
Good improvements @beccapearce @starswan especially the use the hash rather than array to perform expectation tests

PR #687, opened 2019-03-01
Add active user count to user admin page

beccapearce:


spec/views/admin/users/index_html_slim_spec.rb
```diff
@@ -5,6 +5,7 @@
     responder = find_or_create :foi_responder
     approver = create :approver
     assign(:users, User.where(id: [responder.id, approver.id]).page(1).decorate)
+    assign(:active_users, User.where(id: [responder.id, approver.id]).page(1).decorate)
```
starswan: was hoping for an assert that the user count is displayed correctly? Not sure about line 8 - in the controller, the user list isn't decorated

app/controllers/admin/users_controller.rb
```diff
@@ -7,6 +7,8 @@ def index
                    .order(:full_name)
                    .page(params[:page]).per(100)
                    .decorate
+
+      @active_users = User.active_users
```
starswan: might be better to grab the count here as that's all we are passing to the view?

starswan (changes requested):
unsure about User.active_users vs User.active_users.count?

starswan (approved):

PR #691, opened 2019-03-04
CT-2129 Update readme

starswan:
Added notes around local postgresql installation and running with optional coverage stats


README.md
```diff
@@ -138,6 +147,61 @@ $ xcode-select --install
 ```
 </details>
 
+<details>
+<summary>Installing redis</summary>
+
+Redis is needed to run the 'db:reseed' task below (although after that it isn't needed)
```
beccapearce: I thought it was used, for uploading when running locally.

README.md
```diff
@@ -190,7 +254,7 @@ To create 200 cases in various states with various responders for search testing
 ```
 rake seed:search:data
 ```
-
+It appears that redis needs to be running to attempt this task - but it doesn't currently work for unknown reasons.
```
beccapearce: Think we looked at this, it works when redis is running and s3 is set up, but fails on late closed cases due to the late team id

README.md
```diff
@@ -101,6 +101,15 @@ There are two options for installing postgres:
 	$ brew install postgresql@9.5
 	```
 
+	Having done this, make sure all the post-install variables have been put in
+  `.bash_profile` or similar e.g.
+
+  export PKG_CONFIG_PATH="/usr/local/opt/postgresql@9.5/lib/pkgconfig"
+  export CPPFLAGS="-I/usr/local/opt/postgresql@9.5/include"
+  export LDFLAGS="-L/usr/local/opt/postgresql@9.5/lib"
+  export PATH=$PATH:/usr/local/opt/postgresql@9.5/bin
```
beccapearce: This block would look nicer with some code block formatting

beccapearce:
I think we need to have a look at the readme as a whole. 
We should prioritise what the code base does and the things we need to get started with the code base. 
I'm also concerned about the parts of the readme we are hiding with the collapsable sections vs what we are displaying. For example we hide how to install postgresql but expose the possible errors.

I'd say to make this better we should expose the parts of the read me associated with crucial details to do with setup etc. and move the details of what the code base does up. Other details such as the yardoc, paper trail, possible errors and jsonb should be hidden.

README.md
```diff
@@ -89,18 +89,27 @@ environments. Also, because the `structure.sql` file generated by PostgreSQL
...
-	$ brew install postgresql@9.5
-	```
+```
+$ brew install postgresql@9.5
+```
+
+Having done this, make sure all the post-install variables have been put in
+`.bash_profile` or similar e.g.
+
+export PKG_CONFIG_PATH="/usr/local/opt/postgresql@9.5/lib/pkgconfig"
```
beccapearce: Would still be good to format these variables

README.md
```diff
@@ -138,6 +147,97 @@ $ xcode-select --install
 ```
 </details>
 
+#### Issues installing PostgreSQL (pg) gem
```
beccapearce: These errors are more prominent than the actual install instructions.

README.md
```diff
@@ -190,7 +290,11 @@ To create 200 cases in various states with various responders for search testing
 ```
 rake seed:search:data
 ```
-
+In order for this task to run sucessfully, redis needs to be running
+```
+redis-server
+```
+and AWS access needs to have been setup, as creating cases (even on development) uploads file to S3
```
beccapearce: Do we need to provide details on how to get your AWS account?

beccapearce (changes requested):

README.md
starswan: we could do, but I don't actually know how mine happened?

beccapearce (changes requested):
I still think we we fix up the order of the readme, things like the yard documentation could be moved down or even removed.
And I still feel that the postgresql errors are more prominent than the actually install instructions, could we rejig it to have the install instructions more visible than the possible errors.

There's also a thing with the headings within the papertrail section, the jsonb subheading is larger than the papertrail heading it belongs to

beccapearce (approved):

mseedat-moj:
PR has quite a few additional files which may or may not be desired @beccapearce @starswan ?

PR #694, opened 2019-03-05
CT 2119 Time taken displaying incorrectly

mseedat-moj:
### Summary
Changes the way `time_taken` to close a case is calculated. The number of business days now includes the day the case was actually responded. This means all existing reports where `time_taken` was used to output the number of business days taken to respond to a case was incorrect by 1 day.

### DIFF
--- a/spec/support/features/steps/cases/responding_steps.rb
+++ b/spec/support/features/steps/cases/responding_steps.rb
@@ -85,6 +85,8 @@ def close_sar_case_step(timeliness: 'in time', tmm: false, editable: true)
     .to eq Date.today.strftime(Settings.default_date_format)
   expect(show_page.response_details.timeliness.data.text)
     .to eq "Answered #{timeliness}"
+
+  # Regex required to handle both cases of 1 or more days to respond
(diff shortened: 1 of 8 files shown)


spec/support/features/steps/cases/responding_steps.rb
```diff
@@ -85,7 +85,7 @@ def close_sar_case_step(timeliness: 'in time', tmm: false, editable: true)
     .to eq Date.today.strftime(Settings.default_date_format)
   expect(show_page.response_details.timeliness.data.text)
     .to eq "Answered #{timeliness}"
-  expect(show_page.response_details.time_taken.data.text).to match(/\d+ working days/)
+  expect(show_page.response_details.time_taken.data.text).to match(/\d+ working day[s{1}]?/)
```
starswan: It would be nice to have a comment here explaining this

starswan (changes requested):
One little minor test needing a comment. It's tempting to suggest removing the duplication of asserts?

starswan (approved):

PR #699, opened 2019-03-12
Ct 2143 spike excel rag reporting

starswan:
Spike - use of XLSX library to generate excel spreadsheets. All code currently in controller - needs to wait for performance improvements so that reports can be run as inline CSVs. Once this happens code can run to_csv on the query items, and then convert to XLSX format rather than loading and parsing a CSV string as here.

### DIFF
--- a/app/controllers/stats_controller.rb
+++ b/app/controllers/stats_controller.rb
@@ -65,15 +65,15 @@ class StatsController < ApplicationController
 
   def cell_colour(value, thresholds)
     if value < thresholds[:red]
-               'FF0000'
-             elsif value < thresholds[:amber]
-               'FF0000'
-             else
-               '00FF00'
-             end
+      'FF0000'
+    elsif value < thresholds[:amber]
+      'FF0000'
+    else
+      '00FF00'
+    end
   end
 
-  def download_custom_report
+  def download_custom_report #rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength
     report= Report.find(params[:id])
     filename = report.report_type.filename
 
--- a/app/controllers/stats_controller.rb
+++ b/app/controllers/stats_controller.rb
@@ -57,21 +57,58 @@ class StatsController < ApplicationController
     end
   end
 
+  # RAG (Red, Amber, Green) thresholds for the different types
+  # Columns E, K and Q are percentages (non-trigger, trigger and overall)
+  # These are row items 4, 11 and 16. Data starts at row 4 (3 zero indexed)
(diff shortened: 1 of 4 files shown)


app/controllers/stats_controller.rb
```diff
@@ -57,11 +57,67 @@ def create_custom_report
...
+    report = Report.find(params[:id])
     filename = report.report_type.filename
 
-    send_data report.report_data, {filename: filename, disposition: :attachment}
+    # This is a Spike - so just hack the report about to show that an Excel report is possible
+    if report.report_type.abbr == 'R003'
+      # split report data into pieces - convert "" into empty string
+      excel_data = report.report_data.split("\n").map { |j| j.split(',') }
+      # Hack alert - SAR/FOI have different thresholds. This isn't really a good way of telling them apart
+      is_foi_report = excel_data[0][0].include?('FOI')
```
beccapearce: I think SAR reports are r103 so we might be able use the report type to select the thresholds?

app/controllers/stats_controller.rb
```diff
@@ -57,11 +57,67 @@ def create_custom_report
...
+  RAG_THRESHOLDS_FOI = {red: 85, amber: 90}
+  RAG_THRESHOLDS_SAR = {red: 80, amber: 85}
+
+  COLOURED_ROWS = {'E' => 4, 'K' => 11, 'Q' => 16}
+
+  def cell_colour(value, thresholds)
+    if value < thresholds[:red]
+      'FF0000' #Red - bright
+    elsif value < thresholds[:amber]
+      'FFFF00' #Yellow - again very bright
```
beccapearce: Oh so bright ✨🤩

app/controllers/stats_controller.rb
```diff
@@ -57,11 +57,67 @@ def create_custom_report
...
+    end
+  end
+
+  def download_custom_report #rubocop:disable Metrics/MethodLength
+    report = Report.find(params[:id])
     filename = report.report_type.filename
 
-    send_data report.report_data, {filename: filename, disposition: :attachment}
+    # This is a Spike - so just hack the report about to show that an Excel report is possible
+    if report.report_type.abbr == 'R003'
```
beccapearce: I think this only makes the foi business unit report into excel, did we want to do the SAR one too as part of this ticket?

beccapearce (changes requested):
Looks good, just wondered if we could use the report types to set the thresholds.

Also wondering are we looking to merge this as a spike into prod and then refactor later?

PR #702, opened 2019-03-15
CT-2114 remove scheduled jobs, download report now runs immediately

starswan:
remove sidekiq-scheduler and ReportGeneratorJob, convert stats controller to only run and not download a pre-canned version


mseedat-moj (approved):
Prodigious deletion. Good changes 👍

PR #706, opened 2019-03-19
Ct 2110 fix incorrect SAR removal event name

starswan:


app/models/case_transition.rb
```diff
@@ -24,9 +24,16 @@ class CaseTransition < ActiveRecord::Base
              class_name: 'Case::Base',
              foreign_key: :case_id
 
+  # This list should be bigger, but don't have time or inclination to move all event names here (yet)
```
mseedat-moj: 😂

I was only joking about a comment!

Ok as a compromise can you move these constants out of this file into somewhere in `initialisers` or something so we can consider using them across the site and possibly even in the state_machin/*.yml files!

spec/services/csv_exporter_spec.rb
```diff
@@ -107,15 +108,34 @@
...
+                                        'Extension Count' => 1
+                                      })
+        end
+      end
+      context 'extension removed' do
+        let!(:kase) do
+          create(:sar_case, :extended_deadline_sar,
+                 current_state: 'drafting',
+                            message: 'my SAR message',
+                            subject_full_name: 'Theresa Cant').tap do |k|
```
mseedat-moj: Civil Service is apolotical. Satire is frowned upon. Probably.

mseedat-moj (changes requested):
Uneasy about constants, compromise suggested

app/models/case_transition.rb
starswan: I thought the CaseTransition class seemed a reasonable place for these constants as they are all to do with case transitions...

mseedat-moj (approved):

PR #708, opened 2019-03-20
Ct-2124 Add report columns

mseedat-moj:
## Summary
Adds a number of new columns related to team names and timeliness. PR is against Case Creator work 

### Ticket
https://dsdmoj.atlassian.net/browse/CT-2124

## Notes
Unsure if the outputted values for timeliness are as requested by the user. Assume Team names are based on the `responder` team.

## Screenshots
Screen Shot 2019-03-20 at 03 24 54

#### Columns added:
```
    'Casework officer',
    'Created by',
    'Date created',
    'Business group',
    'Directorate name',
    'Director General name',
    'Director name',
    'Deputy Director name',
    'Draft in time',
    'In target',
    'Number of days late'
```


app/services/csv_exporter.rb
```diff
@@ -101,4 +130,30 @@ def dequote_and_truncate(text)
...
+    end
+  end
+
+  # Caseworker officer is blank for non-trigger and
+  # always a disclosure specialist for trigger cases.
+  # Catch exception from Case#assigned_disclosure_specialist
+  def casework_officer(kase)
+    return unless kase.workflow == 'trigger'
+
+    kase.assigned_disclosure_specialist.user.full_name rescue nil
```
starswan: really not sure about these rescue nil lines. Certainly the one in 'draft_in_time' doesn't look right. Why would any of these methods throw an exception, and if they did why would we want to ignore it?

mseedat-moj: In `case/base.rb` the method `assigned_disclosure_specialist` throws an exception if there are no assigning users:

```
  def assigned_disclosure_specialist
    ass = assignments.approving.accepted.detect{ |a| a.team_id == BusinessUnit.dacu_disclosure.id }
    raise 'No assigned disclosure specialist' if ass.nil? || ass.user.nil?
    ass.user
  end
```

In my opinion not being able to generate a report because of a single shoddy record seems unfair.

starswan: there shouldn't be any shoddy records. If there are, then this code just hides the bugs

app/services/csv_exporter.rb
```diff
@@ -101,4 +130,33 @@ def dequote_and_truncate(text)
...
+      days = (kase.date_draft_compliant - kase.internal_deadline).to_i
+      days > 0 ? days : nil
+    end
+  end
+
+  # Caseworker officer is blank for non-trigger and
+  # always a disclosure specialist for trigger cases.
+  # Note Case#assigned_disclosure_specialist throws an exception if no
+  # 'approving' assignees are found
+  def casework_officer(kase)
```
starswan: Both this method and the one below seem to know too much about the implementations of the methods they are calling. I think the functionality needs encapsulating - I don't think this code should know that trigger workflows result in assigned_disclosure_specialist not being valid?

app/services/csv_exporter.rb
```diff
@@ -101,4 +130,33 @@ def dequote_and_truncate(text)
...
+    kase.assigned_disclosure_specialist.user.full_name
+  end
+
+  def draft_in_time(kase)
+    return unless kase.respond_to?(:date_draft_compliant)
+
+    humanize_boolean(kase.within_draft_deadline?)
+  end
+
+  # Note Case#business_unit_responded_in_time? throws an exception if
```
starswan: I think the comment can go

spec/services/csv_exporter_spec.rb
```diff
@@ -1,30 +1,53 @@
...
   }
-  let(:sar_case) { create :closed_sar,
-                          name: 'SAR case name',
-                          postal_address: "2 High Street\nAnytown\nAY2 4FF",
-                          subject: 'Full details required',
-                          email: 'theresa@moj.com',
-                          message: 'my SAR message',
-                          subject_full_name: 'Theresa Cant'
+
+  # These columns can have arbitrary values as they are
```
starswan: rather than removing them, I think we should be setting these values explicitly when we care about them

starswan (changes requested):
some encapsulation of case methods required IMHO

app/services/csv_exporter.rb
moseeds: This was a requirement from the officer about how only to display a full_name if it's a trigger case. There's no suggestion that non-trigger cases should not have people assigned/names for display except for this particular report.

spec/services/csv_exporter_spec.rb
moseeds: They aren't being removed from checking. They are checked separately without having to override the FactoryBot create codes.

app/services/csv_exporter.rb
mseedat-moj: This was a requirement from the officer about how only to display a full_name if it's a trigger case. There's no suggestion that non-trigger cases should not have people assigned/names for display except for this particular report.

spec/services/csv_exporter_spec.rb
mseedat-moj: The columns are still checked without having to explicitly set teams for the sake of the team_name which varies by Id number only.

app/services/csv_exporter.rb
starswan: I think this points to issues with the kase#assigned_disclosure_specialist method. If there shouldn't be one (which is true for non-trigger cases) it seems find for this method to return nil rather than throw something nasty. That way the knowledge stays encapsulated within Case::Base where it belongs

spec/support/features/steps/cases/delete_steps.rb
```diff
@@ -9,10 +9,10 @@ def delete_case_step(kase:)
 
   confirm_destroy_page.confirm_button.click
   # Without a reason for deletion, page debounces back to user
-  expect(confirm_destroy_page).to have_reason_for_deletion_error
-
-  confirm_destroy_page.fill_in_delete_reason
-  confirm_destroy_page.confirm_button.click
+  # expect(confirm_destroy_page).to have_reason_for_deletion_error
```
starswan: Please put these back

starswan (approved):

spec/support/features/steps/cases/delete_steps.rb
mseedat-moj: OK!

app/services/csv_exporter.rb
mseedat-moj: Good discussion. Ended up creating a 'non bang!' version of the method which returns `nil` when no user is available rather than throw an exception.

PR #709, opened 2019-03-20
CT-2124 Save case creator

mseedat-moj:
## Summary
To allow Case download reports to show which user created the case, the logged in user must be assigned to the saved case. This is missing functionality that has been added in this PR.

## Ticket
https://dsdmoj.atlassian.net/browse/CT-2124

## Notes
All cases must require a user (creator). Retrospective cases are assigned to a mock 'Default User' with a permanent id of -100 and a blank name.

The changes to the rest of the code base are required because of the decision to ensure all cases are saved with a `creator`


app/models/case/base.rb
```diff
@@ -26,6 +26,7 @@
 #  type                 :string
 #  appeal_outcome_id    :integer
 #  dirty                :boolean          default(FALSE)
+#  user_id              :integer          not null, default(-100), foreign key
```
starswan: Shouldn't this be creator_id?

db/migrate/20190312104101_add_user_to_case.rb
```diff
@@ -0,0 +1,15 @@
+class AddUserToCase < ActiveRecord::Migration[5.0]
+  # To allow Case reports to show which user created the case.
+  # For existing cases, a 'dummy' user must be created with a fixed ID.
+  # The user creation is done here rather than data_migrations to ensure
+  # the DefaultUser exists before the column is created.
+  def up
+    DefaultUser.build!
```
starswan: As the default user is only used during the migration, wouldn't it be better for it to be in here? Actually I don't think we need a default user class at all - simply create a user in the migration, save with validation: false and we're done.

lib/cts/cases/create.rb
```diff
@@ -165,20 +169,23 @@ def get_created_at_date
...
+
       @clear_cases = options.fetch(:clear, false)
       @dry_run = options.fetch(:dry_run, false)
       @created_at = options[:created_at]
+
       if options[:received_date]
         @received_date = options[:received_date]
       elsif @created_at.present? &&
-            DateTime.parse(@created_at) < DateTime.now
+        DateTime.parse(@created_at) < DateTime.now
```
starswan: I think this is the wrong indentation - the original was correct

lib/cts/cases/create.rb
```diff
@@ -366,30 +373,36 @@ def get_journey_for_flagged_state(flag)
...
-                       end
-                     else
-                       CTS::find_user(options[:responder])
-                     end
+      @responder ||= begin
+        if !options.key?(:responder)
+          if responding_team.responders.empty?
+            raise "Responding team '#{responding_team.name}' has no responders."
+          else
+            responding_team.responders.first
```
starswan: responding_team.responders.first!

spec/models/case/overturned_ico/sar_spec.rb
```diff
@@ -82,14 +73,13 @@
                       'email'                   => 'stephen@stephenrichards.eu',
                       'ico_officer_name'        => 'Dan Dare'
               }).to_unsafe_hash
+          params[:creator] = creator
```
starswan: This isn't needed IMHO - actually we have already created, so the validity test isn't very useful. build() is the idiomatic pattern here

spec/support/db_housekeeping.rb
```diff
@@ -20,7 +20,7 @@ def self.clean(seed: true)
       search_queries
     )
     tables.each do |table|
-      ActiveRecord::Base.connection.execute("TRUNCATE #{table}")
+      ActiveRecord::Base.connection.execute("TRUNCATE #{table} CASCADE")
```
starswan: There is a DatabaseCleaner gem available...

starswan (changes requested):
Unsure about the DefaultUser class in the main codebase

app/models/case/base.rb
mseedat-moj: I thought these were the actual database column names rather than the virtual names

spec/support/db_housekeeping.rb
mseedat-moj: This minor change is good enough unless there's a compelling reason to use the gem?

app/models/case/base.rb
starswan: yes they are. Would still make sense to call it creator_id IMHO. What if we needed an actual user_id for some reason?

lib/cts/cases/create.rb
mseedat-moj: Just re-wrote the method to make intent clearer

lib/cts/cases/cli.rb
```diff
@@ -149,10 +152,18 @@ def create(*target_states)  # rubocop:disable Metrics/CyclomaticComplexity, Metr
       if options.key? :received_date
         CTS.info "Setting received date to: #{options[:received_date]}"
       end
+      if options.key? :creator
```
starswan: oops - you call it :user in your options declaration, but its :creator here. I think we might need some more tests in this area?

lib/cts/cases/create.rb
```diff
@@ -366,30 +372,32 @@ def get_journey_for_flagged_state(flag)
...
-                             if options.key?(:responder)
-                               responder.responding_teams.first
-                             else
-                               BusinessUnit.responding.sample
-                             end
-                           else
-                             CTS::find_team(options[:responding_team])
-                           end
+      @responding_team ||= begin
+        if !options.key?(:responding_team)
```
starswan: while you're here, it would be good to flip the ! again. I'm amazed that rubocop doesn't detect this - !x feels like an anit-pattern in every language that I have ever used as it is so small and hard to spot

spec/models/case/base_spec.rb
```diff
@@ -26,6 +26,7 @@
 #  type                 :string
 #  appeal_outcome_id    :integer
 #  dirty                :boolean          default(FALSE)
+#  user_id              :integer          not null, default(-100), foreign key
```
starswan: How does this get added to the header?

starswan (changes requested):
One minor thing left

app/models/case/base.rb
mseedat-moj: Trying not to think that far ahead!

lib/cts/cases/cli.rb
mseedat-moj: Good spot

lib/cts/cases/create.rb
mseedat-moj: Ok

spec/models/case/base_spec.rb
mseedat-moj: Manually? It's all horribly out of date...

starswan (approved):

lib/cts/cases/create.rb
mseedat-moj: I took this comment to mean I have liberty to reformat the entire file. Sorry.

PR #710, opened 2019-03-20
CT-2152 Add Circle CI config file with branch build whitelist

mseedat-moj:
## Summary
Cloud Platform Migration requires moving off Jenkins, ideally Circle CI. To stop Circle CI builds failing this config will only build (during test phase) on whitelisted branches

## Ticket
https://dsdmoj.atlassian.net/browse/CT-2139


starswan (commented):
Thought the circle config file was called .circleci/config.yml?

starswan (changes requested):
I think the circle:ci config file is called .circleci/config.yml - so this is in the wrong place?

starswan (approved):

PR #711, opened 2019-03-20
Ct 2150 multi role user unable to post to trigger cases

starswan:
When a user has both a manager and a lesser role, when adding messages to a case it needs to be actioned as the manager role


mseedat-moj (approved):
Well caught!

PR #713, opened 2019-03-21
CT-2134 Excel rag reports

starswan:
RAG reporting for business unit, monthly and appeals reports. Those reports have been converted to Excel format


app/controllers/stats_controller.rb
```diff
@@ -2,29 +2,42 @@ class StatsController < ApplicationController
...
+                filename: report.report_type.filename('xlsx'),
+                disposition: :attachment,
+                type: SPREADSHEET_CONTENT_TYPE
+    else
+      report.run_and_update!
+      report.trim_older_reports
+
+      send_data report.report_data, filename: report.report_type.filename('csv')
+    end
   end
```
mseedat-moj: The `if report.xlsx? ... else ... end` condition suggests this chunk of code should be inside the `Report` object or another place perhaps, or maybe 2 classes: `ExcelReport` and `CSVReport` ? Feel uneasy about this block being in controller

app/controllers/stats_controller.rb
```diff
@@ -34,31 +47,78 @@ def create_custom_report
...
+  BRIGHT_RED = 'FF0000'
+  BRIGHT_YELLOW = 'FFFF00'
+  BRIGHT_LIME_GREEN = '00FF00'
+
+  # mapping of RAG ratings to spreadsheet cell colours
+  RAG_RATING_COLOURS = { red: BRIGHT_RED, amber: BRIGHT_YELLOW, green: BRIGHT_LIME_GREEN }
+
+  # Assumes no report spans more than 26 columns
+  SPREADSHEET_COLUMN_NAMES = ('A'..'Z').to_a
+
```
mseedat-moj: Is this the right place for these settings? Perhaps related to the comment above regarding `ExcelReport` class?

app/models/report.rb
```diff
@@ -35,8 +38,13 @@ def report_type_abbr=(abbr)
   def run(*args)
     report_service = report_type.class_constant.new(*args)
     report_service.run
+    report_service
+  end
```
mseedat-moj: I like returning self, argument here to ensure all the `base_<whatever>` reports and child classes return self in `def run`?

app/models/report.rb
```diff
@@ -47,8 +55,20 @@ def trim_older_reports
...
+
   private
 
+  def generate_csv(report_service)
+    CSV.generate(headers: true) do |csv_generator|
+      report_service.to_csv.each do |csv_row|
+        csv_generator << csv_row.map(&:value)
+      end
+    end
+  end
```
mseedat-moj: Very good

app/services/stats/r006_kilo_map.rb
```diff
@@ -18,24 +18,25 @@ def self.description
       'Includes a list of all teams and users that respond to requests for information'
     end
 
-
+    # KiloMap isn't really a 'Report' in the normal sense of the word
```
mseedat-moj: In which sense of the word!?

app/services/stats/stats_collector.rb
```diff
@@ -55,36 +55,50 @@ def check_row_and_col_exist(row, col)
...
+
+      def initialize(stats, column_hash, first_column_header, superheadings, row_names_as_first_column)
+        @stats = stats
+        @column_hash = column_hash
+        @first_column_header = first_column_header
+        @superheadings = superheadings
+        @row_names_as_first_column = row_names_as_first_column
+      end
+
+      def each
```
mseedat-moj: An explanation of how this method works would be wonderful

mseedat-moj (changes requested):
Really good work, some initial feedback to turn spike into fully fledged release.

app/controllers/stats_controller.rb
starswan: report shouldn't be a model - its more like an 'action'. Felt this was the best solution rather than a complete re-write.

app/controllers/stats_controller.rb
starswan: Yes these might have been better in the view, but it's not easy to do given our current flow. It's more normal to know what your expected type is when you start the action - in 'custom report type' I don't think we know (unless maybe its always XLSX?) so the normal rails view pattern doesn't work

app/models/report.rb
starswan: nope - we need to return the 'report service' here (which is the derived report class). This whole structure is backwards (starting from the poor design making Report a model_

app/services/stats/r006_kilo_map.rb
starswan: Because its just a 'CSV output' thing, and it doesn't use the 'Report' and 'ReportType' structures used by the other reports

starswan: I've deleted the comment - I agree its not that useful

mseedat-moj (approved):

PR #714, opened 2019-03-21
CT-2148 closed cases turn up in my open cases tab

starswan:


mseedat-moj (approved):

mseedat-moj (dismissed):
Request from Stephen to un-approve

app/services/case_finder_service.rb
```diff
@@ -120,19 +122,20 @@ def in_states(states)
     scope.in_states(states.split(','))
   end
 
+  CASE_TYPES_ONLY_VISIBLE_WITH_FURTHER_CLEARANCE = [
```
beccapearce: Not sure about what this is doing, could we put a comment in?

spec/controllers/cases_controller_spec.rb
```diff
@@ -446,6 +443,26 @@ def sar_closure_params(sar)
...
+      let!(:in_time_case) { create(:accepted_case) }
+      let!(:late_case) { Timecop.freeze(50.days.ago) { create(:accepted_case) } }
+      let!(:closed_case) { create(:closed_case) }
+
+      before do
+        sign_in responder
+      end
+
+      it 'gets my in time cases' do
+        get :my_open_cases, params: { tab: 'in_time' }
```
beccapearce: do we have a similar test for incoming cases? (I'm assuming that's what page was reported in the ticket)

beccapearce (changes requested):
Just a comment and test request

app/services/case_finder_service.rb
starswan: I just moved it from a dynamic attribute to a static one whilst in the code. I don't really know any more

spec/controllers/cases_controller_spec.rb
```diff
@@ -443,36 +444,34 @@ def sar_closure_params(sar)
     end
 
     context 'without stubs' do
-      let!(:in_time_case) { create(:accepted_case) }
-      let!(:late_case) { Timecop.freeze(50.days.ago) { create(:accepted_case) } }
+      # let!(:in_time_case) { create(:accepted_case) }
```
beccapearce: Do we still need this code?

beccapearce (changes requested):
A few commented bits of code left in but otherwise looks good

spec/controllers/cases_controller_spec.rb
starswan: no clearly not :-)

beccapearce (approved):

PR #716, opened 2019-03-22
CT-2141 Removed last vestiges of DACU from front end

starswan:


config/locales/en.yml
```diff
@@ -370,7 +370,7 @@ en:
         message: Full request
         postal_address: Address
         flag_for_disclosure_specialists:
-          Clearance required by DACU Disclosure?
+          Clearance required by Disclosure BMT?
```
beccapearce: This should just be disclosure not disclosure bmt

config/locales/state.en.yml
```diff
@@ -5,7 +5,7 @@ en:
     awaiting_responder_email: Waiting to be accepted
     awaiting_dispatch: Ready to send
     drafting: Draft in progress
-    pending_dacu_clearance: Pending clearance
+    pending_dacu_clearance: Pending clearance - Disclosure
```
beccapearce: Do we want to change this? It will change it on all the case listings

beccapearce (changes requested):
Replace Disclosure with disclosure BMT also a question about changing the state file

config/locales/state.en.yml
mseedat-moj: Reverted so available on rebase to master

beccapearce (approved):

PR #725, opened 2019-03-28
CT-2133 Enable edit specs

starswan:
Enabled testing of edit for more case types


mseedat-moj (approved):
Understand this change demonstrates incomplete functionality and therefore a previously failing test. Test code has been revised to explicitly name parameters rather than relying on `args`

PR #726, opened 2019-03-28
CT-2173 take out eager loads of lib, fix up pit extension migration

starswan:
Hopefully these 2 small fixes will bring back the dev environment into a known good state


mseedat-moj (approved):
My only request is to add some context in the PR as it's not clear what the actual issue is

PR #733, opened 2019-04-02
CT-2175 Deactivated Team Fix and gem updates [bug]

mseedat-moj:
## Summary
Includes 3 main changes:
1. RSpec 3.8.2 Gem upgrade required some test/expectation changes: 
2. Bug fix for `Team#original_team_name` 
3. Dependabot based selected Gems/Gemfile.lock upgrades

## Ticket
https://dsdmoj.atlassian.net/browse/CT-2175

### 1. RSpec 3.8.2 Gem upgrade required some test/expectation changes: 
`have_enqued_job` behaviour has changed in RSpec 3.8.2 as described here: https://www.rubydoc.info/gems/rspec-rails/file/Changelog.md
> Backport: Make the ActiveJob matchers fail when multiple jobs are queued for negated matches.

### 2. Bug fix for `Team#original_team_name` 

When a business unit is deactivated, the team name is changed to include Timezone information in string format. Due to British Summer Time this broke tests on servers using UTC.

### 3. Dependabot suggested Gem upgrades

Due to the above issues, rather than tackling PRs piecemeal and to allow investigation of the above bugs in a sensible manner, this PR supersedes individual dependabot generated PRs for the following:

1. https://github.com/ministryofjustice/correspondence_tool_staff/pull/732/files
2. https://github.com/ministryofjustice/correspondence_tool_staff/pull/731/files
3. https://github.com/ministryofjustice/correspondence_tool_staff/pull/730/files
4. https://github.com/ministryofjustice/correspondence_tool_staff/pull/729/files
5. https://github.com/ministryofjustice/correspondence_tool_staff/pull/646/files
6. https://github.com/ministryofjustice/correspondence_tool_staff/pull/600/files


beccapearce:
The error is in the current master branch. That's why I haven't merged them

app/models/team.rb
```diff
@@ -133,21 +135,22 @@ def inactive?
...
 
+  # @note (Mohammed Seedat 2019-04-03) Original Team Name workaround
+  #   Team table has no way of storing historical data. On deactivation, the
+  #   Team's name is changed - name therefore stores deletion information.
+  #   To retrieve the Team name without any of the  deactivation information
+  #   we require this string replacement workaround
   def original_team_name
-    team_name = name.remove('DEACTIVATED')
-    team_name.remove(deleted_at.to_s)
+    name.remove(DEACTIVATED_LABEL, /@\((.)*\)/).strip
```
starswan: I think this regex needs a comment...?

app/services/team_deletion_service.rb
```diff
@@ -21,10 +21,16 @@ def call
...
+  #   ease purposes. Note use of brackets to allow extraction of team name
+  #   without additional information when required.
   def soft_delete
-    @team.update_attributes!(deleted_at: Time.current,
-                             name: "DEACTIVATED #{@team.name} " + Time.now.to_s)
-  end
-
+    deletion_date = DateTime.now.strftime('%F %T')
 
+    @team.update_attributes!(
```
starswan: I think we need a simpler design. As we know the team is deleted, we could simply override the team_name property for a deleted team rather than actually updating it in the DB, and then trying to parse it out again.

starswan (changes requested):
Think the team name should be altered dynamically in the code rather than altered in the DB and then parsed out afterwards. Much simpler and more robust

app/services/team_deletion_service.rb
mseedat-moj: Out of scope

app/models/team.rb
mseedat-moj: Added additional note to explain string format scheme

starswan (approved):

PR #748, opened 2019-04-16
CT-2182 Refactor and fix reporting period range

mseedat-moj:
### Summary
Performance reports are all generated on-request. The date ranges used in the required SQL query used `Date` rather than `Time` and therefore Cases generated on the day of the report-request were not being included in the query result set (due to the query restricting results by `created_at - timestamp type and requested_date - date type)`. This PR ensures `period_start` and `period_end` includes the full day.

This PR also takes the opportunity to refactor the reporting period code as part of general product healthiness and enforces `Timecop` thread safe behaviour throughout the testing code base to prevent intermittent build errors.

### Ticket
https://dsdmoj.atlassian.net/browse/CT-2182

### Notes
- `ReportingPeriodCalculator` has been split into a number of classes broadly representing a 'strategy pattern'
- Rails extension methods `end_of_day` and `beginning_of_day` are used to ensure the report generation SQL query includes full days. These Times include timezone information but issues with relation to BST/GMT are not covered by this PR as they are considered extreme edge cases
- Building on Semaphore CI fails intermittently due to Timecop thread safety issues. Therefore all Timecop.freeze usage is in `Timecop.freeze do ... end` blocks and existing tests have been amended where necessary. This is to stop parallel testing failing due to the lack of thread safety when calling `Timecop.freeze`.


spec/models/case/scopes_spec.rb
```diff
@@ -94,11 +94,18 @@
...
+
+      Timecop.freeze(1.minute.ago) do
+        open_case = create :case
+      end
+
+      Timecop.freeze(2.minutes.ago) do
+        responded_case = create :responded_case
+      end
+
       Timecop.return
```
stephenrichards: I don't think you need the `Timecop.return` if your using the `Timecop.freeeze do....end` form.

stephenrichards (changes requested):
Looks good to me.  I think there is one superfluous `Timecop.return`, but apart from that - ready to go.

stephenrichards (approved):

PR #749, opened 2019-04-16
CT-2069 Move delete button and make proper buttons

digitalali-moj:
Done a merge instead of rebase here by following instructions on GIT and knowing it's safe. But normally I'd rebase.


mseedat-moj (approved):
Looks good, sensible approach

PR #758, opened 2019-04-23
CT-2196 Ensure unassigned FOI and SAR are included in performance report

mseedat-moj:
### Summary
Performance reports now include `unassigned` cases (FOI or SAR) when calculating various values ('open in time', 'responded late', etc).

### Ticket
https://dsdmoj.atlassian.net/browse/CT-2196


app/models/case/base.rb
```diff
@@ -734,11 +734,16 @@ def remove_pit_deadline!(initial_deadline)
...
   end
 
-
   def trigger?
     TRIGGER_WORKFLOWS.include?(workflow)
   end
 
+  # @todo (Mohammed Seedat): require clarification as to whether a 'flagged'
+  #   case and 'trigger?' case are the same thing
+  def trigger_status
```
stephenrichards: Yes, trigger and flagged are the same thing.

app/models/case/base.rb
```diff
@@ -572,7 +572,7 @@ def responded_in_time_for_stats_purposes?
   end
 
   def already_late?
-    Date.today > external_deadline
+    Date.current > external_deadline
   end
```
stephenrichards: Nice.  I didn't know the difference between `Date.today` and `Date.current` until recently.

app/services/stats/base_monthly_performance_report.rb
```diff
@@ -2,12 +2,14 @@ module Stats
...
+      month:    'Month'
+    }.freeze
 
     R005_SPECIFIC_SUPERHEADINGS = {
-        month:     ''
-    }
+      month:     ''
+    }.freeze
+
+    INDEXES_FOR_PERCENTAGE_COLUMNS = [1, 7, 13].freeze
```
stephenrichards: nice.  I think there is a rubocop rule that you can add which will insist on all constants being frozen.

app/services/stats/case_analyser.rb
```diff
@@ -45,6 +45,11 @@ class CaseAnalyser
 
     }.freeze
 
+    RESPONDED_IN_TIME = :responded_in_time
+    RESPONDED_LATE    = :responded_late
+    OPEN_LATE         = :open_late
+    OPEN_IN_TIME      = :open_in_time
+
```
stephenrichards: freeze constants.

PR #760, opened 2019-04-25
CT-2067 Move take on and escalate into action bar at top

digitalali-moj:
Also removed proxy partial that's not being used due to simpler
formatting. (not requiring title/sectino)


app/views/cases/show.html.slim
```diff
@@ -30,6 +29,10 @@ div id="case-#{@correspondence_type_key}" class="case"
         - if policy(@case).can_respond?
           = action_button_for(:respond)
 
+        - if policy(@case).can_accept_or_reject_approver_assignment?
+          = render partial: 'take_case_on_or_de_escalate',
+          locals: { case_details: @case }
```
mseedat-moj: The formatting is a bit off on this. Ideally should be:

```
= render partial: 'take_case_on_or_de_escalate', locals: { case_details: @case }    
```

or if that results in a line that goes beyond 80 characters then:

```
= render partial: 'take_case_on_or_de_escalate',
     locals: { case_details: @case }    
```

mseedat-moj (changes requested):

app/views/cases/show.html.slim
digitalali-moj: Added indentation here to be consistent with formatting.

mseedat-moj (approved):

PR #767, opened 2019-04-29
CT-2070 Move extend for PIT and make button and update tests

digitalali-moj:


app/helpers/cases_helper.rb
```diff
@@ -189,6 +189,15 @@ def action_links_for_allowed_events(kase, *events)
...
 
+  def action_buttons_for_allowed_events(kase, *events)
+    allowed_events = events.find_all do |event_name|
+      policy(kase).__send__ "#{event_name}?"
+    end
+    allowed_events.map do |event_name|
+      __send__ "action_button_for_#{event_name}", kase
+    end
+  end  
+
```
mseedat-moj: Can rewrite the above code to be more 'ruby' style:

```
events
  .find_all { |name| policy(kase).send("#{name}?") } 
  .map  { |name| send("action_button_for_#{name}", kase) }
```

Is this new code? If so should have a corresponding test ideally.

app/views/cases/show.html.slim
```diff
@@ -43,6 +43,10 @@ div id="case-#{@correspondence_type_key}" class="case"
         - if policy(@case).assignments_execute_reassign_user?
           = action_button_for(:reassign_user)
 
+      = action_buttons_for_allowed_events(@case, :extend_for_pit, :remove_pit_extension).join(' ').html_safe
+
+      - if (@filtered_permitted_events - [:edit_case, :destroy_case]).any?
+        
         - if policy(@case).destroy_case?
           = action_button_for_destroy_case(@case)
```
mseedat-moj: No need for the extra line-break on Line 49.

mseedat-moj (changes requested):
Almost there, some suggested changes and possibly require a test if it does not already exist for the method in cases_helper.rb

app/helpers/cases_helper.rb
digitalali-moj: Used this slick version instead :) 
Used existing test, but let me know if we need a more generic test, rather than testing the button that's generated.

mseedat-moj (approved):
Good work.

PR #768, opened 2019-04-30
CT-2207 Implement working closed cases using Stats system

mseedat-moj:
### Summary
Allow Closed Cases to be downloaded from the Performance Section of the application.

### Ticket
https://dsdmoj.atlassian.net/browse/CT-2207

### Notes
Closed Cases are download directly as a `cases_controller` response type (csv).

This PR turns `Closed Cases` into another `ReportType` and treats Closed Cases as other Performance Reports. This will allow output of the report in Excel format if desired with better formatting.

This implementation is a departure from the requirements to prevent code duplication and to prevent the introduction of a 3rd way of outputting reports.

This PR does not address performance issues causing server timeouts.

`BaseReport.rb` now uses the `**options` splat argument to accept any (named) parameters. This has required corresponding changes in `BaseReport` sub-classes.

### Features
1. Default reporting period is 'Last Month' via a new `ReportingPeriod::LastMonth` period - however not currently displayed as a download option due to requirements change
2. Creating a 'custom report' now shows 'Closed Cases' as an option


app/services/stats/r007_closed_cases_report.rb
```diff
@@ -0,0 +1,27 @@
...
+      closed_scope = Case::Base.presented_as_closed
+
+      if @user.responder_only?
+        case_ids = Assignment.with_teams(@user.responding_teams).pluck(:case_id)
+        closed_scope.where(id: case_ids).most_recent_first
+      else
+        closed_scope.most_recent_first
+      end
+    end
+
```
Polsonby: I'm probably missing something but I don't think we have test coverage for the `if` and the `else` branches here?

This report seems to need to know a lot about the world - would it make sense to move that `Assignment.with_teams...pluck(:case_id)` into a private method or a method on User?

spec/features/user_journeys/foi/downloading_stats_spec.rb
```diff
@@ -38,7 +38,7 @@
     create :report_type, :r005
     create :report_type, :r102
     create :report_type, :r103
-    create :r105_report_type
+    create :report_type, :r105
```
Polsonby: Nice

app/controllers/stats_controller.rb
```diff
@@ -38,35 +40,25 @@ def download_audit
...
       if @report.xlsx?
-        report_data = @report.run(@report.period_start, @report.period_end)
-
-        axlsx = create_spreadsheet(report_data)
-
-        send_data axlsx.to_stream.read,
-                  filename: @report.report_type.filename('xlsx'),
-                  disposition: :attachment,
-                  type: SPREADSHEET_CONTENT_TYPE
+        generate_custom_excel_report(@report)
```
Polsonby: nice

Polsonby (approved):
Nice work :) I understand this is a stepping stone to where you want to get to, but looks good

Just a couple questions - I see one spec is failing:
`spec/controllers/stats_controller/create_custom_report_spec.rb:106`

app/controllers/stats_controller.rb
```diff
@@ -1,12 +1,16 @@
...
 
   before_action :authorize_user
 
-  SPREADSHEET_CONTENT_TYPE = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'
+  SPREADSHEET_CONTENT_TYPE = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet'.freeze
 
   def index
     @foi_reports = ReportType.standard.foi.order(:full_name)
     @sar_reports = ReportType.standard.sar.order(:full_name)
+    @closed_reports = ReportType.closed_cases_report
```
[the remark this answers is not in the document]
mseedat-moj: Done as suggested

app/services/stats/r007_closed_cases_report.rb
mseedat-moj: I have reused the `CaseFinderService` existing code rather than copying in the code.

PR #769, opened 2019-05-01
CT-2067 Fix de-escalate to make button and change class

digitalali-moj:
### Summary
Following up from moving the link, this now makes the link a button. 

### Ticket
https://dsdmoj.atlassian.net/browse/CT-2067

### Notes
CSS class name is changed to remove `js` since this is not a JavaScript trigger. So updated tests accordingly.


mseedat-moj (approved):
Sensible changes

PR #778, opened 2019-05-09
Upgrade to Webdrivers gem and fix feature specs

Polsonby:
Attempting to get the feature specs working - particularly the JS ones - following the retiral of the chrome driver helper gem and move to webdrivers gem. 

Specs needing the view to be scrolled to find an element in order to click on it were failing. 

After much effort expended trying to figure out why this was no longer working as it had done previously (possibly a bug in the chromedriver or Chrome itself, or a change in behaviour) a solution was finally found, which is to use a relatively recently introduced method in Capybara to scroll the required element into the viewport before clicking on it

See https://stackoverflow.com/questions/55406656/why-is-there-no-scroll-down-up-functionality-with-selenium-webdriver
and
https://www.rubydoc.info/gems/capybara/Capybara/Node/Element#scroll_to-instance_method

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,7 +72,7 @@ group :test do
   gem 'codeclimate-test-reporter', '~> 1.0'
   gem 'i18n-tasks', '~> 0.9.29'
   gem 'rails-controller-testing', require: false
-  gem 'selenium-webdriver'
+  gem 'webdrivers', '~> 3.0'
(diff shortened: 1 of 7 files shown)


digitalali-moj (approved):
Works locally for me, talked through with Ian

digitalali-moj (approved):

PR #796, opened 2019-05-21
OFFENDER-SARS Setup new design system JavaScript and CSS

digitalali-moj:
Installs the dxw_govuk_frontend_rails gem to use the new govuk
design system. Imported into SASS and JS.


digitalali-moj:
No longer required as team decision is to not go with new design system

PR #797, opened 2019-05-22
Test installation of gem only to see if tests go bezerk

digitalali-moj:


digitalali-moj:
This was a trial no longer needed for testing

PR #802, opened 2019-05-31
Fix date-dependent spec failure

Polsonby:
The specs hard-code a date of May 30, 2018 - the app prevents you from creating a case that's more than a year old! So apply time cop freeze to allow specs to pass 

Subsequently had some parallel and order-dependent test failures


moseedat (approved):
ace!

mseedat-moj (approved):

PR #812, opened 2019-06-11
CT-2323 log in as branston user

Polsonby:
In the previous work, I'd added in a Correspondence Type for Offender SAR and manually added it to the list of new case types for every user, provided the feature flag was turned on. 

Now we're still respecting the feature flag, but I've also set up a basic team and user in the developer seeds so that we can log in as a Branston user. 

Credentials are for Brian Rix (i.e. Branston Registry user)
correspondence-staff-dev+brian.rix@digital.justice.gov.uk

## Jira Ticket
https://dsdmoj.atlassian.net/browse/CT-2323

### DIFF
--- a/db/seeders/correspondence_type_seeder.rb
+++ b/db/seeders/correspondence_type_seeder.rb
@@ -1,5 +1,6 @@
 class CorrespondenceTypeSeeder
 
+  #rubocop:disable Metrics/MethodLength
   def seed!
     puts "----Seeding Correspondence Types----"
 
@@ -42,6 +43,6 @@ class CorrespondenceTypeSeeder
                 internal_time_limit: 10,
                 external_time_limit: 20,
                 deadline_calculator_class: 'CalendarDays')
-
   end
+  #rubocop:enable Metrics/MethodLength
 end
--- a/db/seeders/correspondence_type_seeder.rb
+++ b/db/seeders/correspondence_type_seeder.rb
@@ -34,5 +34,14 @@ class CorrespondenceTypeSeeder
                 external_time_limit: 30,
                 deadline_calculator_class: 'CalendarDays')
 
+    rec = CorrespondenceType.find_by(abbreviation: 'OFFENDER')
+    rec = CorrespondenceType.new if rec.nil?
+    rec.update!(name: 'Offender SAR (OFFENDER)',
+                abbreviation: 'OFFENDER',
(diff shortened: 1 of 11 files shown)


db/seeders/correspondence_type_seeder.rb
```diff
@@ -34,5 +35,14 @@ def seed!
...
 
+    rec = CorrespondenceType.find_by(abbreviation: 'OFFENDER')
+    rec = CorrespondenceType.new if rec.nil?
+    rec.update!(name: 'Offender SAR (OFFENDER)',
+                abbreviation: 'OFFENDER',
+                escalation_time_limit: 3,
+                internal_time_limit: 10,
+                external_time_limit: 20,
+                deadline_calculator_class: 'CalendarDays')
   end
```
mseedat-moj: What is `rec`?

Is this duplicating the current Data Migration? (the `abbreviation` should be `OFFENDER_SAR`)

db/seeders/dev_team_seeder.rb
```diff
@@ -40,9 +40,11 @@ def find_or_create_business_unit(attributes)
 
   #rubocop:disable Metrics/MethodLength
   def add_business_units
-    @foi = CorrespondenceType.foi
-    @sar = CorrespondenceType.sar
-    @ico = CorrespondenceType.ico
+    @foi      = CorrespondenceType.foi
+    @sar      = CorrespondenceType.sar
+    @ico      = CorrespondenceType.ico
+    @offender = CorrespondenceType.offender_sar
```
mseedat-moj: Should be `@offender_sar` ?

mseedat-moj (changes requested):
Almost there, some minor alterations requested.

db/seeders/correspondence_type_seeder.rb
Polsonby: Yeah, I don't know what "rec" stands for - "record" perhaps? It's the same as all the other ones in this file, so just keeping it consistent

The abbreviation shouldn't be OFFENDER_SAR until your branch is merged :)

db/seeders/dev_team_seeder.rb
Polsonby: The abbreviation shouldn't be OFFENDER_SAR until your branch is merged :)

mseedat-moj (approved):

PR #822, opened 2019-06-21
CT-2280 offender sars validation

Polsonby:
## Description
Performing the correct minimal validation on each screen of the create case flow for Offender SAR

This is the reason why we have the form object; it does the work of displaying each step of the wizard and knows which validations are required on each step. We have all the validations in the Case::SAR::Offender model, but we have monkey patched ApplicationRecord with a `valid_attributes?` method which takes a hash of params to validate using the model's validators as opposed to calling `valid?` which validates all the attributes in one go. By definition we need to only validate certain fields at each step. 

Because of the way radio button groups work, nothing is sent in the params if neither button is selected, which defeats the partial validation which wants to validate only fields that are passed on each form submission. Accordingly, and rather annoyingly, we need to merge in some empty params if the radio group is empty.
 
## Self-review checklist

* [x] (1) Quick stakeholder demo done
* [ ] (2) Bug with before and after screenshots
* [x] (3) Tests passing
* [x] (7) Table indexes added if required
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2280

### Screenshots
validations latest


app/form_models/offender_sar_case_form.rb
```diff
@@ -27,10 +28,12 @@ class OffenderSARCaseForm
            :received_date_dd,
            :received_date_mm,
            :received_date_yyyy,
+           :received_date,
            :reply_method,
            :creator,
            :send_by_post?,
            :send_by_email?,
+           :foo,
```
mseedat-moj: Accidental?

app/form_models/offender_sar_case_form.rb
```diff
@@ -72,14 +75,40 @@ def valid?
...
+      params.delete("third_party_relationship") unless params["third_party"] == "true"
+
+      params.merge!("reply_method": "") unless params["reply_method"].present?
+      params.delete("email") unless params["reply_method"] == "send_by_email"
+      params.delete("postal_address") unless params["reply_method"] == "send_by_post"
+    when "date-received"
+      params.merge!("received_date" => "") unless params["received_date"].present?
+    end
+
+    params
```
mseedat-moj: Is there a cleaner way of doing this by having say `defaults` and merging that with params if possible?

app/form_models/offender_sar_case_form.rb
```diff
@@ -72,14 +75,40 @@ def valid?
...
+  end
+
   def build_case_from_session
-    values = @session[:offender_sar_state] || {}
+    values = @session[:offender_sar_state] || {date_of_birth: nil}
```
mseedat-moj: Space in braces `{ date_of_birth: nil }`

app/models/case/sar/offender.rb
```diff
@@ -1,33 +1,72 @@
-class Case::SAR::Offender < Case::SAR::Standard
+class Case::SAR::Offender < Case::Base
   class << self
     def type_abbreviation
       'OFFENDER'
```
mseedat-moj: Can this be changed to `OFFENDER_SAR`

app/models/case/sar/offender.rb
```diff
@@ -1,33 +1,72 @@
...
+                 previous_case_numbers: :string,
+                 other_subject_ids: :string,
+                 external_deadline: :date,
+                 date_of_birth: :date,
+                 third_party: :boolean,
+                 third_party_relationship: :string,
+                 subject_type: :string,
+                 received_date: :date,
+                 reply_method: :string,
+                 flag_for_disclosure_specialists: :boolean
```
mseedat-moj: Can the list of attributes be ordered alphabetically and do any of them require default values?

spec/form_models/offender_sar_case_form_spec.rb
```diff
@@ -63,23 +63,25 @@
 
   describe "#valid_attributes?" do
     context "when params is empty" do
-      it "returns true" do
-        expect(case_form.valid_attributes?({})).to be true
+      params = ActionController::Parameters.new( {} ).permit!
```
mseedat-moj: Extra spaces around braces?

mseedat-moj (changes requested):
Looks on the right track. Some initial cursory comments

app/form_models/offender_sar_case_form.rb
Polsonby: Indeed! Fixed, thanks :)

app/form_models/offender_sar_case_form.rb
Polsonby: Fixed thanks

app/models/case/sar/offender.rb
Polsonby: I've had a go at this and it breaks things; obviously I'm not changing it in all the right places...! But I see you're changing this in your PR and I think we should not change it here therefore

spec/form_models/offender_sar_case_form_spec.rb
Polsonby: Fixed, thanks!

app/models/case/sar/offender.rb
Polsonby: Changed, and no, I don't think so...?

app/form_models/offender_sar_case_form.rb
Polsonby: Thanks for the suggestion! I agree it was really hard to read previously - have had a go at using methods to clarify what's going on and added a comment, is this any better?

mseedat-moj (approved):
Looks good thanks for the changes

PR #826, opened 2019-07-02
CT-2342 assign new case to branston

Polsonby:
New Offender SAR cases got wired up to Disclosure BMT to start with, like every other case type. Which distressingly meant that you couldn't view a case you'd just created when logged in as a Branston user. Now amended so that everything else still defaults to the normal way, but giving cases the ability to specify an alternative default. 

## Deployment note
We need to add a code to the Branston team for this to work - probably need to add a data migration to ensure the Branston team is there and set the `code` field to BRANSTON

### DIFF
--- a/app/models/business_unit.rb
+++ b/app/models/business_unit.rb
@@ -110,6 +110,10 @@ class BusinessUnit < Team
     code == Settings.foi_cases.default_managing_team
   end
 
+  def self.dacu_branston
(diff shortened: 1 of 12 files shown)


mseedat-moj (approved):
Looks good

Polsonby:
@mseedat-moj Please can you re-review? Added a couple more commits since you approved

mseedat-moj:
Great glad the commented out tests are re-enabled

PR #828, opened 2019-07-02
Create pull request template for new PRs

Polsonby:
Added the template we agreed on a while back so it should appear automatically for new PRs :)


digitalali-moj (approved):
Didn't know you can do this, very cool. thanks.

PR #831, opened 2019-07-10
CT-2236 Allow Show Offender SAR for all users

mseedat-moj:
### Summary
During testing all logged in users can see Offender SARS. This is necessary during the pre-workflow state and permissions stage of development.


Polsonby (approved):
cunning :)

PR #846, opened 2019-07-16
CT-2366 Amend FoiController to allow non-selection of FOI Type

mseedat-moj:
### Description
Fixes 3 bugs/issues:
1. If during FOI creation, user does not select an FOI type and presses 'Save', the page will refresh without throwing an exception
3. Issue with ComplianceReview/TimelinessReview type creation resolved
 
## Self-review checklist

* [x] (3) Tests passing
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2366
 
1. Log in as any user
2. Click 'Create Case' and select 'FOI'
3. Try saving without entering any details at all

### DIFF
--- a/app/helpers/cases_helper.rb
+++ b/app/helpers/cases_helper.rb
@@ -285,7 +285,12 @@ module CasesHelper
     user.manager? && kase.closed? && !kase.allow_event?(user, :update_closure)
   end
 
+  # Note use of Case::FOI::Standard for Timeliness/ComplianceReview FOI cases
   def case_create_action(kase)
-    self.send("#{kase.model_name.singular}_index_path")
+    if kase.foi?
+      self.send("case_foi_standard_index_path")
+    else
+      self.send("#{kase.model_name.singular}_index_path")
+    end
   end
 end
--- a/app/controllers/cases/foi_controller.rb
+++ b/app/controllers/cases/foi_controller.rb
@@ -16,7 +16,9 @@ module Cases
     end
 
     def case_type
-      Case::FOI::Standard.factory(params.dig(@correspondence_type_key, 'type'))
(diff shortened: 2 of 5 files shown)


app/helpers/cases_helper.rb
```diff
@@ -285,7 +285,12 @@ def manager_updating_close_details_on_old_case?(user, kase)
...
   end
 
+  # Note use of Case::FOI::Standard for Timeliness/ComplianceReview FOI cases
   def case_create_action(kase)
-    self.send("#{kase.model_name.singular}_index_path")
+    if kase.foi?
+      self.send("case_foi_standard_index_path")
+    else
+      self.send("#{kase.model_name.singular}_index_path")
+    end
```
Polsonby: What do you think about getting rid of this conditional checking the type of the object by moving this logic into two methods, one on `Case::Base` and one on `Case::FOI::Standard` so you can just say `kase.case_create_action` ?

or maybe I am missing something - I'm not clear why we're doing this in the first place, TBH, so perhaps there's a good reason why it's this way

Polsonby (approved):
Good job 👍 

Just one suggestion/question

app/helpers/cases_helper.rb
mseedat-moj: Yes as you've noticed this is a total kludge which I have had to settle for this particular work around. It's a workaround because of the way `TimelinessReview` and `ComplianceReview` are not treated as fully bonafide resources still - i.e. they don't have their on `CRUD` routes - hence the check to process them using the `FOI::Standard -> create` route. It's also related to why the FOI Type selection radio buttons are not automatically selected when there are error messages displayed on FOI Case Creation.

PR #847, opened 2019-07-22
CT-2238 workflow states

Polsonby:
## Description
Implement support for the available states for Offender SARs and provide an interface to move it through the workflow states. 

## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots
state transitions

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2238
 
### Deployment
n/a
A newly created case should be able to be viewed and start off in "Data to be requested". You can then use the green button to move it through the states shown here. The current implementation throws an exception when you try to close the case - next ticket will get that working 

image

### DIFF
--- a/app/controllers/cases/case_transitions_controller.rb
+++ b/app/controllers/cases/case_transitions_controller.rb
@@ -4,26 +4,40 @@ module Cases
     before_action :set_date_of_birth_instance_var, except: []
 
     def mark_as_waiting_for_data
-      @case.state_machine.mark_as_waiting_for_data!(acting_user: current_user, acting_team: current_user.managing_teams.first)
+      @case.state_machine.mark_as_waiting_for_data!(params_for_transition)
       reload_case_page_on_success
     end
 
     def mark_as_ready_for_vetting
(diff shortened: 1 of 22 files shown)


app/controllers/cases/case_transitions_controller.rb
```diff
@@ -0,0 +1,10 @@
+module Cases
+  class CaseTransitionsController < CasesController
+    def create
+      @case = Case::Base.find(params[:case_id])
+      authorize @case
+
+      @case.state_machine.mark_as_waiting_for_data!(acting_user: current_user, acting_team: current_user.managing_teams.first)
+    end
```
Polsonby: @mseedat-moj Line 7 gives `NoMethodError: undefined method valid? for nil:NilClass` when you run `spec/controllers/cases/case_transitions_controller_spec.rb` - spent ages digging around trying to see what nil is getting valid? sent to it. Checked out the before filters in the model with no luck. Can you see why this is happening? I might be sending in the wrong things to `mark_as_waiting_for_data!` but it seemed about right to me.

PR #853, opened 2019-07-29
CT-2257 add note to offender sar

Polsonby:
## Description
Allow Branston users to add notes to cases that get displayed inline in the case history 

image

It validates for empty message using the HTML required attribute 

image

but shows a form validation error just like the conversation feature IF the browser doesn't respect the `, required: true` setting on the text field. 

image

Notes are stored using the `message` property in the JSON field for the case, and get displayed as the details element when case transition items render in the case history. Normal conversation messages are still filtered out based on their different event name, so if we were to add the conversation feature back for offender SAR, it should still work. 
 
## Self-review checklist

* [x] (3) Tests passing
  
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2257
 
### Deployment
No database malarkey
 
The task is to hide the conversation feature when viewing an offender SAR and instead display the "add note to case" form below the case history section. This should not show up for any other case type.

### DIFF
--- a/app/controllers/cases/notes_controller.rb
+++ b/app/controllers/cases/notes_controller.rb
@@ -1,0 +1,52 @@
+module Cases
+  class NotesController < ApplicationController
+    before_action :set_case, except: []
+    before_action :set_date_of_birth_instance_var, except: []
+
+    def create
+      begin
+        @case.state_machine.add_note_to_case!(
+          acting_user: current_user,
+          acting_team: case_team,
+          message: params[:case][:message_text]
+        )
+      rescue ActiveRecord::RecordInvalid => err
+        if err.record.errors.include?(:message)
+          flash[:case_errors] = { message_text: err.record.errors[:message] }
+        end
+      ensure
+        redirect_to case_path(@case, anchor: 'case-history')
+      end
+    end
+
+    private
+
+    # need to sort with manager first so that if we are both manager and something else, we don't
+    # try to execute the action with our lower authority (which might fail)
+    def case_team
+      User.sort_teams_by_roles(case_teams).first
+    end
+
+    def case_teams
+      if current_user.teams_for_case(@case).any?
+        current_user.teams_for_case(@case)
+      else
+        current_user.teams
+      end
+    end
+
+    def set_case
+      @case = Case::Base.find(params[:case_id])
+      authorize(@case, :can_add_note_to_case?)
+    end
+
+    # this method is here to fix an issue with the gov_uk_date_fields
+    # where the validation fails since the internal list of instance
+    # variables lacks the date_of_birth field from the json properties
+    #     NoMethodError: undefined method `valid?' for nil:NilClass
+    #     ./app/state_machines/configurable_state_machine/machine.rb:256
+    def set_date_of_birth_instance_var
+      @case.date_of_birth = @case.date_of_birth
+    end
+  end
+end
--- a/app/controllers/cases/notes_controller.rb
+++ b/app/controllers/cases/notes_controller.rb
@@ -1,44 +1,17 @@
 module Cases
   class NotesController < ApplicationController
+    include Notable
     include GovUKDateFixes
 
     before_action :set_case, :set_date_of_birth
 
     def create
-      begin
-        @case.state_machine.add_note_to_case!(
-          acting_user: current_user,
-          acting_team: case_team,
-          message: params[:case][:message_text]
-        )
-      rescue ActiveRecord::RecordInvalid => err
-        if err.record.errors.include?(:message)
-          flash[:case_errors] = { message_text: err.record.errors[:message] }
-        end
-      ensure
-        redirect_to case_path(@case, anchor: 'case-history')
-      end
-    end
-
-    private
-
-    # need to sort with manager first so that if we are both manager and something else, we don't
-    # try to execute the action with our lower authority (which might fail)
-    def case_team
-      User.sort_teams_by_roles(case_teams).first
-    end
-
-    def case_teams
-      if current_user.teams_for_case(@case).any?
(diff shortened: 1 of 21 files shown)


app/controllers/cases/notes_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+          message: params[:case][:message_text]
+        )
+      rescue ActiveRecord::RecordInvalid => err
+        if err.record.errors.include?(:message)
+          flash[:case_errors] = { message_text: err.record.errors[:message] }
+        end
+      ensure
+        redirect_to case_path(@case, anchor: 'case-history')
+      end
+    end
```
mseedat-moj: Existing pattern is to put the `create` block of code into a dedicated Service.

mseedat-moj (changes requested):

app/controllers/cases/notes_controller.rb
mseedat-moj: Also should this controller be in the `/controllers/cases` namespace so that the URL is `/cases/:id/notes` ?

app/controllers/cases/notes_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    # this method is here to fix an issue with the gov_uk_date_fields
+    # where the validation fails since the internal list of instance
+    # variables lacks the date_of_birth field from the json properties
+    #     NoMethodError: undefined method `valid?' for nil:NilClass
+    #     ./app/state_machines/configurable_state_machine/machine.rb:256
+    def set_date_of_birth
+      @case.date_of_birth = @case.date_of_birth
+    end
```
mseedat-moj: This caveat bit of code is now showing up in multiple places. Perhaps time to put a fix into the actual gov_uk gem?

app/controllers/cases/notes_controller.rb
mseedat-moj: Have I understood correctly that `Notes` and `Case History` are literally the same thing but just have different outputs? If so can `messages_controller` not be repurposed? Or if that is undesirable a `module` called `Notes` or similar. `messages_controller` is already inside `/controllers/cases/` with appropriate routes

Polsonby: This is basically a copy of messages_controller and the way it works is the same, "Existing pattern is to put the create block of code into a dedicated Service" isn't the case for messages controller. They need to redirect differently and need to save with a different event, so I think having a separate route and controller is fine, actually, and better than adding complexity to messages controller

app/controllers/cases/notes_controller.rb
Polsonby: The problem with that is that the gem has already gone up a version and we can't use it because it's specific to the new gov uk design system. We could propose a patch to the earlier version, but I wouldn't know where to start with amending the behaviour inside the gem. It's more about the interaction with the json fields so I'm not sure it's a worthwhile use of time anyway

app/controllers/cases/notes_controller.rb
Polsonby: Oh, and although the two things, notes and messages, use the same json property name, I think they're different concepts and deserve their own controller. A key point being you can in theory use both alongside each other.

app/controllers/cases/notes_controller.rb
Polsonby: Moved the repeated method into a concern

app/controllers/cases/notes_controller.rb
mseedat-moj: Agree with them being different controllers. Weary we are pretending they are different through the actual domain mechanism as well. Just thinking about this from a testing perspective, as a result of the duplicated code the actual persistence mechanism is tested twice. And with two different implementations the risk of diverging inadvertently increases IMO. 

What I had in mind was something akin to the following for example:

```
+controllers/concerns/notable.rb+

module Notable
  def add(event_name:, on_success:)
    begin
      @case = Case::Base.find params[:case_id]
      @case.state_machine.send(event_name,
        acting_user: current_user,
        acting_team: case_team,
        message: params[:case][:message_text],
        controller: self.class
      )
    rescue ActiveRecord::RecordInvalid => err
      if err.record.errors.include?(:message)
        flash[:case_errors] = { message_text: err.record.errors[:message] }
      end
    ensure
      redirect_to on_success
    end    
  end

  private

  # Need to sort with manager first so that if we are both manager and something else, we don't
  # try to execute the action with our lower authority (which might fail)
  def case_team
    User.sort_teams_by_roles(case_teams).first
  end  

  def case_teams
    if current_user.teams_for_case(@case).any?
      current_user.teams_for_case(@case)
    else
      current_user.teams
    end
  end  
end

+cases/notes_controller.rb+

module Cases
  class NotesController < ApplicationController
    include Notable
    include GovUKDateFixes

    # Move this before_action into GovUKDateFixes ?
    before_action :set_date_of_birth

    def create
      add(
        event_name: 'add_note_to_case!',
        on_success:  case_path(@case, anchor: 'case-history')
      )
    end
  end
end

+cases/messages_controller.rb+
module Cases
  class MessagesController < ApplicationController
    include Notable

    def create
      add(
        event_name: 'add_message_to_case!',
        on_success:  case_path(@case, anchor: 'messages-section')
      )
    end
  end
end

+models/case_transition.rb+
# Addition of 'controller' property to allow debugging/retrieval
  jsonb_accessor :metadata,
                 message:                    :text,
                 filenames:                  [:string, array: true, default: []],
                 final_deadline:             :date,
                 linked_case_id:             :integer,
                 original_final_deadline:    :date,
                 controller:                 :string
```

mseedat-moj (approved):
Ta for the changes, great piece of work

PR #856, opened 2019-07-31
CT-2359 Hide date received legend but keep for accessibility only

digitalali-moj:
## Description
Hide text for legend of date recieved because it's already in heading, but keep the legend for accessibility.
 
## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots
NEW

image

OLD

image

 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2359
 
### Deployment
N/a
 
GO through offender sar journey then page 4 should have no legend text

### DIFF
--- a/app/views/cases/offender_sar/_date_received_step.html.slim
+++ b/app/views/cases/offender_sar/_date_received_step.html.slim
@@ -9,9 +9,11 @@
 = form_for @case, url: case_create_action(@case.object), as: :offender_sar do |f|
 
   .form-group
-    = f.gov_uk_date_field :received_date, { legend_text: '',
-      form_hint_text: t('cases.new.date_sar_received_copy'),
-      today_button: {class: ''}}
+    = f.gov_uk_date_field :received_date, {
+        legend_text: '',
+        form_hint_text: t('cases.new.date_sar_received_copy'),
+        today_button: true
+      }
   p
     = link_to "Previous", "/cases/new/offender/requested-info"
   input name="current_step" type="hidden" value=@case.current_step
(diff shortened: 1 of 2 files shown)


app/views/cases/offender_sar/_date_received_step.html.slim
```diff
@@ -9,9 +9,9 @@
 = form_for @case, url: case_create_action(@case.object), as: :offender_sar do |f|
 
   .form-group
-    = f.gov_uk_date_field :received_date, { legend_text: t('cases.new.date_sar_received'),
-        form_hint_text: t('cases.new.date_sar_received_copy'),
-        today_button: {class: ''}}
+    = f.gov_uk_date_field :received_date, { legend_text: '',
+      form_hint_text: t('cases.new.date_sar_received_copy'),
+      today_button: {class: ''}}
```
mseedat-moj: Formatting is a bit off. Try this:
```
= f.gov_uk_date_field :received_date, { 
    legend_text: '', 
    form_hint_text: t('cases.new.date_sar_received_copy'),
    today_button: { class: '' }
  }
```

If `class` is being set to `''` then is it still required?

mseedat-moj (changes requested):
Syntax and question

app/views/cases/offender_sar/_date_received_step.html.slim
digitalali-moj: Yes might as well refactor this while visiting. Done!

digitalali-moj: OK, so we need the class so it doesn't render it's own class (button) and change the look from link to button, even though it's a link.
Also need backslashes to allow the braces to be on separate lines

mseedat-moj (approved):
Thanks for the changes. The trailing slashes are unfortunate but better than having all the attributes on a single giant line 👍

PR #864, opened 2019-08-07
CT-2401 Always set subject to subject full name

digitalali-moj:
## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots
image
image

 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2401
 
### Deployment
n/a
 
Create a new Offender SAR case then check the title on the case page and the request in cases table. Should both be the subject name.

### DIFF
--- a/app/models/case/sar/offender.rb
+++ b/app/models/case/sar/offender.rb
@@ -73,6 +73,12 @@ class Case::SAR::Offender < Case::Base
   validate :validate_date_of_birth
   validate :validate_received_date
 
+  before_save :set_subject
+
+  def set_subject
+    self.subject = subject_full_name
+  end
+
   def validate_received_date
     super
   end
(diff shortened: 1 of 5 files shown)


app/models/case/sar/offender.rb
```diff
@@ -73,6 +73,12 @@ def searchable_fields_and_ranks
   validate :validate_date_of_birth
   validate :validate_received_date
 
+  before_save :set_subject
```
mseedat-moj: This will be triggered fro case create/update - is that intended or just for case creation?

mseedat-moj (approved):
Looks good - one question from me regarding the set_subject hook

app/models/case/sar/offender.rb
Polsonby: Intended to do it for update as well as create - so when someone edits the user's name, the subject changes too

PR #865, opened 2019-08-07
[02] CT-2192 Introduce Warehouse::CaseReport (PR No. 2)

mseedat-moj:
## Description
Introduces the `warehouse` module which stores denormalised `Case::Base` data into the `warehouse_case_reports` table. This PR does not include the on-the-fly sync operations, report generation - only the warehouse generation/reconciliation form this PR.

### Notes
The method `CaseReport#process_cases` requires attention as it performs a manual `GC.start` and `sleep(10) # 10 second thread sleep`. This is to allow the Ruby Garbage Collector to have enough time to free memory allocations, and to ensure any CPU burst is minimised by processing small batches. This is important on shared-resouce systems such as Cloud Platform/AWS. The `10 second` setting is completely arbitrary and during basic testing on AWS *small* EC2 instance with 10,000 records, did not cause any issues.

On the application server it is envisaged on-the-fly sync operations will mean this method will not be called. Batch processing of `./cts cases warehouse` is recommended for execution in the `background-jobs` container to prevent resource drain from user-facing servers.

## Self-review checklist

* [X] (3) Tests passing
 
##### Warehouse_Case_Reports table structure
Screen Shot 2019-08-07 at 19 38 43

##### Warehouse_Case_Reports denormalised data
Screen Shot 2019-08-07 at 19 40 49

##### Warehousing command line process
Screen Shot 2019-08-07 at 19 39 58

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2192
 
### Deployment
Note migration for `warehouse_case_reports` has foreign key constraint against `cases` table.
 
From the command line in the rails app root `./cts cases warehouse`

### DIFF
--- a/app/models/warehouse/case_report.rb
+++ b/app/models/warehouse/case_report.rb
@@ -6,15 +6,16 @@ module Warehouse
 
     CASE_BATCH_SIZE = 500
 
-    # Usage of class methods is to allow this class to be used in async jobs
+    # Class methods to allow this class to be used in async jobs
     class << self
       def for(kase)
         kase.warehouse_case_report || self.new(case_id: kase.id)
       end
 
       # Every field is deliberately set explicitly, please do not use any
-      # clever magic etc as each field has a history of how it should
+      # clever meta-magic as each field has a history of how it should
       # be calculated and therefore needs to be readily understood
+      #
       #rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength
       def generate(kase)
         case_report = self.for(kase)
--- a/app/models/warehouse/case_report.rb
+++ b/app/models/warehouse/case_report.rb
@@ -57,7 +57,7 @@ module Warehouse
         case_report.sar_subject_type = kase.respond_to?(:subject_type) ? kase.subject_type.humanize : nil
         case_report.sar_subject_full_name = kase.respond_to?(:subject_full_name) ? kase.subject_full_name : nil
         case_report.business_unit_responsible_for_late_response = kase.decorate.late_team_name
-        case_report.extended = extension_count(kase) > 0 ? 'Yes' : 'No'
+        case_report.extended = self.extension_count(kase) > 0 ? 'Yes' : 'No'
         case_report.extension_count = self.extension_count(kase)
         case_report.deletion_reason = kase.reason_for_deletion
         case_report.casework_officer = kase.casework_officer
--- a/app/models/warehouse/case_report.rb
+++ b/app/models/warehouse/case_report.rb
@@ -12,9 +12,10 @@ module Warehouse
         kase.warehouse_case_report || self.new(case_id: kase.id)
       end
 
-      #rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength
       # Every field is deliberately set explicitly, please do not use any
-      # clever magic etc
+      # clever magic etc as each field has a history of how it should
+      # be calculated and therefore needs to be readily understood
+      #rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength
       def generate(kase)
         case_report = self.for(kase)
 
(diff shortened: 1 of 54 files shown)


mseedat-moj:
@Polsonby this PR is randomly failing due to unrelated code specs - the usual report type intermittent failures and on the last run 800+ views. None of this makes sense to me. Suggest review anyway and decision to force merge.

app/models/warehouse/case_report.rb
```diff
@@ -0,0 +1,161 @@
...
+
+    # Usage of class methods is to allow this class to be used in async jobs
+    class << self
+      def for(kase)
+        kase.warehouse_case_report || self.new(case_id: kase.id)
+      end
+
+      #rubocop:disable Metrics/CyclomaticComplexity, Metrics/MethodLength
+      # Every field is deliberately set explicitly, please do not use any
+      # clever magic etc
```
Polsonby: I love this comment, but it would be helpful to additionally explain why not :)

app/models/warehouse/case_report.rb
```diff
@@ -0,0 +1,161 @@
...
+        case_report.postal_address = kase.postal_address
+        case_report.email = kase.email
+        case_report.appeal_outcome = kase.appeal_outcome&.name
+        case_report.third_party = kase.respond_to?(:third_party) ? self.humanize_boolean(kase.third_party) : nil
+        case_report.reply_method = kase.respond_to?(:reply_method) ? kase.reply_method.humanize : nil
+        case_report.sar_subject_type = kase.respond_to?(:subject_type) ? kase.subject_type.humanize : nil
+        case_report.sar_subject_full_name = kase.respond_to?(:subject_full_name) ? kase.subject_full_name : nil
+        case_report.business_unit_responsible_for_late_response = kase.decorate.late_team_name
+        case_report.extended = extension_count(kase) > 0 ? 'Yes' : 'No'
+        case_report.extension_count = self.extension_count(kase)
```
Polsonby: This extension_count probably doesn't need the self, if the one above doesn't?

app/models/warehouse/case_report.rb
```diff
@@ -0,0 +1,161 @@
...
+        case_report.deletion_reason = kase.reason_for_deletion
+        case_report.casework_officer = kase.casework_officer
+        case_report.created_by = kase.creator.full_name
```
Polsonby: Can kase.creator ever be nil?

db/migrate/20190609165906_warehouse_case_report.rb
```diff
@@ -0,0 +1,74 @@
+# Denormalised data-warehousing of all case data required to generate
+# Open/Closed reports. There is a one-to-one relationship between
+# this warehouse table and the cases table.
+class WarehouseCaseReport < ActiveRecord::Migration[5.0]
+
+  # Copied from CSVExporter::CSV_COLUMN_HEADINGS (July 2019)
```
Polsonby: Worth adding a comment to CSVExporter that if we change them there, they need to be added to a migration to update them in the schema?

Polsonby (approved):
I think we need to fix the duplicate keys issue before merging this to master - I've seen it completely consistently both running locally in parallel and on CI today, and I think we'll end up with it in master too. But that aside, great work. Just a few comments/questions but happy to approve other than the problem with the specs

app/models/warehouse/case_report.rb
mseedat-moj: Made consistent by adding the missing `self.` to explicitly remind the user it's a class method

app/models/warehouse/case_report.rb
mseedat-moj: In theory no because all records prior to March 2019 have the 'default user' assigned as the creator (id: -100) with the name as an empty string.

Polsonby:
Looking good 👍  - you going to rebase this after PR 1 is merged?

PR #867, opened 2019-08-09
[01] CT-2408 Intermittent ReportType spec fails (PR No. 1)

mseedat-moj:
## Description
These changes are the latest attempt to rectify by ensuring the report_types table is in a consistent state (as well as the other tables) and therefore reduce the potential for test errors when using parallel tests.
 
## Self-review checklist

* [X] (3) Tests passing
  
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2408

### Notes
ReportType has the standard id as a primary key. It also has an abbr string field which is not constrained in the codebase. However, a database index exists which prevents abbr duplicates from being inserted and therefore makes abbr a secondary primary key.

This causes immense problems with intermittent build failures when using parallel_tests (i.e. on SemaphoreCI).

Run `rails:parallel:spec` locally and note results

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -1,4 +1,4 @@
-if ENV['COVERAGE'].present?
+if !!ENV['COVERAGE']
   require 'simplecov'
   SimpleCov.start 'rails' do
     add_group "Services", "app/services"
--- a/spec/services/stats/base_report_spec.rb
+++ b/spec/services/stats/base_report_spec.rb
@@ -9,9 +9,8 @@ module Stats
   end
 
   describe BaseReport do
-
-    before(:all)  { create :report_type }
-    after(:all)   { ReportType.delete_all }
+    before(:all) { create_report_type(abbr: :r006) }
+    after(:all) { DbHousekeeping.clean(seed: true) }
 
     describe '.new' do
       it 'raises if BaseReport is instantiated' do
--- a/spec/services/stats/r005_monthly_performance_report_spec.rb
+++ b/spec/services/stats/r005_monthly_performance_report_spec.rb
@@ -3,8 +3,8 @@ require 'rails_helper'
 module Stats
   describe R005MonthlyPerformanceReport do
 
-    before(:all) { create :report_type, :r005 }
-    after(:all) { ReportType.delete_all }
+    before(:all) { create_report_type(abbr: :r005) }
+    after(:all) { DbHousekeeping.clean(seed: true) }
 
     describe '.title' do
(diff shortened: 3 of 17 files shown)


spec/rails_helper.rb
```diff
@@ -1,4 +1,4 @@
-if ENV['COVERAGE'].present?
+if !!ENV['COVERAGE']
```
mseedat-moj: For reasons that remain a mystery, a new issue was the Rails framework wasn't loaded at this point, and therefore the `.present?` method did not exist, meaning every test failed!

spec/services/stats/base_report_spec.rb
```diff
@@ -9,9 +9,8 @@ def report_type
   end
 
   describe BaseReport do
-
-    before(:all)  { create :report_type }
-    after(:all)   { ReportType.delete_all }
+    before(:all) { create_report_type(abbr: :r006) }
+    after(:all) { DbHousekeeping.clean(seed: true) }
```
mseedat-moj: I decided to explicitly set the `seed` parameter as it is quite an important distinction (default true, can be set to false) and as a result was giving some false test outputs (for which tests had been adjusted!)

spec/support/db_housekeeping.rb
```diff
@@ -18,6 +18,7 @@ def self.clean(seed: true)
       cases_users_transitions_trackers
       linked_cases
       search_queries
+      report_types
```
mseedat-moj: Notice how report_types is now being cleared, because the user is expected to use `create_report_type` rather than the factory method directly (`create :report_type`)

spec/services/stats/r005_monthly_performance_report_spec.rb
```diff
@@ -65,7 +65,7 @@ module Stats
           [[1, :red], [7, :red], [13, :red]],   # March
           [[1, :red], [7, :green], [13, :red]], # April
           [],
-          [[1, :red], [7, :red], [13, :red]],   # Totals
+          [[1, :red], [7, :amber], [13, :red]],   # Totals
```
Polsonby: Surprised this changed from `:red` to `:amber` - is that intentional?

Polsonby (approved):
Great work, Mo :) 

Just one question

spec/services/stats/r005_monthly_performance_report_spec.rb
mseedat-moj: Yes I think the tests previously assumed there was leftover data in the dB which contributed to the rag rating

PR #868, opened 2019-08-09
[03] CT 2192 Add sync operations (PR No. 3)

mseedat-moj:
## Description
Implements the synchronisation/reconciliation mechanism for CaseReport constituent data changes. 

## Self-review checklist

* [x] (3) Tests passing
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2192
 
### Deployment
N/A
 
1. Edit any Case, User, Team and note `warehouse_case_reports` is consistent
2. Assigning a User should ensure the CaseReport is consistent
3. Closing a case shouold ensure `CaseReport` is consistent

### Notes
The implementation is fairly basic using the `after_commit` hook on `ApplicationRecord` to check whether the Model is:

1. One of the Models that contributes data to the 'Cases Report'
2. If so then the `sync` operation is queued as a Job. This is to offload the work to a separate thread/container and prevent the web application from hanging, and to reduce resource usage on the main server.
3. The sync operation may update multiple `CaseReport` entries depending on the type of data that is being updated e.g. a User's Name changing could affect multiple CaseReport entries

### DIFF
--- a/.rspec_parallel
+++ b/.rspec_parallel
@@ -1,0 +1,1 @@
+--format progress
--- a/Gemfile
+++ b/Gemfile
@@ -93,7 +93,6 @@ group :development, :test do
   gem 'pry-byebug'
   gem 'rspec-collection_matchers'
   gem 'rspec-rails', '~> 3.8'
-  gem 'rspec-retry', '~> 0.6.1'
   gem 'rubocop', '~> 0.71.0', require: false
   gem 'rubocop-rspec', '~> 1.33.0', require: false
   gem 'rubocop-performance', require: false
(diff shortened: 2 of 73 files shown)


Polsonby (approved):
Wow... :) 

Very impressive piece of work. 

So if I understand you right, every time any model writes changes to the database, ApplicationRecord is going to check for any syncable columns and if it finds them, schedule a sync task later? 

The only thing I wondered was whether you could remove those checks and simplify the syncable mechanism by adding this as a model concern to specific models that need to sync? It feels a little complex, and a simpler, more direct implementation to trigger syncing might help. 

But you're the one who's thought the problem through in detail and this is just a drive-by comment after the fact, so happy to go with your judgement either way - great work :)

PR #874, opened 2019-08-12
CT-2396 Workaround for changing wording, maybe can do on en.yml some how

digitalali-moj:
## Self-review checklist

* [ ] (3) Tests passing


Polsonby (changes requested):
It's important we don't change the `type_abbreviation` for a case class - aside from anything else, this breaks any existing cases in the database and will have other side effects too :) 

Let's find another way to achieve this goal

digitalali-moj:
Used another PR as this one is totally useless based on your comment :)

PR #877, opened 2019-08-16
CT-2421 Remove nil values when syncing to prevent error

mseedat-moj:
## Description
When a warehousable record is updated which does not effect any Warehouse::CaseReport a nil error is generated rather than skipping any processing. This PR fixes that error.
 
## Self-review checklist

* [X] (3) Tests passing
 
### Screenshots
N/A
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2421

### Sentry error
https://sentry.service.dsd.io/mojds/correspondence-tool-staff/issues/37161/

### Deployment
N/A
 
N/A

### DIFF
--- a/spec/services/stats/warehouse/case_report_sync_spec.rb
+++ b/spec/services/stats/warehouse/case_report_sync_spec.rb
@@ -12,6 +12,17 @@ describe Stats::Warehouse::CaseReportSync do
   end
 
   context '.affected_cases' do
+    it 'returns a list of cases affected by the record' do
+      record = Object.new
+      kase = create :foi_case
+
+      setting = {
+        fields: ['fake', 'useless'],
+        execute: ->(_record, _query) { [kase] }
+      }
+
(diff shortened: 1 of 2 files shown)


Polsonby (approved):
Looks good! 

Just one question for my education :)

spec/services/stats/warehouse/case_report_sync_spec.rb
```diff
@@ -191,7 +207,7 @@
       let(:new_team_name) { Faker::Company.name }
 
       before do
-        Warehouse::CaseReport.generate(responded_case)
+        ::Warehouse::CaseReport.generate(responded_case)
```
Polsonby: Just wondering, what's the purpose of this `::` here at the start of the line? I don't think I've seen that before

mseedat-moj: Ensures module namespace is traversed from root rather than the current class/module. I found semaphore builds kept failing without the :: from root whereas locally it didn't

PR #879, opened 2019-08-18
CT-2258 Offender SARs Record Data Requests

mseedat-moj:
## Description
Allow an Offender SAR to have data requests recorded.
 
## Self-review checklist

* [x] (3) Tests passing
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2258
 
1. Log in as a Branston User e.g. Brian Rix
2. Create a new Offender SAR 
3. Note 'Record data request' green button in action bar
4. Proceed to add data requests
5. Note case now is in a 'waiting for data' state, and 'Record data request' button is grey colour


app/models/case/sar/offender.rb
```diff
@@ -104,4 +112,24 @@ def current_team_and_user_resolver
...
+      self.send("#{field}=", self.read_attribute(field))
+    end
+  end
+
+  # User can add data requests at any point in the workflow, but
+  # the case state should not change
+  def allow_waiting_for_data_state?
+    self.current_state == 'data_to_be_requested'
+  end
 end
```
mseedat-moj: @Polsonby This is the workaround I mentioned regarding preventing GovUKDateFields from throwing the 'no method error valid? on nil class'. I believe it occurs when creating an object using the ActiveRecord relationship. In my case doing `offender_sar.data_requests.create!(...)` (IN A TEST!!) resulted in the `valid? on nil:class` error. This is my proposed work around. I have lost too many hours to this issue.

app/models/case/sar/offender.rb
```diff
@@ -41,11 +49,7 @@ def searchable_fields_and_ranks
     send_by_email: 'send_by_email',
   }
 
-  acts_as_gov_uk_date :date_of_birth,
-                      :date_responded,
-                      :date_draft_compliant,
-                      :external_deadline,
-                      :received_date
+  acts_as_gov_uk_date(*GOV_UK_DATE_FIELDS)
```
Polsonby: Is this easier to read? Seems a little indirect, but not such that I'd complain about it. Ah - just read further down - we use the list again 👍 

But maybe we could have the `acts_as_gov_uk_date(*GOV_UK_DATE_FIELDS)` right after the frozen array so they're next to each other?

app/models/case/sar/offender.rb
Polsonby: Same here! Looks cool - nice job, I'm sure it will work better than my previous "fix"

app/policies/case/sar/offender_policy.rb
```diff
@@ -7,4 +7,8 @@ def transition?
   def can_add_note_to_case?
     true
   end
+
+  def can_record_data_request?
+    self.case.offender_sar?
+  end
```
Polsonby: Are there ever any cases for `case/sar/offender_policy` that `self.case` could be something other than an offender_sar? Just wondering why not say "true" here

spec/controllers/cases/data_requests_controller_spec.rb
```diff
@@ -0,0 +1,122 @@
+require 'rails_helper'
+
+RSpec.describe Cases::DataRequestsController, type: :controller do
+  let(:manager) { find_or_create :disclosure_bmt_user }
```
Polsonby: This looks a wee bit odd - should the user not be a Branston user?

spec/services/data_request_service_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+describe DataRequestService do
+  let(:user) { create :user }
+  let(:offender_sar_case) { create :offender_sar_case }
+  let(:data_request_attributes) {
+    {
+      '0' => { location: 'The Clinic', data: 'Lots of paper please' },
+      '1' => { location: 'The House', data: 'More paper please' },
+      '2' => { location: 'A Prison', data: 'Less paper please' },
+      '3' => { location: 'The LA', data: 'All your paper please' }
+    }
```
Polsonby: :)

spec/services/data_request_service_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+
+      result = service.process?(
+        location: nil,
+        data: 'This is some data  '
+      )
+
+      expect(result).to be true
+    end
+
+    it 'returns false when both location and data are'
```
Polsonby: Something missing here?

Polsonby (commented):
Great work, just a few questions

app/policies/case/sar/offender_policy.rb
mseedat-moj: errrr..... good point!  I think I did it to explicitly tell the reader you can only record a data request if the current case is an Offender SAR but I suppose that's a given!

spec/services/data_request_service_spec.rb
mseedat-moj: Rectified in #895

app/models/case/sar/offender.rb
mseedat-moj: I haven't verified that this fix works for all the existing places where the annoying error crops up

app/views/cases/data_requests/_data_request.html.slim
```diff
@@ -0,0 +1,12 @@
+.data-request__single.js-data-request__form data-index="#{form.index}"
+  h3.page-heading--secondary.data-request__position
```
Polsonby: Should be h2 for accessibility, I believe - screen reader users expect the headings to be used in sequential order and not miss out levels. But super tiny issue

app/views/cases/offender_sar/_data_requests.html.slim
```diff
@@ -0,0 +1,28 @@
...
+            th Date received
+        tbody
+          - case_details.data_requests.order(date_requested: :asc).each do |data_request|
+            tr
+              td
+                = data_request.location
+              td
+                = truncate(data_request.data, length: 140)
+              td
+                = data_request.date_requested.strftime('%F')
```
Polsonby: I believe this date should be formatted like other dates on the page

```suggestion
                = l data_request.date_requested, format: :default
```

app/helpers/cases_helper.rb
```diff
@@ -116,6 +116,12 @@ def action_button_for(event)
               case_sar_extensions_path(@case),
               id: 'action--remove-extended-deadline-for-sar',
               class: 'button-secondary', method: :delete
+    when :record_data_request
+      btn_type = @case.current_state == 'data_to_be_requested' ? 'secondary' : 'tertiary'
+      link_to 'Record data request',
+              new_case_data_request_path(@case),
+              id: 'action--record-data-request',
+              class: "button-#{btn_type}", method: :get
```
Polsonby: This still POSTs to the next page - really annoying! But not worth spending time on IMO

Polsonby (approved):
Great work, Mo 👍 

The headings on the new data request form should be h2 not h3, and the date format on the list of  should be consistent with other dates. Approving anyway but if you're pushing again, perhaps you could address these two tiny points

app/helpers/cases_helper.rb
mseedat-moj: I will check this before merging to master as I have it definitely being s GET to the next page!

app/views/cases/data_requests/_data_request.html.slim
mseedat-moj: This is a partial included on the `cases/show` page which already has `<h2>` elements. Therefore using `<h3>` here.

app/helpers/cases_helper.rb
mseedat-moj: Can you check this again - I cannot reproduce:
Screen Shot 2019-08-30 at 10 23 38

Polsonby: image
It's consistent in Chrome and Safari - click the "Record data request" button and hit reload in the browser (without the web inspector open) and it asks you to resubmit the form - header shows POST not GET. 

But as I said, the code in the view looks fine to me and IMO it's not worth spending time on investigating further just now - just annoying. Presumably there is some UJS overriding the data-method attribute somewhere

app/views/cases/data_requests/_data_request.html.slim
Polsonby: That's the other partial - this is the form fields in the new data requests form. When you go to `http://localhost:3000/cases/9/data_requests/new`

app/helpers/cases_helper.rb
mseedat-moj: Fixed - it's because in the helper I copy pasted a previous link/button, and it has `method: :get` set. This causes some interesting behaviour!

```
    when :record_data_request
      btn_type = @case.current_state == 'data_to_be_requested' ? 'secondary' : 'tertiary'
      link_to 'Record data request',
              new_case_data_request_path(@case),
              id: 'action--record-data-request',
              class: "button-#{btn_type}", method: :get
```

Removing the `method: :get` solves the problem!!

mseedat-moj:
Superceded by Parent PR:

https://github.com/ministryofjustice/correspondence_tool_staff/pull/900

PR #884, opened 2019-08-20
CT-2396 Show type with hyphen and update case type in creation

digitalali-moj:
## Description
Updated the decorator to add _ in place of _ for cases page view.
Also changed case creation journey view from Create SAR to
Create Offender SAR using locale file. And added correct case
name in seeder for offender SAR cases. 
Also add migration to change name on production.

## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots
(Old screenshots already on ticket)
New:
image
image
image
image

 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2396
 
### Deployment
Will require migration for production
 
Go through offender SAR case creation journey and check titles. Check grey sub heading on case view page. Check type column on cases view page.

### DIFF
--- a/app/decorators/case/base_decorator.rb
+++ b/app/decorators/case/base_decorator.rb
@@ -180,7 +180,7 @@ class Case::BaseDecorator < Draper::Decorator
   end
 
   def pretty_type
-    object.class.type_abbreviation
+    object.class.type_abbreviation.tr('_', '-')
   end
 
   def late_team_name
--- a/app/decorators/case/base_decorator.rb
+++ b/app/decorators/case/base_decorator.rb
@@ -180,7 +180,7 @@ class Case::BaseDecorator < Draper::Decorator
   end
 
   def pretty_type
-    object.class.type_abbreviation.tr('_', '-')
+    object.class.type_abbreviation.dasherize
   end
 
   def late_team_name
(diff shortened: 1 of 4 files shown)


db/seeders/correspondence_type_seeder.rb
```diff
@@ -37,7 +37,7 @@ def seed!
 
     rec = CorrespondenceType.find_by(abbreviation: 'OFFENDER_SAR')
     rec = CorrespondenceType.new if rec.nil?
-    rec.update!(name: 'Offender SAR (OFFENDER)',
+    rec.update!(name: 'Offender subject access request',
```
mseedat-moj: Should this be `Offender Subject Access Request` ?

digitalali-moj: Just copied what's on the ticket. I guess we can check this with Alex though as you might be right here.

digitalali-moj: I confirmed with Alex who suggested to keep as is since it's consistent with other case types: 
image

app/decorators/case/base_decorator.rb
```diff
@@ -180,7 +180,7 @@ def admin_internal_deadline
   end
 
   def pretty_type
-    object.class.type_abbreviation
+    object.class.type_abbreviation.tr('_', '-')
   end
```
mseedat-moj: `object.class.type_abbreviation.dasherize` is what you are looking for I think!

https://apidock.com/rails/String/dasherize

mseedat-moj (changes requested):
Slight change to abbreviation output

app/decorators/case/base_decorator.rb
digitalali-moj: This worked, done

mseedat-moj (approved):

PR #886, opened 2019-08-20
CT-2428 Rename offender sar method

mseedat-moj:
## Description
Makes `case#offender_sar?` consistent with other similar methods.

## Self-review checklist

* [X] (3) Tests passing
 
### Screenshots
N/A
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2428
 
### Deployment
N/A
 
N/A

### DIFF
--- a/app/models/case/base.rb
+++ b/app/models/case/base.rb
@@ -769,7 +769,7 @@ class Case::Base < ApplicationRecord
   def overturned_ico?;      false;  end
   def overturned_ico_sar?;  false;  end
   def overturned_ico_foi?;  false;  end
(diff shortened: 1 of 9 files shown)


Polsonby (commented):
@mseedat-moj The reason for naming it this way was that I used `offender_sar?` to start with and it causes terrible problems - that method is overriden or generated somewhere else and I couldn't find where, so ended up going for the `is_` flavoured one. If it now works for some reason, great, otherwise I suggest we leave it as I had it so we don't need to go down that rabbit hole

Polsonby (approved):
Great detective work :) 👍

PR #908, opened 2019-09-06
CT-2318 Redirect if case is deleted

Polsonby:
## Description
One of our most frequently recurring Sentry issues - it transpired that the issue happens when someone clicks a link to edit the case assignment from an email but on a case that has been subsequently deleted. The exception is caused most directly because we were checking for the existence of the assignment directly via Assignment, but then trying to reach it through the case which is nil at that point. 

The wider problem is that the system was attempting to display the edit assignment page in the first place on a deleted case, so I added a before action to redirect to the home page with an appropriate message, and a feature spec to reproduce the bug and validate the new behaviour. 
 
## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots

image

 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2318
 
### Deployment
none
 
none

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -17,6 +17,8 @@ class AssignmentsController < ApplicationController
     :unaccept,
   ]
 
+  before_action :check_case_not_deleted, only: [:edit]
+
   before_action :validate_response, only: :accept_or_reject
 
   def new
@@ -249,7 +251,7 @@ class AssignmentsController < ApplicationController
     if Case::Base.exists?(id: params[:case_id])
       set_case
     end
-    if Assignment.exists?(id: params[:id])
+    if @case && Assignment.exists?(id: params[:id])
       @assignment = @case.assignments.find(params[:id])
     end
   end
@@ -260,6 +262,13 @@ class AssignmentsController < ApplicationController
               .find(params[:case_id])
               .decorate
     @case_transitions = @case.transitions.decorate
+  end
+
+  def check_case_not_deleted
+    if @case.nil? && Case::Base.unscoped.find(params[:case_id])
+      flash[:notice] = 'Case has been deleted.'
+      redirect_to open_filter_path
+    end
   end
 
   def set_team_users
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -17,8 +17,6 @@ class AssignmentsController < ApplicationController
     :unaccept,
   ]
 
-  before_action :check_case_not_deleted, only: [:edit]
-
   before_action :validate_response, only: :accept_or_reject
 
   def new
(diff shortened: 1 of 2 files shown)


app/controllers/assignments_controller.rb
```diff
@@ -17,6 +17,8 @@ class AssignmentsController < ApplicationController
     :unaccept,
   ]
 
+  before_action :check_case_not_deleted, only: [:edit]
```
Polsonby: I'm wondering about removing the `, only: [:edit]` here because there are other actions you shouldn't be able to do on deleted cases too - WDYT @mseedat-moj ?

app/controllers/assignments_controller.rb
```diff
@@ -262,6 +264,13 @@ def set_case
     @case_transitions = @case.transitions.decorate
   end
 
+  def check_case_not_deleted
+    if @case.nil? && Case::Base.unscoped.find(params[:case_id])
+      flash[:notice] = 'Case has been deleted.'
+      redirect_to open_filter_path
+    end
+  end
+
```
mseedat-moj: Recommend something like this:

```
  def redirect_on_nil_case!
    return if Case::Base.exists?(params[:case_id])
    redirect_to open_filter_path, alert: 'Case has been deleted.'
  end
```

spec/features/cases/foi/case_assignment_responding_spec.rb
```diff
@@ -123,4 +123,14 @@
...
 
+  scenario 'kilo clicks on a link to an assignment that has been deleted' do
+    assignment_id = assignment.id
+    assignment.accept responder
+    assigned_case.update_attribute(:deleted, true)
+
+    visit edit_case_assignment_path assigned_case, assignment_id
+
+    expect(page).to have_current_path(open_filter_path)
+    expect(page).to have_content("Case has been deleted.")
```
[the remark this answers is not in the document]
mseedat-moj: See patch diff - this test is not technically correct because the soft-delete gem will treat `Case::Base.find(id)` as a 'not found exception'. The test above works because `assigned_case` is still in memory/cache during the test and therefore the corresponding code in the controller is required to do the `Case::Base.unscoped.find(id)` rather than just `Case::Base.exists?(id)` or `Case::Base.find_by(id: id)`

See patch file for what the test/corresponding method look like

mseedat-moj (changes requested):
See git patch file for changes to the test and some ideas for changing `assignments_controller`

https://drive.google.com/open?id=1m_S-j2n1zcIFGp78bkj2h6Ld0zNo2vNU

spec/features/cases/foi/case_assignment_responding_spec.rb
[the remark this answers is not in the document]
Polsonby: Pinged you on Slack :)

app/controllers/assignments_controller.rb
Polsonby: As discussed moved the check into `set_case_and_assignment` so that it gets applied to all these actions :)

app/controllers/assignments_controller.rb
Polsonby: As discussed, I'd like to keep the logic as-is so that it specifically relates to deleted cases. Renamed method more or less as you propose - now `redirect_on_deleted_case!`

spec/features/cases/foi/case_assignment_responding_spec.rb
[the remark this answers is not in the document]
Polsonby: As discussed, IMO the test was correct but I've changed it to your suggested approach which also works fine here 👍

mseedat-moj (approved):
Excellent sleuthing

PR #917, opened 2019-09-10
CT-2482 Add better tests for closing case; fix error when closing

Polsonby:
## Description
I'd omitted to test the case closing properly; this PR addresses that, fixes the bug when closing a case, adds more validation to the closing states and checks to see the outcome is reflected on the case page
 
## Self-review checklist

* [x] (3) Tests passing
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2482
 
Create a new offender SAR case, then move it through the states using the transition button until you get to close case. Check you can't submit an empty response, then fill in the date responded and continue. Again, check you can't submit an empty response and choose an outcome. Click to submit and ensure the values are shown on the case page and the case is closed successfully


app/views/cases/offender_sar/_closure_outcomes_form.html.slim
```diff
@@ -7,9 +7,6 @@
   .js-info-held-status
     = f.radio_button_fieldset :info_held_status_abbreviation, choices: CaseClosure::InfoHeldStatus.all, text_method: :name, value_method: :abbreviation
 
-  .js-outcome-group
-    = f.radio_button_fieldset :outcome_abbreviation, choices: CaseClosure::Outcome.active, text_method: :name, value_method: :abbreviation
-
```
Polsonby: For things like FOI requests, there are partial responses and loads of criteria that need to be completed on closing a case. Offender SAR don't need these so removed from the view

danielatdigital (approved):

app/views/cases/offender_sar/_closure_outcomes_form.html.slim
```diff
@@ -10,5 +10,6 @@
   .grid-row
     .column-two-thirds
       .button-holder
+        = hidden_field :offender_sar, value: "Test"
```
Polsonby: If the form consists solely of radio buttons as is the case here, submitting the form with nothing selected leaves the params[:offender_sar] completely unset and it blows up on the `.require(:offender_sar)` in the offender sar params method. This hidden field is there to prevent this error and has no other side effects. 

I'm thinking there should be a more canonical way to do this but I've gone blank on what it is - better ideas welcomed!

Polsonby:
Reviews so far would be welcome - there's more to do after talking to Alex just now, preventing the outcome buttons from displaying after all and adding a placeholder message to users pending the ticket to add expanded and custom Branston reasons for lateness if the response is late

app/views/cases/offender_sar/_closure_outcomes_form.html.slim
mseedat-moj: This is too dirty! Looking at `offender_sar/_closure_outcomes_form.html.slim`, is the `form_for` output not correct?

```
= form_for kase, as: :"#{@correspondence_type_key}", url: url do |f|
    ...
```

Where `@correspondence_type_key` should output `offender_sar`

mseedat-moj (changes requested):
Generally looks good but kludge using hidden field requires re-thinking

app/views/cases/offender_sar/_closure_outcomes_form.html.slim
```diff
@@ -4,14 +4,15 @@
   - if @case.responded_late?
     = render partial: 'cases/shared/team_caused_late_response', locals: { kase: @case, f: f }
 
-  .js-info-held-status
-    = f.radio_button_fieldset :info_held_status_abbreviation, choices: CaseClosure::InfoHeldStatus.all, text_method: :name, value_method: :abbreviation
+  p Coming soon - reasons for late case
```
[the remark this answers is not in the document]
Polsonby: I don't think so - this is just a placeholder for an imminent piece of work, this won't appear in the app proper. We've been a bit hit and miss about using the translations file recently, maybe we should indeed be stricter elsewhere

app/models/case/sar/offender.rb
```diff
@@ -113,6 +114,10 @@ def offender_sar?
     true
   end
 
+  def responding_team
+    managing_team
+  end
```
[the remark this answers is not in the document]
Polsonby: Good point, done

mseedat-moj (approved):
Good work

PR #922, opened 2019-09-20
Ct 2423 generate acknowledgement letter text

Polsonby:
## Description
Letter templates and letter generation - new model for letter templates. In the end decided to use ERB since it's built in to Ruby hence no new dependencies in the app. 

The letter_template takes a values object, which can either be a case or specially constructed - we'll be adding a Letter class which wraps a case in order to add features subsequently such as knowing when a previous letter was sent and referring to it in a template, but for now it takes the case
 
## Self-review checklist

* [ ] (3) Tests passing


app/models/letter_template.rb
```diff
@@ -0,0 +1,13 @@
+class LetterTemplate < ApplicationRecord
+  validates :name, :body, :template_type, presence: true
+
+  enum template_type: {
+    dispatch: 'dispatch',
```
mseedat-moj: `despatch` !!

config/routes.rb
```diff
@@ -157,6 +157,11 @@
     end
 
     resources :data_requests
+
+    resource :letters, only: [], path: "letters/:type"  do
+        get "/new", to: 'letters#new'
+        get "/render", to: 'letters#render_letter', as: 'render'
```
mseedat-moj: Should `render_letter` be `show` ?

app/controllers/cases/letters_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  def render_letter
+    @type = params[:type]
+    unless params[:letter_form] and params[:letter_form][:letter_template_id]
+      flash[:notice] = 'Please select a template.'
+      redirect_to new_case_letters_path(@case.id, @type) and return
+    end
+    letter_template_id = params[:letter_form][:letter_template_id]
+    @letter = Letter.new(letter_template_id)
+    @letter_template = LetterTemplate.find(letter_template_id)
+  end
```
mseedat-moj: Would prefer all this logic to either be in the `letter` model or in an appropriate service.

Don't use the `and` operator, use `&&` instead, unless you really do mean `and` ?

Should use strong parameters for consistency with codebase

app/models/letter_template.rb
Polsonby: Nope :) - they're both valid UK English but Alex and the team chose to settle on `dispatch` e.g.       `ready_to_dispatch`. We still use `de-` in places accidentally, but it's correct here

image

config/routes.rb
Polsonby: Excellent wheeze, never occurred to me. I knew it wasn't create...

app/controllers/cases/letters_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def new
+    @type = type_name(params[:type])
+    @letter = Letter.new(params[:letter_template_id])
+    @letter_templates = LetterTemplate.where(template_type: @type)
+  end
+
+  def render_letter
+    @type = params[:type]
+    unless params[:letter_form] and params[:letter_form][:letter_template_id]
```
Polsonby: Yes, this `and` is used for flow control and short-circuit exits early, never evaluating the right hand side if the left is false - here it prevents the second part from blowing up if params[:letter_form] is nil

app/controllers/cases/letters_controller.rb
Polsonby: > Would prefer all this logic to either be in the letter model or in an appropriate service.
Will take a look at that, thanks...

> Should use strong parameters for consistency with codebase
This is what seemed strange to me - I went to add params but there are actually lots of places we just access the params directly. So this is IMO consistent with the codebase, which uses both approaches. I've ensured the params are not reflected back to the user in the page, and they're all used safely IMO. But I'll have another look at it, as I agree it's best-practice.

app/models/letter_template.rb
mseedat-moj: Booo. 51st state here we come

app/controllers/cases/letters_controller.rb
mseedat-moj: The right hand side will not be executed if `params[:letter_form]` is `true` when using `&&`
You can probably replace
```
unless params[:letter_form] and params[:letter_form][:letter_template_id]
```
with
```
unless params.dig(:letter_form, :letter_template_id)
 ...
end
```

app/controllers/cases/letters_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  private
+
+  def set_case
+    @case = Case::Base.find(params[:case_id])&.decorate
+  end
+
+  def type_name(type)
+    LetterTemplate.template_types[type] || 'unknown'
+  end
+end
```
mseedat-moj: `def type_name` should be in the `LetterTemplate` model

app/controllers/cases/letters_controller.rb
Polsonby: oh nice

Polsonby (commented):
Renamed render letter to show and small tweaks. Agree about moving things into service, but would prefer to do this in a follow up PR - the Word doc one is up next(ish)

mseedat-moj:
Generally in the right direction

Some comments in no particular order:

- You seem to have upgraded to Postgress 11? As a result the structure.sql file makes significant changes. Production is on 9.5
- The parameter `letter_template_id` should just be `template_id` so that in use it reads easier (`@letter.template_id`)
- In routes.rb file, the route for `/letters` can be rejigged so that individual `get` paths don't need to be defined
- No Seeds?
- if using ERB templates then does the letter_templates table need to exist? e.g. Easier to design and edit ERB files in the  file system, in the same way ActiveMailer templates are just .erb files in the actual filesystem? Or need to have more meta-data for each template type going forward?
- `render` function currently has an unnecessary temporary variable assignment
```
  def render(values)
    ERB.new(body).result(binding)
  end
```  

- At the moment the routes make a distinction between /dispatch/new and /acknowledgement/new. What's the reason for that?

Polsonby:
Thanks Mo! 

> You seem to have upgraded to Postgress 11? As a result the structure.sql file makes significant changes. Production is on 9.5

I've not upgraded, that's the version that brew installed originally. I noticed this and trying to work up the courage to remove and reinstall the right version! As you'll see from recent commits I've been trying to work through what's different and got there in the end

> The parameter letter_template_id should just be template_id so that in use it reads easier (@letter.template_id)

maybe... I think that's subjective. 

> In routes.rb file, the route for /letters can be rejigged so that individual get paths don't need to be defined

True, I'm not sure it's worth the time though, we got rid of render but there's a word generation coming

> No Seeds?

WIP

> if using ERB templates then does the letter_templates table need to exist? e.g. Easier to design and edit ERB files in the file system, in the same way ActiveMailer templates are just .erb files in the actual filesystem? Or need to have more meta-data for each template type going forward?

Good point, but yes it does need to exist; we have templates of different types and in future might wish to have CMS editing of these. 

> render function currently has an unnecessary temporary variable assignment def render(values)

It's actually necessary

> At the moment the routes make a distinction between /dispatch/new and /acknowledgement/new. What's the reason for that?

It's how the system distinguishes between different template types when displaying the list of available templates of that type

Polsonby:
Amended routes as suggested, thanks @mseedat-moj

app/controllers/cases/letters_controller.rb
Polsonby: Refactored the controller - I agree about possibly using a service for this, will take a look at that in the next PR when we add Word document export

mseedat-moj (approved):

PR #924, opened 2019-09-24
Exclude filter specs outside of working hours

Polsonby:
These filter specs work great during the normal working day, but failed
when run out of hours or at the weekend. Since this code doesn't change
much and it proved problematic to solve using Timecop, we chose to only
run them during working hours when they run properly

## Description
Following Mo's excellent suggestion, made the recalcitrant spec conditional and only run when it has a chance of passing
 
## Self-review checklist

* [x] (3) Tests passing

### DIFF
--- a/spec/features/cases/filters/external_deadline_spec.rb
+++ b/spec/features/cases/filters/external_deadline_spec.rb
@@ -1,7 +1,7 @@
 require "rails_helper"
 
 def working_hours
-  Date.today.workday? &&  Time.now.during_business_hours?
+  Date.today.workday? && Time.now.during_business_hours?
 end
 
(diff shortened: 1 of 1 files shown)


spec/features/cases/filters/external_deadline_spec.rb
```diff
@@ -1,6 +1,10 @@
 require "rails_helper"
 
-feature 'filtering by external deadline' do
+def working_hours
+  Date.today.workday? &&  Time.now.during_business_hours?
```
mseedat-moj: Extra space after ampersands

mseedat-moj (commented):
Brill very annoying. minor change request.

danielatdigital (approved):

PR #941, opened 2019-10-09
fix-duplicate-key Try to avoid duplicate key errors

Polsonby:
For some reason two or more workers are attempting to run the same sidekiq job simultaneously, both creating a new Warehouse::CaseReport and trying to save it

This fails for the second one with a duplicate key exception on case_id

Experiment with persisting the record to the database as soon as we find it's nil

## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/spec/models/warehouse/case_report_spec.rb
+++ b/spec/models/warehouse/case_report_spec.rb
@@ -30,17 +30,19 @@ RSpec.describe ::Warehouse::CaseReport, type: :model do
       expect(kase.warehouse_case_report).to be_nil
 
       case_report = described_class.for(kase)
-      expect(case_report.new_record?).to be true
+      kase.reload
+
+      expect(kase.warehouse_case_report).to eq case_report
     end
 
     it 'returns existing CaseReport if it exists' do
       kase = create :accepted_sar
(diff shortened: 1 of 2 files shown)


spec/models/warehouse/case_report_spec.rb
```diff
@@ -30,17 +30,19 @@
...
 
       case_report = described_class.for(kase)
-      expect(case_report.new_record?).to be true
+      kase.reload
+
+      expect(kase.warehouse_case_report).to eq case_report
     end
 
-    it 'returns existing CaseReport if it exists' do
+    fit 'returns existing CaseReport if it exists' do
```
mseedat-moj: what is fit?

mseedat-moj (approved):

spec/models/warehouse/case_report_spec.rb
Polsonby: balls, sorry - that's rspec shorthand for focusing a test. Really bad that I left it in there - it prevents other tests in the file from executing

Polsonby: Removed!

PR #946, opened 2019-10-15
CT-2450 Add deactivation to r003

danielatdigital:
## Self-review checklist

* [x] (3) Tests passing
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2452?atlOrigin=eyJpIjoiMTJiNTVlZDUwOTI0NGUyODgxYjgwNTRmMTBiOTQ2MzAiLCJwIjoiaiJ9

### Deployment
No migrations, this uses current model structure
 
Deactivate a team, run the report that covers it's closed questions.
Check that the deactivated date shows in the correct columns.

Report pre-change:
Screen Shot 2019-10-15 at 10 46 52

Report post-change:
Screen Shot 2019-10-15 at 10 47 12

### DIFF
--- a/app/models/team.rb
+++ b/app/models/team.rb
@@ -150,6 +150,14 @@ class Team < ApplicationRecord
     name.remove(DEACTIVATED_LABEL, /@\((.)*\)/).strip
   end
 
+  def deactivated
+    if active?
+      ""
+    else
+      deleted_at.strftime("%d/%m/%Y")
+    end
+  end
+
   private
 
   # This method applies to Business Groups and Directorates only.
--- a/app/services/stats/base_business_unit_performance_report.rb
+++ b/app/services/stats/base_business_unit_performance_report.rb
@@ -5,9 +5,10 @@ module Stats
       business_group:                  'Business group',
       directorate:                     'Directorate',
       business_unit:                   'Business unit',
-      responsible:                     'Responsible'
+      responsible:                     'Responsible',
+      deactivated:                     'Deactivated',
     }
-    
+
     R003_BU_PERFORMANCE_COLUMNS = {
       bu_performance:             'Performance %',
       bu_total:                   'Total received',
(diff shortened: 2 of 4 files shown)


app/models/team.rb
```diff
@@ -150,6 +150,14 @@ def original_team_name
     name.remove(DEACTIVATED_LABEL, /@\((.)*\)/).strip
   end
 
+  def deactivated
+    if active?
+      ""
+    else
+      deleted_at.strftime("%d/%m/%Y")
+    end
+  end
```
Polsonby: Nothing wrong with this at all, but just for discussion's sake :) 

Ruby style tends to shy away from explicit indented conditionals and people often try to avoid them - you often see alternative phrasings preferred instead, like a guard clause returning early:

```
  def deactivated
    return "" if active? 
    deleted_at.strftime("%d/%m/%Y")
  end
```
or just sending messages to objects - this example essentially cheats by using the safe navigation operator `&` to hide the if statement and do a concise check for nil, which is what `active?` does  - you'll see this pattern used elsewhere in the file

```
  def deactivated
    deleted_at&.strftime("%d/%m/%Y").to_s
  end
```

Regarding the method name, I'm wondering whether `deactivated` sounds a bit like a boolean - would it make more sense to call it `deactivated_date` or something like that?

Polsonby (commented):
Good stuff :) - LGTM - happy to approve but thinking Mo should have a look too as it's reports

app/models/team.rb
Polsonby: All very subjective! Final thought, the way it's already done elsewhere in the file is probably clearer than mine, actually...

```  
  def team_leader_name
    team_leader&.value || ''
  end

  def deactivated_date
    deleted_at&.strftime("%d/%m/%Y") || ''
  end
```

Polsonby (approved):

PR #951, opened 2019-10-28
CT-2585 Update navigation config and hide search

digitalali-moj:
## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots
OLD:

image
NEW:
image


spec/features/common/primary_navigation_spec.rb
```diff
@@ -49,30 +49,32 @@
...
-      expect(nav_links[0]).to have_text('New cases')
-      expect(nav_links[1]).to have_text('Cases')
-      expect(nav_links[4]).to have_text('Search')
-      expect(nav_links[5]).to have_text('Team')
+      expect(nav_links.count).to eq 7
+      expect(nav_links[0]).to have_text('Create case')
+      expect(nav_links[1]).to have_text('New cases')
+      expect(nav_links[2]).to have_text('Cases')
+      expect(nav_links[5]).to have_text('Search')
+      expect(nav_links[6]).to have_text('Team')
```
Polsonby: The numbering looked weird before, and it looks weird now only in a slightly different way :)   - any idea why that was, @digitalali-moj ? Why isn't it sequential? Is there something we're meant to have and don't for accessibility or something?

mseedat-moj (approved):
@digitalali-moj  Looks good but there's a question from @Polsonby  that requires answering.

spec/features/common/primary_navigation_spec.rb
digitalali-moj: Thanks for adding the missing items. The sequence still matches the if you account for the hidden elements, which can be seen with CSS turned off.

PR #956, opened 2019-11-04
CT-2228 Basic word export working

Polsonby:
## Description
User needs to be able to download the generated letter as a Word document. Added an alternate rendering format for :docx, a link hitting same, and amended the show action to produce the document. Some basic tests and a little refactoring of the Letter class so it now wraps letter template and Case. 
 
## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots

image

image

 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2228
 
### Deployment
We need to run the seeder again, as some of the balance of stuff in the database versus the Word template/view file has changed a bit. `rake db:seed:dev:letter_templates` locally if you want to try it out, `rake db:seed:letter_templates` required on the server after deployment. 
 
Case in Waiting for data - press the "Send acknowledgement letter" button. Choose the template, view the letter, click "Save as Word" to download a file and check the replaced fields look sensible


app/models/letter.rb
```diff
@@ -8,8 +8,22 @@ class Letter
...
+  end
+
+  def values
+    @case
+  end
+
+  def letter_date
+    Date.today.strftime('%e %b %Y')
   end
 end
```
Polsonby: Regarding the initialize method signature... Can now pass in an optional Case instance. Annoyingly this still takes an integer for letter template ID - would prefer it took a LetterTemplate instance but that breaks how this is used as a form object when building a new letter, so left as-is for this iteration. In future iteration might split out an actual form object for that role and make this Letter class depend on a valid LetterTemplate

Polsonby: I chose not to explicitly delegate named methods through to the case - instead just expose the case as "values" for the letter. This is because I don't want Letter to know so much about Case.

mseedat-moj (approved):
Great work

danielatdigital (approved):

PR #968, opened 2019-11-11
CT-2589 Move team members to new business unit

danielatdigital:
## Self-review checklist

* [x] (3) Tests passing
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2589?atlOrigin=eyJpIjoiMGZhY2Y3ZDIyYTAyNDkwMWE4ODBlZGMyZGI1MWQ3MzQiLCJwIjoiaiJ9


app/services/team_move_service.rb
```diff
@@ -34,6 +34,8 @@ def transfer_team
     @new_unit.name << " (Moved from #{@business_unit.directorate.name})"
     @new_unit.correspondence_type_roles = @business_unit.correspondence_type_roles
     @new_unit.properties = @business_unit.properties
+    @new_unit.user_roles = @business_unit.user_roles
+    @business_unit.user_roles.delete_all
```
mseedat-moj: Would prefer to replace `@business_unit.user_roles.delete_all` with a method such as `@business_unit.destroy_related_user_roles!` or similar to describe intent here

spec/services/team_move_service_spec.rb
```diff
@@ -35,8 +35,17 @@
...
+      it 'Moves team user_roles to the new team, and removes them from the original team' do
+
+        tur = TeamsUsersRole.create user: create(:user),
+                                    team: bu,
+                                    role: 'manager'
+
+        team = TeamMoveService.new( bu, target_dir)
+        expect(team.new_unit.user_roles.first).to eq tur
+        expect(bu.user_roles.first).to eq nil
+      end
```
mseedat-moj: Formatting is a bit off, i.e. 

- test name string should not be capitalised (moves/Moves)
- should be newline Line 38
- no new line on Line 40
- no space inopening bracket Line 44
- prefer human readable variable names rather than abbr e.g `tur/team_user_role`
- can use `be_nil` rather than `eq nil` but that's personal preference imo

i personally prefer formatting of multi-line statements as follows:

```
team_user_role = TeamsUsersRole.create(
  user: create(:user),
  team: bu,
  role: 'manager'
)
```
That helps avoid random right-alignment issues and keeps things tidy/easy to read in column format which works nicely on widescreens.

Regarding execution, I notice the `TeamMoveService` performs the required actions on construction, rather than requiring an additional method call e.g. `TeamMoveService.new(bu, target_dir).call` or similar. Is that a deliberate decision? (Many of the services require .call with appropriate flags set).

mseedat-moj (changes requested):
General functionality is fine, some formatting suggestions, and a single general question:

- How is the Team Move Service guaranteed atomicity? e.g. transactions, flags etc and therefore how is failure handled? Is that for the next phase of work?

Atomicity handled in a later ticket

spec/services/team_move_service_spec.rb
danielatdigital: Thanks Mo, comments on formatting noted. 
Since the service does one thing only, it made sense to me to do the work on construction rather than with .call + possible params. Simplest thing possible is my preference.

danielatdigital: Regarding atomicity, that's in another card https://dsdmoj.atlassian.net/browse/CT-2591?atlOrigin=eyJpIjoiOTAzZjg0MzUzMThjNDA0MjhmZTA3MjQ5ZmEwMDQ5ZWUiLCJwIjoiaiJ9

app/services/team_move_service.rb
danielatdigital: Done

mseedat-moj (approved):
Good job, thanks for the changes:

Had a telephone call to discuss the following:

- TeamMoveService.call transaction should be _inside_ the try/catch block
- move_team would not have thrown an exception if save failed, use `.save!`
- renamed `moved_team` to `move_team!` to reflect exceptional nature

PR #970, opened 2019-11-12
redirect all requests and display a maintenance page

Polsonby:
## Description
In order to prevent users from editing the database or uploading files - or changing any sort of state, in fact - during our migration to Cloud Platform, this PR introduces "Maintenance mode"

When a file is present in the rails root directory named `maintenance.txt` then the app redirects all HTTP requests to maintenance page. 

When the file is not present, the app functions as normal. 

Can't do a feature spec for this because it will break parallel tests - as soon as we write the file for the test in one thread to check the app behaviour, any other tests will break. 

But as long as the test suite passes otherwise, I think we should be good. 
 
## Self-review checklist

* [ ] (3) Tests passing
 
### Screenshots

image

image
 
image

image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2602
 
### Deployment
This has no migrations. 
 
In the Rails app root, create a file called 'maintenance.txt'

```touch maintenance.txt```

Verify all requests hit the maintenance page by going to various urls such as '/cases/open', '/users/sign_in/' 

Remove the file 

``` rm maintenance.txt```

And verify that all requests work as usual.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -37,7 +37,7 @@ class ApplicationController < ActionController::Base
   private
 
   def check_maintenance_mode
-    if maintenance_mode_on? and request.fullpath != '/maintenance'
+    if maintenance_mode_on? && request.fullpath != '/maintenance'
       redirect_to '/maintenance' and return
     end
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -30,17 +30,21 @@ class ApplicationController < ActionController::Base
   end
 
   def maintenance_mode
+    redirect_to '/' and return unless maintenance_mode_on?
     render layout: nil, file: "layouts/maintenance"
   end
 
   private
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: app/controllers/application_controller.rb [hunk 1]
```diff
@@ -27,8 +29,23 @@ def set_user
...
+    render layout: nil, file: "layouts/maintenance"
+  end
+
   private
 
+  def check_maintenance_mode
+    if maintenance_mode_on? and request.fullpath != '/maintenance'
+      redirect_to '/maintenance' and return
+    end
+  end
```
mseedat-moj: Replace `if maintenance_mode_on? and` with &&

mseedat-moj (approved):
Minor change otherwise looks good

app/controllers/application_controller.rb
Polsonby: Good spot, thanks! Done

PR #982, opened 2019-11-20
CT-2604 Deployment, replication and rolling deployment

Polsonby:
## Description
Updating deployment scripts - this PR provides two shell scripts, one to build, tag and push an image of the desired branch up to the ECR, and the other to take the tag for that image and deploy it to a given namespace. The scripts are presently in the site root but we'll move them into the config/kubernetes/scripts folder I imagine

This PR also adds more replicas and asks for rolling restarts

```  
  replicas: 4
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 0
      maxSurge: 1
  selector:
    matchLabels:
      app: track-a-query
```

Build and deploy usage looks like this:

```
    # build the app to get an image tag
    ./build.sh
    ...many lines of output...
    Image created with tag: cts-CT-1234-my-lovely-branch-6bece953

    # deploy image-tag to dev
    ./deploy.sh cts-CT-1234-my-lovely-branch-6bece953 dev

    # deploy latest image of master to production
    ./deploy.sh cts-master-8f3tgba3 production

```

The build script deploys from git based on the currently checked out working copy you run it inside; and doesn't suck in changed files from the working directory. 

It checks to see whether you have modified or uncommitted files and warns you but it will proceed. 

It checks to see whether the checked out commit has been pushed to GitHub and exits if not; the implication being that CI tests have not been run on this code.

When successful, it tags and pushes the image up to the docker ECR and outputs the tag that you'll need for deployment. We shouldn't use :latest anywhere, but we push and tag an image using :latest just in case, as it's the default in the deployment.yml. Actual deployments use the tag you pass to it, so it's just a safety net having something decent there under the tag :track-a-query-latest. 

The tag format is a bit long but information rich - it comprises the service prefix, branch name and commit sha.

image

The deploy script takes two required arguments; a valid-looking image tag name and a valid deployment namespace environment. It also checks whether the repository has been unlocked with git-crypt and exits if not.

It has a safety catch asking you to double check what you're deploying and to where. Deploying to production requires a different response to mitigate against "up arrow / enter / Y / enter" errors.  

image

We can remove the safety catch when we make a version of the script for automated build and deployment, but I like the additional step personally.

## Self-review checklist

* [x] (3) Tests passing
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2604
 
### Deployment
See above
 
See above


build.sh
```diff
@@ -0,0 +1,98 @@
...
+          --build-arg BUILD_TAG=$docker_build_tag \
+          --pull \
+          --tag ${docker_registry_tag} \
+          --file ./Dockerfile \
+          $git_fetch_url
+
+  # 7. Tag and push the image to the ECR
+  case $current_branch in
+    master)
+      latest_tag=${docker_registry}:${component}-latest
```
mseedat-moj: Should this be cts rather than component?

Polsonby: nope - that's intentional. I don't want the deploy script to be able to deploy :latest tagged images

mseedat-moj (approved):
Top work

build.sh
```diff
@@ -0,0 +1,98 @@
...
+    p "Deploying from a clean working directory..."
+  else
+    p "\e[31mWarning! Working directory contains new or modified files...\e[0m\n"
+    git status --porcelain
+  fi
+  p "------------------------------------------------------------------------"
+
+  # 3. Get a logged in context so we can push images to the ECR
+  p "Docker login to registry (ECR)..."
+  $(aws ecr --profile "$aws_profile" get-login --no-include-email --region "$region" --profile "$aws_profile")
```
Polsonby: Nick, this is the bit I was talking about earlier in passing - never bothered to stop and think about it, but the `$(...)` here is the equivalent of the copy/paste you were doing when getting the authentication to the ECR - it generates the command and executes it in one statement.

PR #990, opened 2019-11-22
CT-2562 Fix roles on create case link and remove other button

digitalali-moj:
## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots
n/a no changes to f/e, just Create case not shown to non/manager and bmt
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2562
 
Log in as manager, you should see create case in header, and a green button
Log in as anyone else, you should see no create case in header and no button.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -113,6 +113,8 @@ global_navigation:
   pages:
     create_case:
       path: /cases/new
+
+      visibility: manager
       
     incoming_cases:
       path: /cases/incoming
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -73,7 +73,6 @@ homepage_navigation:
   pages:
     incoming_cases:
       path: /cases/incoming
-
       visibility: approver
       scope:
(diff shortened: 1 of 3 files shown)


Polsonby (commented):
LGTM! Presumably we already amended all the tests to use the new one which is why no tests changed in this PR?

I have serious reservations about removing the button, though, I pinged Alex to mention it. I don't see why adding it in the main nav means we can't also have it on the page, I think it's really good in the current location. Anyway, happy to approve PR if we're really going to get rid of it :)

config/settings.yml
```diff
@@ -113,6 +113,8 @@ global_navigation:
   pages:
     create_case:
       path: /cases/new
+
```
Polsonby: Probably don't need this whitespace, maybe remove?

digitalali-moj: yes, removed, thanks

Polsonby (approved):
LGTM :) 👍

PR #995, opened 2019-11-28
CT-2645 Add new method to finder service which works with rejected cases

Polsonby:
## Description
"ActionView::Template::ErrorCasesController#show
No accepted assignment with role 'responding' for user 1061 on case 16493" https://sentry.service.dsd.io/mojds/correspondence-tool-staff/issues/39150/

The problem is the case has an attachment uploaded by a user whose team is no longer assigned to the case, because of the case being rejected after upload. 

It uses this method from the TeamFinderService 

 ```
 def team_for_unassigned_user
    assigned_teams = @kase.assignments.where(state: [:accepted, :pending], role: @assignment_role).map(&:team)
    raise UserNotFoundError.new(self) if assigned_teams.empty?
    user_teams = @user.team_roles.where(role: @team_role).map(&:team)
    (assigned_teams & user_teams).first
  end

```
Solution here is create a new method which returns the team whether or not it is currently assigned to the case.

## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots
n/a
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2645
 
### Deployment
n/a
 
Not massively easy to reproduce locally, but you can reproduce by finding a case which has an attachment uploaded by a user belonging to a particular team. Find the `assignment` for that team on the case, and set the status to "rejected" View the case and it should throw an exception on master. 

On production there is a case which demonstrates the issue; https://track-a-query.service.justice.gov.uk/cases/16493 - if we view this as an admin on Correspondence Tool for Staff Production after deployment it should be viewable instead of giving an error. 

image

### DIFF
--- a/app/models/case/base.rb
+++ b/app/models/case/base.rb
@@ -405,6 +405,10 @@ class Case::Base < ApplicationRecord
     TeamFinderService.new(self, user, role).team_for_unassigned_user
   end
 
+  def team_for_user(user, role)
+    TeamFinderService.new(self, user, role).team_for_unassigned_user
+  end
+
   def allow_event?(user, event)
     state_machine.permitted_events(user.id).include?(event)
   end
--- a/app/models/case/base.rb
+++ b/app/models/case/base.rb
@@ -406,7 +406,7 @@ class Case::Base < ApplicationRecord
   end
 
   def team_for_user(user, role)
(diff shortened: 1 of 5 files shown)


mseedat-moj (approved):
Well done and looks like a simple, sensible fix!

PR #997, opened 2019-12-02
Ct 2588 move business unit open cases

danielatdigital:
## Self-review checklist

* [x] (3) Tests passing


app/services/team_move_service.rb
```diff
@@ -30,29 +30,30 @@ def initialize(team, directorate)
...
-
     @new_team.directorate = @directorate
     @new_team.name << " (Moved from #{@team.directorate.name})"
     @new_team.correspondence_type_roles = @team.correspondence_type_roles
     @new_team.properties = @team.properties
     @new_team.user_roles = @team.user_roles
-
     @team.destroy_related_user_roles!
-    @new_team.save!
+    @new_team.save
```
Polsonby: Wondering about the removal of bang here, which is probably a good idea, but whether we could go a step further and return something truthy from the method if it all goes well 

Not sure if would give any benefit or not, any thoughts?

Polsonby (approved):
Looks good to me - just one question

PR #1005, opened 2019-12-10
CT-2588 Move open cases additional

Polsonby:
## Description
We found some edge cases through testing; this PR addresses situations where users might not get moved along with the team, making the tests and code work for business units with "code" set and without, and for trigger cases ensuring that the approving team stays the same after the move.  
 
## Self-review checklist

* [x] (3) Tests passing
 
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2292
 
Move a team :) 

All the users should still be there, and cases belonging to them should still be viewable (was throwing an exception) - check trigger and non-trigger cases.

### DIFF
--- a/app/services/team_move_service.rb
+++ b/app/services/team_move_service.rb
@@ -56,11 +56,12 @@ class TeamMoveService
     @new_team.directorate = @directorate
     # Team.name does not allow duplicates on validation for active teams,
     #  so to allow us to save the new team we must give it a unique name
-    @new_team.name << " (Moved from #{@team.directorate.name})"
+    @new_team.name = "(Moved from #{@team.directorate.name})"
+    # Duplicate @team.code is a problem too
+    @new_team.code = "#{@team.code}-NEW"
     @new_team.correspondence_type_roles = @team.correspondence_type_roles
     @new_team.properties = @team.properties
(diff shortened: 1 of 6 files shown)


app/services/team_move_service.rb
```diff
@@ -81,10 +82,11 @@ def link_old_team_to_new_team
     @team.moved_to_unit = @new_team
     @team.save
   end
-  
+
   def restore_new_team_name_to_original_name
     # New team gets original team name to retain consistency for the users
     @new_team.name = @team.original_team_name
+    @new_team.code = @new_team.code.sub(/-NEW$/, "") unless @team.code.blank?
```
mseedat-moj: Does this need to be gsub or case insensitive?

mseedat-moj (approved):
Great work, minor question

app/services/team_move_service.rb
Polsonby: Great question! I suck at regex, but I don't think so... the idea is to only replace the pattern if found at the end of the string so I think it's ok as sub rather than gsub. Similarly, we're adding the -NEW in uppercase so I only want to replace it if it's uppercase

PR #1068, opened 2020-04-03
CT-2223 Edit case details

Polsonby:
## Description
This introduces the ability to edit Offender SAR cases, adding "change" links for each group of fields corresponding to the grouping on the Create Case journey

The form partials are common to both create and edit journeys, and the back link is now a helper so that it can be told which context it's to work in. In the create context, it works out the previous step from the "steps" defined for the form object. 

There's changes to the gem file and bundler version on CircleCI that I now cannot remember why I've got included in this PR, and I am not sure these are needed now that master has moved on - this is party why I want to get this merged sooner rather than later and have the tidy up work done on master, not this

## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots

image

 
image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2223
 
### Deployment
none
 
Click change to edit the fields, ensuring the right form partial is displayed.
Continue should return to the case show page if you're editing. 
Form validation issues should reload the same view. Back should always take you back to the show page without saving any changes.

Need to check the case creation journey too for Offender SARs (log in as Brian Rix)


njpreddy (approved):
LGTM.  Just wondering, in app/controllers/concerns/setup_case.rb did you want to leave '# random commend'?

PR #1106, opened 2020-05-12
CT-2847 Make the "request info" as optional

ymao2:
## Description
Make the "request info" box as optional when creating offender_sar case 

### Related JIRA tickets
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=CT&modal=detail&selectedIssue=CT-2847

## Self-review checklist

* [x] (3) Tests passing
 
### Screenshots

Before: 
Screenshot 2020-05-13 at 14 23 31
After 
Screenshot 2020-05-13 at 11 21 53
Screenshot 2020-05-13 at 17 22 44

Click change
Screenshot 2020-05-13 at 11 21 22
Screenshot 2020-05-13 at 11 21 33

### Deployment

 No need

1 - login as Brian.rix
2 - Create a offender_sar case
3 - when reaching to the page of ```/cases/offender_sars/requested-info```
      leave it as blank and click "Continue" until to the end
4 - Once the case detail page is shown and move down to the request box, should see the message of "Use this space to store additional case information." in the box


digitalali-moj (approved):
Looks Good to Me!

app/views/cases/_case_request.html.slim
```diff
@@ -10,6 +10,9 @@
     - if case_details.message.present?
       p.message-body
         = show_hide_message(case_details)
+    - if !case_details.message.present? and case_details.offender_sar?
+      p.messag-body
```
Polsonby, line 14: Small typo

```suggestion
      p.message-body 
```

The number of conditionals is becoming too much in this file, I think. At this point I'd think about creating an offender_sar flavoured partial and using that instead

Polsonby (changes requested):
Great work, nearly there! - just one typo and question over whether it's time to split out a separate partial

app/views/cases/_case_request.html.slim
Polsonby: Oh, also is the `and` intentional here? There's a difference between using it for logical operations versus flow control. If it's the former, and I think it might be here, it's preferred to use `&&`

This contrasts with situations like `render json: @user and return` etc

Polsonby: ```if !case_details.message.present? and case_details.offender_sar?```

is the `and` intentional here? There's a difference between using it for logical operations versus flow control. If it's the former, and I think it might be here, it's preferred to use `&&`

This contrasts with situations like `render json: @user and return` etc

Also, we'd use if if `case_details.message.empty?` in place of `!case_details.message.present?`

Polsonby: Just thinking about it, maybe we should add a view spec for the logic displaying the alternate message.

spec/views/cases/_case_request_html_slim_spec.rb
```diff
@@ -36,20 +36,8 @@
   let(:empty_message) { "" }
```
Polsonby, line 36: we don't need this to be a let (also, the preferred style (which I detest) is single quotes unless you're interpolating) - we can delete this

spec/views/cases/_case_request_html_slim_spec.rb
```diff
@@ -36,20 +36,8 @@
...
 
   describe 'Displaying help message when the request of an offender-sar case is empty' do
-    let(:offender_sar_case) {
-      (create :offender_sar_case, subject_aliases: 'John Smith',
-              date_of_birth: '2019-09-01', message: empty_message).decorate
-    }
+    let(:offender_sar_case) { (create :offender_sar_case, message: empty_message).decorate }
```
Polsonby, line 39: Cool - we don't need to decorate the case, either. 

If you're using a let to create context, it should be inside the describe block to which it belongs. In this case I'd lose the let and just put the empty string in the create statement but if you _really_ want the additional context from the variable name, perhaps you could move the `let(:empty_message) { "" }` to just above this line

```suggestion
    let(:offender_sar_case) { create :offender_sar_case, message: '' }
```

Polsonby (approved):
Great job 👍 +1

PR #1114, opened 2020-05-16
CT-2403 words change->case history &added "create"

ymao2:
## Description
1- Changed the wording for "action" column for case-history section on case detail page
2- Added "create" event as the initial transition after a case is created
 
## Self-review checklist

* [X] (3) Tests passing
 
### Screenshots
Before
Screenshot 2020-05-16 at 12 24 56

After
Screenshot 2020-05-16 at 12 16 48
Screenshot 2020-05-16 at 12 19 04

 
### Related JIRA tickets
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=CT&modal=detail&selectedIssue=CT-2403
 
### Deployment

 No need

Need to cover all the flows of every case


config/locales/event.en.yml
```diff
@@ -45,3 +46,8 @@ en:
     mark_as_vetting_in_progress: Mark as vetting in progress
     mark_as_ready_to_copy: Mark as ready to copy
     mark_as_ready_to_dispatch: Mark as ready to dispatch
+    mark_as_waiting_for_data__desc: Case is waiting for data
+    mark_as_ready_for_vetting__desc: Case is ready for vetting
+    mark_as_vetting_in_progress__desc: Vetting in progress
+    mark_as_ready_to_copy__desc: Case is ready to copy
+    mark_as_ready_to_dispatch__desc: Case is ready to dispatch
```
Polsonby, line 53: Great :) - this hasn't come up before, but just so you know we try to keep keys in alphabetical order in files, as much as possible. Would you mind moving these to appropriate locations?

Polsonby: You'll usually spot other things out of place when you start looking - up to you whether you fix any "broken windows" on the way past. It's really good to do it, cleaning things as you go even if they're not yours.... 

 - on the other hand, it's better to keep a PR focused on just what the ticket is about. 

So your call! Ideally do any other improvements when you see them but if there's more than just one or two, I'd open a separate PR we can just fast-track

spec/controllers/cases_controller_spec.rb
```diff
@@ -56,7 +56,7 @@
 
     it 'decorates the collection of case transitions' do
       expect(assigns(:case_transitions)).to be_an_instance_of(Draper::CollectionDecorator)
-      expect(assigns(:case_transitions).map(&:class)).to eq [ CaseTransitionDecorator ]
+      expect(assigns(:case_transitions).map(&:class)).to eq [ CaseTransitionDecorator, CaseTransitionDecorator ]
```
Polsonby, line 59: This is making my spidey-senses tingle - something not quite right here?

spec/factory_specs/fois_spec.rb
```diff
@@ -127,7 +127,7 @@
         expect(managing_assignment.state).to eq 'accepted'
         expect(managing_assignment.team).to eq team_disclosure_bmt
 
-        expect(kase.transitions.size).to eq 0
+        expect(kase.transitions.size).to eq 1
```
Polsonby, line 130: Did Alex know that this change would affect ALL case types, and not just Offender SARs? I think it's great, but just checking... :)

Polsonby, line 130: Did Alex know that this change would affect ALL case types, and not just Offender SARs? I think it's great, but just checking... :)

spec/models/case/base_spec.rb
```diff
@@ -79,6 +79,47 @@
     it { should validate_presence_of(:type)          }
   end
 
+  context 'an initial transition is created after any case is create' do
```
Polsonby, line 82:
```suggestion
  context 'an initial transition is created after any case is created' do 
```

spec/models/case/base_spec.rb
```diff
@@ -79,6 +79,47 @@
...
+
+    it 'sar_case' do 
+      expect(sar_case.transitions.size).to eq 1
+      expect(sar_case.transitions.first.event).to eq "create"
+    end
+
+    it 'offender_sar_case' do 
+      expect(offender_sar_case.transitions.size).to eq 1
+      expect(offender_sar_case.transitions.first.event).to eq "create"
+    end
```
Polsonby, line 99: There's a thing I am REALLY bad at, which is looking after tests in the same way you'd look after code. In a normal file you'd never have the same logic repeated like this, you'd make extract the behaviour into a method and DRY up the code, right? Same in tests. It can make tests harder to read but there's always a tradeoff and in this case I think we're well into the right territory for extracting this stuff into
 
1. A shared example called something like `case_has_default_transition` and your examples become more like:
```
    context 'sar_vase' do
      it_behaves_like 'case_has_default_transition'
    end
```
and probably then 2. an array of correspondence types and the above adapted to be used in a loop over them

Look through the tests for "behaves_like" to see how we've used it elsewhere and check out the docs https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples

Polsonby (commented):
This looks great - some questions and a thought on test organisation below :)

spec/factory_specs/fois_spec.rb
ymao2: @Polsonby  yes I did mention this to her, she said it is good to have "created event" in the case history, as it is one of important feedback from the testing running with Branton officers,  so my idea would be to deploy this ticket to dev and ask her to give a run for all those cases (to be honest, there is no any change to other case types only change is this new event)

ymao2: @Polsonby  please check my comment for your similar feedback for another change

app/decorators/case_transition_decorator.rb
```diff
@@ -24,6 +24,15 @@ def event
     state_machine.event_name(object.event)
   end
 
+  def event_desc
+    specific_key = "event.case/#{object.case.type_abbreviation.downcase}.#{object.event}"
+    specific_key_with_desc = "event.case/#{object.case.type_abbreviation.downcase}.#{object.event}__desc"
+    default_key = "event.#{object.event}"
+    default_key_with_desc = "event.#{object.event}__desc"
+    I18n.t(specific_key_with_desc, default: nil) || I18n.t(default_key_with_desc, default: nil) \
+    || I18n.t(specific_key, default: nil) || I18n.t(default_key, default: nil) || event
```
Polsonby, line 33: I think this method is a tad unclear, and took me a while to work out what's going on. It looks like it could maybe do with being split into a few smaller methods?

I'd love to see a nicer way of structuring that massive line of "or"s and the default "nil" looks like these are hidden conditionals - is the idea we're checking for a translation for the first one then if it's nil we try the second and so on?

I think if you start moving things out into one-line, well-named private methods a nicer way forward might emerge

Also, maybe a `has_translation_for(key)` method something like this?
``` 
def has_translation_for(
  I18n.t(key, default: nil)
end
```
Or maybe not. Anyway, do you want to have another run at it?

spec/controllers/cases_controller_spec.rb
Polsonby: Gotcha! Yeah, maybe there is a way to express the intent of the test a little more explicitly, then, but your explanation helped a lot! - probably looks better in the file itself than out of context here

config/locales/event.en.yml
ymao2: @Polsonby,  the changes I made is not for fixing any "broken windows", it is outcome of trying to add flexible extra-description for an event without breaking existing logical, ideally I would like to extend the properties of an event, but the way I had in my mind will change the structure of event.yml too much,  I can explain it tomorrow if you would like a call

spec/models/case/base_spec.rb
ymao2: @Polsonby let me re-coding this line ( actually I had a feeling you will pick up this line)

app/decorators/case_transition_decorator.rb
ymao2: @Polsonby let me re-coding this line ( actually I had a feeling you will pick up this line)

app/decorators/case_transition_decorator.rb
```diff
@@ -19,18 +19,24 @@ def event_and_detail
...
-  def event
-    state_machine = object.case.state_machine
-    state_machine.event_name(object.event)
+  def event_desc
+    description_for_event = has_detail_description_for_event
+    if description_for_event.nil?
+      event
+    else
+      description_for_event 
+    end
```
Polsonby, line 28:
```suggestion
    description_for_event || event
```

config/locales/event.en.yml
Polsonby: Sorry, that "broken windows" remark was not about your PR so far - it was  about when if you start looking, you'll see lots of places where things _aren't_ in alphabetical order and the hypothetical situation you might find yourself in where you might want to add other cases you saw in passing when you're changing the order of yours. Does that make sense? Let's get on a call tomorrow - sorry for not being clear enough!

Polsonby (commented):
How do you feel about those changes? Looks really good to me :) 

Tiny suggestion about how we tend to prefer to write simple conditionals, especially ones that check for nil

app/decorators/case_transition_decorator.rb
```diff
@@ -19,18 +19,24 @@ def event_and_detail
 
   private
 
-  def event
-    state_machine = object.case.state_machine
-    state_machine.event_name(object.event)
+  def event_desc
+    description_for_event = has_detail_description_for_event
```
Polsonby, line 23: You can get rid of this whole line if you like - we'd tend to reserve methods named things like "has_whatever_" for methods that return boolean values - we'd also name it with a trailing `?` in that case. Here I'd consider just calling the method `description_for_event` and losing the local variable altogether

spec/models/case/base_spec.rb
```diff
@@ -79,45 +79,22 @@
...
-      expect(ico_foi_case.transitions.order('created_at').size).to eq 2
-      expect(ico_foi_case.transitions.first.event).to eq "create"
-      expect(ico_foi_case.transitions.second.event).not_to eq "create"
-    end
-
-    it 'overturned_ico_foi' do 
-      expect(overturned_ico_foi.transitions.size).to eq 1
-      expect(overturned_ico_foi.transitions.first.event).to eq "create"
+  describe 'a default transition is created after a case is created' do 
+    shared_examples 'case has default transition' do |case_type|
```
Polsonby, line 83: I think this is loads better, what do you think?

Final wrinkle on this is, I think the shared examples (bizarrely and possibly unhelpfully) would normally go in their own folder `spec/support/shared_examples` - like a junk drawer. Almost all the ones in the app are in there

But I see Misha has also used them inline in one or two other places, so perfectly happy for this to live here 👍 

e.g. 
image

app/decorators/case_transition_decorator.rb
ymao2: @Polsonby  I was thinking about this but was worrying you dislike || as I did in my previous implementation :)

app/decorators/case_transition_decorator.rb
ymao2: @Polsonby , same thing I thought I may had better use your suggested name lol :), let me change it

spec/models/case/base_spec.rb
ymao2: thank you! was thinking to move it out as the above one did , but was thinking mine shared examples wont be used other places (at least for now), and plus I am not sure if I move it out, I may have other issues (trying to get less issue with tests right now, I got enough trouble with them already)

app/decorators/case_transition_decorator.rb
Polsonby: heh - sorry! I'm a nightmare, I must seem so inconsistent 😂 

Hopefully it makes sense that there's a kind of spectrum - say on the left hand end you have "no conditionals at all" (pure OO design ala Smalltalk) then in the middle somewhere you have "the least number of conditionals possible" to on the right hand side "lots and lots of conditionals" and very procedural code - generally speaking we're trying to move as far to the left on that spectrum as possible.

Really it comes from trying not to ask what type of object you're dealing with. Imagine you have a conditional where you check something to see what type of an object you have, and then send a different message depending which object it is. That's not good design. We'd rather inject a player of the role and send it a single message, and let the object work out how to respond to that message. 

Asking something whether it's an instance of nil is basically the same sort of thing, but it's a little further to the "left" on that hypothetical spectrum above - we just want to send messages to objects that play roles, but short of that we want to simplify the logic and know as little as we can about the objects we're dealing with. You can't send messages to nil, and you get nil back a LOT of the time. So lots of nil checks result and iturns out in this sort of app, there's a LOT of checking for nil going on so we try to hide it as much as possible. The above is one way. 

Another way you'll see a lot, while I'm thinking about it, is the safe navigation operator `&.` which basically wraps the message to the right in a check for nil

image

We also really like the ternary operator for reducing 5 line if/else statements to 1 line :) 
In Rails apps things like chained method calls and reaching across collections of objects are really common, for good or ill.

app/state_machines/configurable_state_machine/machine.rb
```diff
@@ -114,7 +114,7 @@ def respond_to_missing?(method, include_private = false)
     end
 
     def event_name(event)
-      if events.include?(event.to_sym)
+      if events.include?(event.to_sym) || event == @kase.class::CREATE_EVENT
```
Polsonby, line 117: Sorry! One last thing that occurs to me - why isn't our event in the `events` already? Seems like we should be adding it there, not adding it to this conditional as a special case. 

The state machine is one of the bits I think we should use "as intended" - did you manage to see where the events collection is set?

Polsonby (commented):
Great job and thanks so much for the changes. One last question to think about before merging, but otherwise LGTM! :) 

If you could just have a look at the state machine events, that would be amazing. Otherwise, happy to approve when you're ready 👍

Polsonby (approved):
Fantastic job, thank you!

PR #1131, opened 2020-05-27
CT-2449 Modify rules for SAR deadline date

ymao2:
## Self-review checklist

* [X] (3) Tests passing
 
### Related JIRA tickets
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=CT&modal=detail&selectedIssue=CT-2449


app/services/deadline_calculator/calendar_month.rb
```diff
@@ -0,0 +1,39 @@
...
+    class CalendarMonth
+  
+      attr_reader :kase
+  
+      def initialize(kase)
+        @kase = kase
+      end
+  
+      def escalation_deadline
+        kase.created_at.to_date + kase.correspondence_type.escalation_time_limit.days
```
Polsonby, line 15: Probably nothing to do with your PR, but wondering why this `to_date` here but not on the other deadline fields 🤔

db/data_migrations/20200527110228_modify_sar_deadline_class.rb
```diff
@@ -0,0 +1,19 @@
+class ModifySarDeadlineClass < ActiveRecord::DataMigration
+  def up
+    CorrespondenceType.all.each do |ct|
+      if ct.abbreviation == 'SAR' || ct.abbreviation == 'OFFENDER_SAR'
+        ct.deadline_calculator_class = 'CalendarMonth'
+        ct.save!
```
Polsonby, line 6: This is fine, but you can also consider 

```suggestion
        ct.update_attribute(:deadline_calculator_class, 'CalendarMonth')
```

There are many ways to update a property in Rails and some interesting wrinkles around which ones fire callbacks and which ones fire validations. The way you're doing it will definitely 'fail fast' and throw an exception if save doesn't work, but I believe so will `update_attribute`. Anyway, just for reference - I think the one liner is tidier generally

spec/factories/correspondence_types.rb
```diff
@@ -41,8 +40,6 @@
     abbreviation { 'OFFENDER_SAR' }
     escalation_time_limit { 3 }
     internal_time_limit { 10 }
-    external_time_limit { 30 }
-    deadline_calculator_class { 'CalendarDays' }
     report_category_name { 'Offender SAR report' }
```
Polsonby, line 44: Just wondering why we don't have the `deadline_calculator_class { 'CalendarMonth' }` here like the SAR one above?

Polsonby (commented):
Looks great so far :)

spec/services/deadline_calculator/calendar_month_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+                test_case = double('sar_case')
+                allow(test_case).to receive(:received_date).and_return(received_date_used)
+                deadline_calculator_local = described_class.new test_case
+                expect(deadline_calculator_local.external_deadline).to eq expected_deadline
+              end
+            end
+        end
+
+        context 'the date of one calendar month later is non working date' do
+          it 'weekend ' do
```
Polsonby, line 76: I wonder if we could add a bit more context for the descriptions, to make it clearer what the business rules are and why this class exists. 

For something like this, it's really good to run the specs with the `--documentation` flag on and make sure it generates readable sentences... This is one bit we really need the tests be our documentation :) 

Maybe a comment block in the CalendarMonth class too?

app/decorators/case_extend_sar_deadline_decorator.rb
```diff
@@ -8,4 +8,8 @@ class CaseExtendSARDeadlineDecorator < Draper::Decorator
   def allow_extension_period_selection?
     !object.deadline_extended? && object.deadline_extendable?
   end
+
+  def time_unit_string(time_period)
+    object.deadline_calculator.time_units_desc_for_deadline(time_period > 1)
```
Polsonby, line 13: We can probably use the built-in Rails pluralize helper instead of writing our own method to pluralize 'day' -> 'days'

app/models/case/sar/standard.rb
```diff
@@ -59,7 +59,11 @@ def self.searchable_fields_and_ranks
                  reply_method: :string,
                  late_team_id: :integer,
                  date_draft_compliant: :date,
-                 deadline_extended: [:boolean, default: false]
+                 # The following fields are due to the need of implementation
```
Polsonby, line 62: Not sure we need this comment?

app/models/case/sar/standard.rb
```diff
@@ -132,10 +136,11 @@ def initial_deadline
     end
   end
 
-  def extend_deadline!(new_deadline)
+  def extend_deadline!(new_deadline, new_extended_times)
     self.update!(
       external_deadline: new_deadline,
-      deadline_extended: true
+      deadline_extended: true,
+      extended_times: new_extended_times
```
Polsonby, line 143: Just wondering what type this expects? It isn't obvious to me from the variable name

app/models/case/sar/standard.rb
```diff
@@ -146,11 +151,20 @@ def reset_deadline!
     )
   end
 
-  # SARs extensions are based on calendar days not working days
+  # The deadlines are all calculated based on the date receiving the case
```
Polsonby, line 154: Wondering if this could be phrased as 
```suggestion
  # The deadlines are all calculated based on the date the case is received
```

app/models/case/sar/standard.rb
```diff
@@ -146,11 +151,20 @@ def reset_deadline!
...
   end
 
+  def extension_time_limit
+    self.correspondence_type.extension_time_limit || Settings.sar_extension_default_limit
+  end 
+
+  def extension_time_default
+    self.correspondence_type.extension_time_default || Settings.sar_extension_default_time_default
+  end 
+
```
Polsonby, line 167: Probably me, but I'm really struggling with the permutations of extension, time, default here :) potentially a bit confusing?
Is there any way we can make the purpose of these settings (and the difference between them) a bit clearer?

I think `sar_extension_default_time_default` has jumped the shark - anything that needs the word "default" in it twice needs to be split up into at least two other things, I think :)

app/models/case/sar/standard.rb
```diff
@@ -146,11 +151,20 @@ def reset_deadline!
     )
   end
 
-  # SARs extensions are based on calendar days not working days
+  # The deadlines are all calculated based on the date receiving the case
   def max_allowed_deadline_date
-    initial_deadline + Settings.sar_extension_limit.to_i.days
+    time_limit = self.correspondence_type.extension_time_limit || Settings.sar_extension_default_limit
```
Polsonby, line 156: I wonder if you could make this line into another private method called `max_time_limit` ?

app/models/correspondence_type.rb
```diff
@@ -25,13 +25,15 @@ class CorrespondenceType < ApplicationRecord
...
+                 extension_time_limit: :integer,
+                 extension_time_default:  :integer,
                  show_on_menu: [:boolean, default: true]
 
 
   enum deadline_calculator_class: {
          'BusinessDays' => 'BusinessDays',
          'CalendarDays' => 'CalendarDays',
-         'CalendarMonth' => 'CalendarMonth',
+         'CalendarMonths' => 'CalendarMonths',
```
Polsonby, line 36: I really like that this plural now matches the previous ones :)

app/services/deadline_calculator/calendar_days.rb
```diff
@@ -24,12 +24,24 @@ def external_deadline
       kase.received_date + kase.correspondence_type.external_time_limit.days
     end
 
+    def extension_deadline(time_limit)
+      kase.received_date + (time_limit + kase.correspondence_type.external_time_limit).days
+    end
+    
+    def max_allowed_deadline_date(time_limit=nil)
+      time_limit ||= (kase.correspondence_type.extension_time_limit || 0)
```
Polsonby, line 32: There's something weird about this method. At first I was looking at it thinking, can we get rid of the second, nested conditional here? But actually, I'm a bit weirded out by the fact this seems to require time_limit and yet it's an optional argument.

So we have three cases?

1. Pass in a time_limit - it just gets used
2. if time_limit omitted from the method call, default to the extension time_limit from the kase's correspondence type
3. if there's no extension_time_limit from there, use `0`

It feels like this method knows too much about other things - would it be a daft idea to think about correspondence_type.extension_time_limit always returning 0 if it's unset?

spec/support/features/interactions.rb
```diff
@@ -264,7 +264,7 @@ def extend_sar_deadline_for(kase:, num_days:, reason: 'The reason for extending'
       expected_case_history = [
         'Extended SAR deadline',
         "#{reason}",
-        " Deadline extended by #{num_days} days" # line-break character translates into a space
+        " Deadline extended by #{num_calendar_months} calendar month#{num_calendar_months > 1 ? "s" : ""}" # line-break character translates into a space
```
Polsonby, line 267: We can use pluralize here

app/services/deadline_calculator/business_days.rb
```diff
@@ -32,6 +32,21 @@ def business_unit_deadline_for_date(date)
...
+    
+    def max_allowed_deadline_date(time_limit=nil)
+      time_limit ||= (kase.correspondence_type.extension_time_limit || 0)
+      days_after_day_one = (time_limit + kase.correspondence_type.external_time_limit) - 1
+      days_after_day_one.business_days.after(start_date(kase.received_date))
+    end
+
+    def time_units_desc_for_deadline(plural=false)
+      plural ? "business days".freeze : "business day".freeze
+    end
```
Polsonby, line 48: assuming the argument is the number of days
```suggestion
    def time_units_desc_for_deadline(time_limit)
      "business #{'day'.pluralize(time_limit)}".freeze
    end
```

app/services/deadline_calculator/calendar_days.rb
```diff
@@ -24,12 +24,24 @@ def external_deadline
...
+
     def business_unit_deadline_for_date(date)
       deadline_method  = @kase.flagged? ? :internal_time_limit : :external_time_limit
       num_days = @kase.correspondence_type.__send__(deadline_method)
       date + num_days.days
     end
 
+    def time_units_desc_for_deadline(plural=false)
+      plural ? "calendar days".freeze : "calendar day".freeze
+    end
```
Polsonby, line 44: See comment about pluralize elsewhere :)

Polsonby (commented):
Great work - really nice. One or two things to think about to maximise clarity maybe?

njpreddy:
I'm happy if Ian is happy.

njpreddy (approved):

app/decorators/case_extend_sar_deadline_decorator.rb
ymao2: thank you will change to that

app/services/deadline_calculator/business_days.rb
ymao2: thank you for the suggestion!

app/models/case/sar/standard.rb
ymao2: it is integer, you can see the type from the top of "jsonb_accessor" part :)

app/models/case/sar/standard.rb
ymao2: will remove it

spec/services/deadline_calculator/calendar_month_spec.rb
ymao2: Added some documents on CalendarMonths,  will add more description here

app/models/case/sar/standard.rb
ymao2: Could do!

app/models/case/sar/standard.rb
ymao2: hmm, it must be my bad english sense,  let me rename it to "sar_extension_default_gap", it means the time-gap when you extent it every tme

app/services/deadline_calculator/calendar_days.rb
ymao2: yes what you guess is correct, if correspondence_type.extension_time_limit is unset,  I think it will return nil?

Polsonby (approved):
Brilliant, thanks for those changes :) 

Fab piece of work

PR #1132, opened 2020-05-27
CT-2403 Minor bug fix based on feedback from testing

ymao2:
## Description

2 changes based on the testing results from Alex
- Wording change from "Vetting in progress" to "Vetting is in progress"
- The table format is broken for offender sar case tye 

## Self-review checklist

* [X] (3) Tests passing
 
### Screenshots

 before 
Screen Shot 2020-05-26 at 16 51 16
after 
Screenshot 2020-05-27 at 18 06 04

### Related JIRA tickets

 https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=CT&modal=detail&selectedIssue=CT-2403

- login as Branton account
- make offender-sar case

### DIFF
--- a/app/views/cases/show.html.slim
+++ b/app/views/cases/show.html.slim
@@ -11,7 +11,7 @@
 
 = render partial: 'shared/components/case_page_heading', locals: {kase: @case}
 
-div id="case-#{@correspondence_type_key}" class="case"
+div id="case-#{@correspondence_type_key.gsub('_', '-')}" class="case"
   section.case-info.has-actions
     h2.visually-hidden
       = "Actions you can take on this case"
--- a/app/views/cases/show.html.slim
+++ b/app/views/cases/show.html.slim
@@ -11,7 +11,7 @@
 
 = render partial: 'shared/components/case_page_heading', locals: {kase: @case}
 
-div id="case-#{@correspondence_type_key.gsub('_', '-')}" class="case"
+div id="case-#{@correspondence_type_key == nil ? 'nil' : @correspondence_type_key.gsub('_', '-')}" class="case"
   section.case-info.has-actions
     h2.visually-hidden
       = "Actions you can take on this case"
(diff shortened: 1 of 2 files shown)


app/views/cases/show.html.slim
```diff
@@ -11,7 +11,7 @@
 
 = render partial: 'shared/components/case_page_heading', locals: {kase: @case}
 
-div id="case-#{@correspondence_type_key}" class="case"
+div id="case-#{@correspondence_type_key == nil ? 'nil' : @correspondence_type_key.gsub('_', '-')}" class="case"
```
Polsonby, line 14: This works great but there's a lot going on for one line in a view template. What do you think about maybe moving this out into a method in a decorator or a helper so we can name the concept and put a test around it?

Wondering where @correspondence_type_key is coming from, presumably one of the bazillion things SetupCase does or something like that?

ymao2: @Polsonby  this bug has been there for a while, just no one notice it until now, the change is nothing to do with any business logic, the issue I found is  if you put id='case-offender_sar', then somehow, it will try to append extra path to locate the css for table layout which I don't why 

IMO, this is purely for layout, it is not going to change anything and any value, I don't really feel it is valuable to add a test for testing whether this is a "visible" line there or not

config/locales/event.en.yml
```diff
@@ -46,7 +46,7 @@ en:
     mark_as_ready_for_vetting: Mark as ready for vetting
     mark_as_ready_for_vetting__desc: Case is ready for vetting
     mark_as_vetting_in_progress: Mark as vetting in progress
-    mark_as_vetting_in_progress__desc: Vetting in progress
+    mark_as_vetting_in_progress__desc: Vetting is in progress
```
Polsonby, line 49: If this didn't break the tests, means we don't have tests around this wording :) 

Do we need any? Do similar bits of the app have tests on the words displayed?

Generally speaking, we'd like to see one change in the app and one change in at least one spec corresponding to it. But this is never going to break the app, so I'd tend to be guided by how this sort of thing is tested presently.

ymao2: @Polsonby,  generally I agree with your principal,
But for this case, this is the wording change in yml which is the root source of any wording , There has been a test for checking whether the wording displayed on UI match the wording from the yml--- which make sense and is enough,  there is no need to define another source of bunch of wording (duplication of words in yml) for against the wording in yml

Polsonby (approved):

PR #1154, opened 2020-06-11
CT-2532 record total page count

njpreddy:
CT-2532 Display offender SAR total page count for data requests.

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
One Request:

1-request

More than one request and the total row is displayed:

2-request

3-requests

### Related JIRA tickets
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=CT&modal=detail&selectedIssue=CT-2532


app/views/cases/offender_sar/_data_requests.html.slim
```diff
@@ -27,11 +27,15 @@
...
               td
                 - if data_request.cached_date_received.present?
                   time datetime="#{data_request.cached_date_received}"
                     = l data_request.cached_date_received, format: :default
+              td
+                = data_request.cached_num_pages
               td
                 = link_to t('cases.data_requests.index.link'), edit_case_data_request_path(case_details, data_request), class: 'data-requests__action'
+
+          / The 'Total' row will only be displayed if more than one data request has been made.
```
Polsonby, line 39: Wondering if we even need the comment? the line below seems really clear :)

njpreddy: Comment now removed.

spec/features/cases/offender_sar/add_data_requests_spec.rb
```diff
@@ -51,6 +51,48 @@
     expect(last_row.total_value).to have_text '0'
   end
 
+  scenario 'total page count is correct', js: true do
```
Polsonby, line 54: This spec doesn't need to use the JS driver, which is much slower than running the feature spec using the default - the rack driver. Turning this off saves nearly 2 seconds running this spec :)

spec/features/cases/offender_sar/add_data_requests_spec.rb
```diff
@@ -51,6 +51,48 @@
...
+    data_request_edit_page.form.date_received_mm.fill_in(with: 6)
+    data_request_edit_page.form.date_received_yyyy.fill_in(with: 2020)
+    data_request_edit_page.form.num_pages.fill_in(with: 128)
+
+    click_on 'Update data received'
+
+    last_row = cases_show_page.data_requests.rows.last
+    expect(last_row.total_value).to have_text '160'
+  end
+
```
Polsonby, line 95: Thanks for adding the test! You've done the right thing in terms of a feature spec, clicking buttons in the UI to add the data, I just wonder whether adding another scenario and adding data requests manually (which is already covered in a feature spec) might be a little heavy handed in this case, when all we want to do is verify that the total calculation of that one line is correct. Ideally the place to test that would be in a view spec, but those can be fiddly and because you need to supply all the state they need to work, they're a bit artificial. But for a calculation I'd probably do it.  

For this tiny feature I think it's fine and pragmatic to use the existing feature spec, but maybe consider creating test data in the state you want and just verify the page displays with the total correct:

```suggestion

  context 'when multiple data requests are present and have pages logged' do
    given!(:data_request) { create(:data_request, offender_sar_case: offender_sar_case, cached_num_pages: 32) }
    given!(:second_data_request) { create(:data_request, offender_sar_case: offender_sar_case, cached_num_pages: 32) }

    scenario 'the total row displays with the correct total pages' do
      cases_show_page.load(id: offender_sar_case.id)
      expect(cases_show_page).to be_displayed
      expect(cases_show_page.data_requests.rows.size).to eq 3 # Unchanged num DataRequest

      last_row = cases_show_page.data_requests.rows.last
      expect(last_row).to have_selector '.total-label'
      expect(last_row.total_label).to have_text 'Total'
      expect(last_row).to have_selector '.total-value'
      expect(last_row.total_value).to have_text '64'
    end
  end
```

njpreddy: I've also moved this test to the log_data_recieved_spec.rb file.  Think thats the better context for it.

spec/features/cases/offender_sar/add_data_requests_spec.rb
njpreddy: Removed in favour of your more concise test. :)

Polsonby (approved):
Great work 👍

PR #1162, opened 2020-06-17
Add more options for reindex command

ymao2:
## Description
Added extra options for cases-reindex command

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2900

run command with following examples 
./cts cases reindex
./cts cases reindex --non-indexed --size=500

### DIFF
--- a/lib/cts/cases/cli.rb
+++ b/lib/cts/cases/cli.rb
@@ -270,6 +270,10 @@ module CTS::Cases
       end
     end
 
+    option :non_indexed, type: :boolean,
+           desc: 'Narrow down the scope of reindex to the cases which have not been indexed yet.'
+    option :size, aliases: 's', type: :numeric, default: 5000,
+            desc: 'Number of cases to for indexes. [5000]'
     desc 'reindex', 'Reindex all the cases'
     def reindex(*_args)
       reindexer = CTS::Cases::Reindex.new(CTS, options)
--- a/lib/cts/cases/cli.rb
+++ b/lib/cts/cases/cli.rb
@@ -273,8 +273,9 @@ module CTS::Cases
     option :non_indexed, type: :boolean,
            desc: 'Narrow down the scope of reindex to the cases which have not been indexed yet.'
     option :size, aliases: 's', type: :numeric, default: 5000,
-            desc: 'Number of cases to for indexes. [5000]'
-    desc 'reindex', 'Reindex all the cases'
+            desc: 'The limit for the number of cases to be indexed. [5000]'
+    desc 'reindex OPTION --non-indexed --size=<size>', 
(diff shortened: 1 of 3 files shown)


lib/cts/cases/cli.rb
```diff
@@ -270,6 +270,10 @@ def show(*args)
       end
     end
 
+    option :non_indexed, type: :boolean,
+           desc: 'Narrow down the scope of reindex to the cases which have not been indexed yet.'
+    option :size, aliases: 's', type: :numeric, default: 5000,
+            desc: 'Number of cases to for indexes. [5000]'
```
Polsonby, line 276: I think there's a word missing, not sure what - limit or something? `Number of cases to for `

lib/cts/cases/reindex.rb
```diff
@@ -11,8 +11,26 @@ def initialize(logger, options = {})
...
       logger.info "\n#{@count} cases reindexed"
     end
+
+    private 
+
+    def process_indexes
+      if @options[:non_indexed]
+        size = @options.fetch(:size)
+        logger.info "Start to query the cases with size #{size}."
+        cases = Case::Base.where("document_tsvector is NULL").limit(size)
```
Polsonby, line 24: So the idea is to pass the `reindex` task two options or none, is that right? Just wondering what happens if the `--size` option is omitted or you pass a string to it?

Should the method exit if you don't pass the size to it?

Polsonby (commented):
I think this is fantastic work, @ymao2 :) Couple of questions - the only real concern I have is where to document this - I can't see it outputting the description for these options in the ./cts help output so wondering if we could add something there

lib/cts/cases/reindex.rb
Polsonby: Interesting contrast between `@options[:non_indexed]` and `@options.fetch(:size)` - wondering from interest why use both syntaxes? They seem to behave slightly differently in edge cases e.g.  https://gist.github.com/stevenharman/5664318

ymao2: hmm,  didn't think about it too much, somehow I am sort of thinking the "fetch" similar as the "get" in python dict operator which will return nil even the key is not included, and [key] will raise a exception if key is not included, but after I checked the link, it is not what I thought,  and for this case,  2 doesn't make different, in that case let me change to [key]

lib/cts/cases/cli.rb
ymao2: changed the wording

Polsonby (approved):
Fab - love it! :)

PR #1170, opened 2020-06-27
CT-2911 added extra option for updating the warehouse table  trunk by trunk

ymao2:
## Description

The command was designed for small amount of cases update, doesn't work on huge volume of updates which is the case for data-migration for offender-sar case. this option allow us to update the warehouse table trunk by trunk

## Self-review checklist

Tested the report on development, there is no duplication there 
* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2911

Warehouse command. e.g.
./cts cases warehouse --size=100


app/models/warehouse/case_report.rb
```diff
@@ -94,18 +94,26 @@ def generate_all
...
+          query =
+            Case::Base
+              .joins('LEFT OUTER JOIN warehouse_case_reports ON warehouse_case_reports.case_id = cases.id')
+              .where("cases.deleted = 'false' AND warehouse_case_reports.case_id IS NULL")
+        else
+          query =
+            Case::Base
+              .joins('LEFT OUTER JOIN warehouse_case_reports ON warehouse_case_reports.case_id = cases.id')
+              .where("cases.deleted = 'false' AND warehouse_case_reports.case_id IS NULL")
+              .limit(job_size)
```
Polsonby, line 115: I think having the main body of the query duplicated might be hard to maintain - it's the same but people might think it's meant to be different. Could we collapse the conditional and just call limit optionally if job_size is not nil?

Something like 

```suggestion
          query = Case::Base
              .joins('LEFT OUTER JOIN warehouse_case_reports ON warehouse_case_reports.case_id = cases.id')
              .where("cases.deleted = 'false' AND warehouse_case_reports.case_id IS NULL")
           query = query.limit(job_size) if job_size
```

spec/lib/cts/cases/cli_spec.rb
```diff
@@ -28,7 +29,12 @@
         size: 2
       }
     )
-    find_or_create :team_dacu
+    allow(cli_warehouse).to receive(:options).and_return(
+      {
+        size: 2
+      }
```
Polsonby, line 35: Tiny thing but do we need all the vertical space here? 

```suggestion
      { size: 2 }
```

Polsonby (commented):
Couple of wee things to consider, maybe? Looks great otherwise 👍

app/models/warehouse/case_report.rb
ymao2: @Polsonby , I did this but somehow I got impression it will return the whole set first (which is the issue I am trying to avoid) then will apply limit afterwards,  let me try it again

willmcb:
lgtm 👍

app/models/warehouse/case_report.rb
Polsonby: Yeah, it's hard to prove - the idea is that these _should_ create ActiveRecord Relation objects that don't hit the DB until the last possible moment, precisely so you can modify them iteratively. But it seems that in the console when you try it you always see it hit the DB because of the way the console wraps up the method calls in order to display the output according to "some guy on StackOverflow". Did you manage to verify the behaviour one way or the other?

Polsonby (approved):
Thanks for making those changes - great work on this

app/models/warehouse/case_report.rb
ymao2: @Polsonby , tried both,  doesn't make any much different based on the figure from the measurement -_-! , this is why I made those changes based on your feedback,  I am only trusting the Ruby on rails should have considered it already,

PR #1172, opened 2020-06-29
CT-2737, CT-2774 for the closed reports bugs

ymao2:
## Description

This change is for fixing the issue for closed report, the issue was that once the time frame for generating the closed report become long like 1 year, there will be some closed cases missing from the report and at the same time some of closed cases will be duplicated twice or more, which in odd way to make up the total number of closed cases was correct 

After analysing the report and the SQL it uses, it turns out that the cases order within a sub-sql (the main for cases table) are different from the order in the warehouse table, so when limit + offset is applied to the final result,  the set of records get calculated by order were wrong.  so the fix is to move the limit + offset into the sub-sql which is the main for cases table to avoid such confusion. 

## Self-review checklist

* [XX] (1) Quick stakeholder demo done OR
I tested the result on local and development, both works fine
* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2737
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2774

Closed report
1 - Choose "report"
2 - Click "Create report"
3 - put one year range from 01/01/2018 to 31/12/2018
4 - download the report
5 - no missing case and no duplication


Polsonby (approved):
Awesome job :) 👍

PR #1174, opened 2020-06-30
Adds CRN number to Offender SAR cases

willmcb:
## Description
Adds CRN number to Offender SAR cases. This is so users can store the relevant identifying information about a data subject, and this data can be searched on later

### Questions to guide a reviewer:
- I have added a test for the details page explicitly. It seems that there are few explicit tests for the `offender_sar/new` page. Tests for create use page_objects which build using the factories. I have updated the factory to include the new field and things seem to pass. I would like your opinion on whether this is the right approach here and whether I can do more on testing the input side explicitly.
- I noticed that there is a `searchable_fields_and_ranks` method - which I am assuming sets out a priority map for data points that will be returned on a search by a user. I haven't added the CRN number to this yet, as I am still not fully aware of the context, but would this be something that I should do? (either here or in a separate PR).

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
**Before:**

Screenshot 2020-06-30 at 17 26 55

**After:**

Screenshot 2020-06-30 at 17 28 12

*case details:*
Screenshot 2020-06-30 at 17 37 20

### Related JIRA tickets
[CT-2892](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2892)

### Deployment
- no new migrations as this adds to the JSONB field in the Case table.

1. Login as user with Applications privileges i.e. Brian Rix
2. Click create a case
3. Add relevant details (including CRN)
4. Click through the Wizard
5. You should see the CRN number appear on the case details page at the end of the flow.


config/locales/en.yml
```diff
@@ -565,6 +566,7 @@ en:
       late_team: Business unit responsible for late response
       name: Requester
       other_subject_ids: "Police national computer number"
+      criminal_record_reference_number: "Criminal records reference number"
```
Polsonby, line 569: Maybe a typo?

```suggestion
      criminal_record_reference_number: "Criminal record reference number"
```

Polsonby (commented):
Nice work, looks good to me! Happy to approve - maybe demo quickly to Alex and we can merge?

Polsonby:
Thanks for the thoughtful approach you've taken here :) 

> I would like your opinion on whether this is the right approach here and whether I can do more on testing the input side explicitly

It's really interesting returning to this with fresh eyes, more experience and a better understanding of the app. I'd love us to improve the testing of the case creation flow, using the approach that Misha built for the other correspondence types which I didn't understand properly at the time. There's a potential excuse that this has been very much an exploration for the team into the business problem, so maybe being a little lighter on the tests was appropriate until we knew what we needed from the app, but really it was my own lack of understanding. 

If you agree it would be valuable to extend the feature spec coverage to support the variations in case creation, I wouldn't hold this PR up for that as long as we can add it this sprint as part of the big overhaul we're doing - seems like the right time to do it. WDYT?

Polsonby:
> I noticed that there is a searchable_fields_and_ranks method - which I am assuming sets out a priority map for data points that will be returned on a search by a user. I haven't added the CRN number to this yet, as I am still not fully aware of the context, but would this be something that I should do? (either here or in a separate PR).

Excellent spot - this is something I'd meant to come back to but hadn't. We should make a ticket or add some details if there's an existing one for the Offender SAR search - could you pick this up with Alex?

config/locales/en.yml
willmcb: It was 'record' on the prototype so I left it as that: https://cts-prototype.herokuapp.com/create-case1#width-hint-hint

Happy to change though if it should be 'records' 👍

willmcb: It was 'record' on the prototype so I left it as that: https://cts-prototype.herokuapp.com/create-case1

Polsonby: Other way round :) - it's records presently, looks like it should be record

Polsonby (approved):
LGTM :)

config/locales/en.yml
willmcb: So after a few conversations with Alex it turns out it should be Case reference number - so I have changed the PR to reflect this 👍

Polsonby (approved):
Brilliant, not that it matters much but the change to "case_reference_number" reads much nicer in the code too for some reason 👍

PR #1183, opened 2020-07-03
CT-2928 add new fields for contact details

Polsonby:
## Description
This was going to include the new subject_address and third_party_name fields but these are handled in other PRs. This PR includes the addition of the `recipient` enum field, renames the `third_party_reference` to `requester_reference` and adds convenience methods to make it easier to get recipient and requester details for subsequent use. 

The values for the enum are [appended _recipient](https://github.com/ministryofjustice/correspondence_tool_staff/pull/1183/files#diff-3951f75259bad9bc95efc4924248544bR65-R70) because the model gets automatically created convenience helper methods including a predicate method. When we originally proposed the values, these included "third_party" which generated a convenience helper method `third_party?` which overwrote an existing predicate method and let to some confusing behaviour. So we've namespaced the values as described to prevent this collision. Now the new predicate method from the enum is `kase.third_party_recipient?`

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
New cases require a valid value from the enumerated options for `recipient` - in order to pass validation there is a new text field with a default value on the recipient_details screen pending the next PR.

image


app/views/cases/offender_sar/_case_details.html.slim
```diff
@@ -22,7 +22,7 @@
           = render partial: 'cases/sar/third_party_relationship', locals: { case_details: case_details }
           = render partial: 'cases/shared/requester_name', locals: { case_details: case_details }
           = render partial: 'cases/offender_sar/third_party_company_name', locals: { case_details: case_details }
-          = render partial: 'cases/offender_sar/third_party_reference', locals: { case_details: case_details }
+          = render partial: 'cases/offender_sar/requester_reference', locals: { case_details: case_details }
```
willmcb, line 25: Is there anything in the spec for the `_case_details_spec` that needs updating if we add new values here?

willmcb: Just thinking As this are optional fields should they shoe if they are empty?

spec/models/case/sar/offender_spec.rb
```diff
@@ -254,6 +280,81 @@
...
+      kase = create :offender_sar_case
+      expect(kase.subject_name).to eq kase.subject_full_name
+      kase.update_attribute(:subject_full_name, "Bob Hope")
+      expect(kase.subject_name).to eq 'Bob Hope'
+    end
+  end
+
+  describe '#subject_address' do
+    it 'returns a dummy string for now' do
+      pending "Will's PR"
```
willmcb, line 294: Just thinking it might be good to quote the ticket number or PR here - just incase someone else ends up picking this up.

spec/models/case/sar/offender_spec.rb
```diff
@@ -254,6 +280,81 @@
...
+      expect(kase.third_party_address).to eq kase.postal_address
+      kase.update_attribute(:postal_address, '11 The Road')
+      expect(kase.third_party_address).to eq '11 The Road'
+    end
+  end
+
+  describe '#requester_name' do
+    it 'returns third_party_name if third_party request' do
+      kase = create :offender_sar_case, :third_party
+      expect(kase.requester_name).to eq kase.third_party_name
```
willmcb, line 312: I'm not sure of the implementation here but, it is possible expectations on literal values might be better. As this is testing that the value of one method is eq to the value of another method on the same class. I think this can create issues in that what if both methods are "wrong" but are eq. Using literals means that we can be sure that `method_a` equals a known value without the proxy of the class.

willmcb (approved):
Approving - some non-blocking suggestions in comments though 👍

app/views/cases/offender_sar/_case_details.html.slim
Polsonby: Excellent points, thanks! I've been assuming we'd review the view spec when we do the ticket for the show page, but maybe that's the wrong way to do it?

On the showing of empty fields, I asked Alex and this is by intent

spec/models/case/sar/offender_spec.rb
Polsonby: Thanks, I was being lazy! Will amend

PR #1185, opened 2020-07-07
CT-2364 offender sar monthly performance report

ymao2:
## Description

The purpose of those changes and this ticket is to make the performance report for offender-sar monthly report producible after data migration 
The major changes I made ( please don't be scared by the number of files I changes :)).
1. Changed the method of limiting the scope of the cases a user can see for all different sar-type cases to avoid loading of whole cases and also consolidate the scope-limitation from finder service to policy for foi cases
   The structure of cases scope of policy will follow 
     - Correspondence Type
        - Manager 
        - Responder
           - within the teams being assigned to the case
        - Approver
           - within the teams being assigned to the case

2. Added another way of generating monthly-performance-report based on the size of cases within the scope, if the size beyond the limit, it will be split into small tasks

3. Changed some parts related to general report
   a. In order to tracking the report status easily and in a explicit way when the report is generated via jobs, I introduced a new properties field for report,  
   b. introduce report_format rather than several flags like csv?, xlsx?, zip? etc. 
   c. re-scope of the data which is stored in report.report_data field to the json stats data only, no any final report format will be involved,  make the controller to generate the final format based on the report_form of each report service. 
   
  The structure of reports 
   Report Type --- define the basic feature for a report type. each type has an report service accordingly
       Report service --- The actual class for handling how a report is produced based on the time frame and the scope of cases
  Report ---  the report being generated from report service. If the report is required to be saved or it is produced via etl process, this report will be saved in report table. all the final features for a specific report like final name, time frame, status, report format and the report data(only the json status data regardless which format will be used for final outcome) will be stored.
  stats_controller --- will handle the UI flow and convert json stats data into the format (xlsx, csv) or no converting for zip data accordingly 
   
4. a bug for filtering the scope of performance report to change the date filter from created_at to received_date

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2364

Generate the monthly-performance-report for offender-sar cases
- login in as Branston 
- Click the reports
- Click the monthly report for Offender sar cases, 2 possible ways depending on how big the cases being involved
   a. you can download the report directly 
   or 
   b. you will be redirected into another page and there is downlaod link which will give you the chance to download the report


app/controllers/stats_controller.rb
```diff
@@ -122,9 +150,10 @@ def create_spreadsheet(report_data)
   def set_fields_for_custom_action
     @custom_reports_foi = ReportType.custom.foi
     @custom_reports_sar = ReportType.custom.sar
+    @custom_reports_offender_sar = ReportType.custom.offender_sar
     @custom_reports_closed_cases = ReportType.closed_cases_report
-    @correspondence_types = CorrespondenceType.custom_reporting_types +
-      [self.class.closed_cases_correspondence_type]
+    @correspondence_types = CorrespondenceType.custom_reporting_types & current_user.permitted_correspondence_types
```
Polsonby, line 155: Cool! So this is using the "Set Intersection" operator from ArrayClass? Just checking because there are a few uses of unary ampersand in Ruby and I think most people (myself included) find it hard to keep them straight. 
https://ablogaboutcode.com/2012/01/04/the-ampersand-operator-in-ruby

Polsonby: I wonder if this line is worth pulling out into a method to help explain what's going on here

ymao2: @Polsonby  could do,  basically the user can only generate the custom report within his/her permitted scope

app/controllers/stats_controller.rb
```diff
@@ -14,19 +16,14 @@ def index
...
-      report_data = report.run
-
-      axlsx = create_spreadsheet(report_data)
-
-      send_data axlsx.to_stream.read,
-                filename: report.report_type.filename('xlsx'),
-                disposition: :attachment,
-                type: SPREADSHEET_CONTENT_TYPE
+    report.run_and_update!(user: current_user)
+    if report.etl?
```
willmcb, line 20: I think we may of spoken about this yesterday - just wondering if `etl?` is the best method name here when I think it is really checking that the report is `processedAsBackgroundJob?`

app/controllers/stats_controller.rb
```diff
@@ -57,23 +61,19 @@ def create
 
   def download_custom
     report = Report.find(params[:id])
-    data, filename = report.report_details
-
-    if report.report_type.etl?
+    report_data, _ = report.report_details
+    if report.etl?
       return download_waiting(report) unless report.etl_ready?
```
willmcb, line 66: if we change `.etl?` we may also want to change `.etl_ready?` to something else. I think just `.ready?` would be fine as the fact that it it was done in the background isn't relevant really.

app/controllers/stats_controller.rb
```diff
@@ -82,6 +82,34 @@ def self.closed_cases_correspondence_type
 
   private
 
+  def generate_final_report(report, report_data=nil)
```
willmcb, line 85: This could possibly be a case statement with smaller methods for each case - might make it easier to parse, I'm thinking something like:

```ruby
def generate_final_report(report, report_data = nil)
    case report.report_format
    when 'xlsx'
       send_xlsx_report(report)
    when 'csv'
      send_csv_report(report)
    else
      send_default_report(report)
    end
end
```
not sure about the method names, but thinking that something like:

`genereate_xlsx_report`, `format_xlsx_report` or `send_xlsx_report` might be candidates.

app/controllers/stats_controller.rb
```diff
@@ -100,10 +128,10 @@ def self.closed_cases_correspondence_type
   # Assumes no report spans more than 26 columns
   SPREADSHEET_COLUMN_NAMES = ('A'..'Z').to_a
 
-  def create_spreadsheet(report_data)
+  def create_spreadsheet(report)
     axlsx = Axlsx::Package.new
     axlsx.workbook.add_worksheet do |sheet|
-      report_data.to_csv.each_with_index do |row, row_index|
+      report.to_csv.each_with_index do |row, row_index|
```
willmcb, line 134: 👍

app/controllers/stats_controller.rb
```diff
@@ -1,3 +1,5 @@
+require 'csv'
+
 class StatsController < ApplicationController
   # @note (Mohammed Seedat): Interim solution to allow 'Closed Cases'
```
willmcb, line 4: General comment about this controller is that it is getting quite large - I think there may be an argument to split things out into service classes that could be more easily unit tested. I think we could discuss it as something to do in a separate PR though.

app/jobs/performance_report_job.rb
```diff
@@ -0,0 +1,18 @@
+class PerformanceReportJob < ApplicationJob
+
+    queue_as :performance_report
+  
+    def perform(report_service_string, report_job_guid, period_start_ts, period_end_ts, offset)
+        RavenContextProvider.set_context
+        begin
+            report_service_class = report_service_string.constantize
```
willmcb, line 8: I was a bit confused here as we are getting a class from a `String`. I think the word string is misleading in a way as it is the datatype rather than the attribute of the argument passed in. A clearer argument name might be `report_service_name` or `report_service_class_name`. 

This way it is clearer when you call `.constantize` that you want to get the class name and then instantiate it.

Maybe:

```ruby
report_service_class_name = report_service_class_name.constantize
report_service = report_service_class_name.new()
# etc...
```

app/models/case/base.rb
```diff
@@ -342,6 +344,7 @@ def create!(_attributes = {})
 
   after_initialize do
     self.workflow = default_workflow if self.workflow.nil?
+    @deadline_calculator = self.class.respond_to?(:type_abbreviation) ? create_deadline_calculator : nil
```
willmcb, line 347: Not sure what this is doing - may be good to wrap it in a descriptive method possibly.

app/models/case/base.rb
```diff
@@ -554,7 +557,7 @@ def responded_late?
 
   def responded_in_time?
     return false unless closed_for_reporting_purposes?
-    date_responded <= external_deadline
+    date_responded.nil? ? false : date_responded <= external_deadline
```
willmcb, line 560: Its rare I think to have expressions in ternary operators - I think it may be clearer to just add another guard clause:

```ruby
return false if date_responded.nil?
date_responded <= external_deadline
```

app/models/report.rb
```diff
@@ -9,24 +9,33 @@
...
-# Because run_and_update! needs to generate CSV
-require 'csv'
-
 class Report < ApplicationRecord
-  # Status names for ETL generated reports
-  COMPLETE = 'complete'.freeze
-  WAITING = 'waiting'.freeze
 
+  jsonb_accessor :properties,
+                 etl: :boolean,
```
willmcb, line 19: `etl` might need to change here as well if we decide to change it

app/policies/case/foi/standard_policy.rb
```diff
@@ -8,7 +8,19 @@ def initialize(user, scope)
 
     def resolve
       if @user.permitted_correspondence_types.include? CorrespondenceType.foi
-        @scope
+        
+        if @user.responder_only?
+          team_restriction_sql = "select case_id from assignments
```
willmcb, line 13: Might be clearer to change the case of the sql here so that it is clearer when parsing the code i.e. `select` => `SELECT` and `join` => `JOIN` etc.

app/policies/case/foi/standard_policy.rb
```diff
@@ -8,7 +8,19 @@ def initialize(user, scope)
...
+        if @user.responder_only?
+          team_restriction_sql = "select case_id from assignments 
+          join teams_users_roles 
+          on assignments.team_id = teams_users_roles.team_id
+          where teams_users_roles.role = '%<role>s' 
+          and teams_users_roles.user_id = %<user_id>i"
+
+          @scope.where(" cases.id in 
+          (#{team_restriction_sql}) " % {role: :responder.to_s, user_id: @user.id})
+        else
```
willmcb, line 21: This still makes me nervous - I think there is probably a way to have `ActiveRecord` construct this query.

See: http://joshfrankel.me/blog/constructing-a-sql-select-from-subquery-in-activerecord/

app/policies/case/sar_policy.rb
```diff
@@ -7,34 +7,42 @@ def initialize(user, scope)
...
 
-        # Unfortunately neither of the scopes suggested here work with with the 'or' clause
-        # as they join across tables. It's not obvious whether a user can be both
-        # a responder and an approver which would result in an 'or' join here - but
-        # the higher level tries to do an 'or' join which rules this out anyway.
         if @user.responder?
-          case_ids = Assignment.with_teams(@user.responding_teams).pluck(:case_id)
-          scopes << @scope.where(id: case_ids)
+          scopes << @scope.where(" cases.id in 
+          (#{team_restriction_sql}) " % {role: :responder.to_s, user_id: @user.id})
```
willmcb, line 29: I think here again we should try to find a way to make this more ActiveRecord friendly rather than rely on raw SQL

app/policies/case/sar_policy.rb
```diff
@@ -7,34 +7,42 @@ def initialize(user, scope)
...
         end
 
         if @user.approver?
-          case_ids = Assignment.with_teams(@user.approving_team).pluck(:case_id)
-          scopes << @scope.where(id: case_ids)
+          scopes << @scope.where(" cases.id in 
+          (#{team_restriction_sql}) " % {role: :approver.to_s, user_id: @user.id})
```
willmcb, line 34: Same here

app/services/stats/base_monthly_performance_report.rb
```diff
@@ -37,13 +39,45 @@ def self.description
       raise '#description should be defined in sub-class of BaseMonthlyPerformanceReport'
     end
 
-    def run(*)
+    def process(offset, report_job_guid=nil)
       CaseSelector.new(case_scope)
```
willmcb, line 43: Nicer name here 👍

app/services/stats/base_monthly_performance_report.rb
```diff
@@ -37,13 +39,45 @@ def self.description
...
+      if !report_job_guid.nil?
+        redis = Redis.new
+        redis.set(report_job_guid, @stats.stats.to_json)
+      end
 
-      @stats.finalise
+    end
+
+    def run(*)
+      offset = 0
```
willmcb, line 58: This is quite a long method - might be worth having some private methods with some descriptive names that make it clear the steps that are happening here - something like:

```ruby
if data_size > ROWS_PER_FRAGMENT
    batch_up_records()
end
```
```

File: app/services/stats/base_monthly_performance_report.rb [hunk 17]
```diff
@@ -64,7 +98,40 @@ def to_csv
...
+    # This function is only when the report is done via ETL (tasks)
+    def report_details(report)
+      redis = Redis.new
+      data_collector = []
+      report.job_ids.each do |job_id|
+        if redis.exists(job_id)
+          data_collector << redis.get(job_id)
+        end
+      end
+      if data_collector.count == report.job_ids.count
```
willmcb, line 114: I think this method also might benefit from being split up a bit into smaller functions

app/services/stats/calculations.rb
```diff
@@ -78,5 +78,58 @@ def self.calculate_percentage(value, total)
...
+          bu_results.each do |key, value|
+            directorate_results[key] += value
+            business_group_results[key] += value
+            overall_total_results[key] += value
+          end
+        end
+      end
+
+      def self.populate_team_details_callback(stats)
+        # This is a hash (indexed by team id) containing a hash of team stats columns
```
willmcb, line 98: I think this could also possibly be split up - the branches of the case statement could be methods or the case statement could be extracted 🤔 

I'm also think the 'totals' section at the bottom could be its own method 👍

app/services/stats/stats_collector.rb
```diff
@@ -75,12 +75,13 @@ def each
         @superheadings.each { |superheading| yield superheading }
         yield cols
         row_names.each do |row_name|
-          if row_name =~ /^_SPACER/
+          row_name_str = row_name.to_s
```
willmcb, line 78: I think this may be clearer as a case statement.

db/data_migrations/20200710233929_update_offender_sar_as_custom_report.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+    end
+    ReportType.offender_sar.each do |report_type|
+      report_type.update_attributes(
+        custom_report: false,
+        offender_sar: false
+      )
+    end
+  end
+end
```
willmcb, line 34: Newline :)

spec/jobs/performance_report_job_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'rails_helper'
+
+
+class DummyWrongTestFirst
```
willmcb, line 4: I like this method of creating Dummy classes - nice and simple way to do things 👍  - one small thing is indentation is 4 spaces here I usually in ruby its 2.

willmcb (changes requested):
Good job! Loads of really great stuff here - I have made a few comments in line it would be good to go through them with you and @Polsonby at some point soon 👍

spec/services/stats/base_monthly_performance_report_spec.rb
```diff
@@ -0,0 +1,127 @@
...
+                expect(@report_another.num_fragments).to eq 4
+                expect(@report_another.data_size).to eq 4
+                expect(@report_another.etl).to eq true
+                expect(@report_another.status).to eq Stats::BaseReport::WAITING
+                expect(@report_another.persist_results?).to eq true
+            end
+        end
+    end  
+  end
+end
```
willmcb, line 127: Newline needed here as well :)

app/services/stats/calculations.rb
ymao2: @willmcb  as I said, this part was not written by me and I didn't change it, so I would suggest to leave it as it is

app/models/case/sar/offender.rb
```diff
@@ -47,7 +47,6 @@ def searchable_fields_and_ranks
                  subject_type: :string,
                  third_party_relationship: :string,
                  third_party: :boolean,
-                 requester_reference: :string,
```
willmcb, line 50: Why has this been removed?

ymao2: @willmcb  sorry for confusion!, requester_reference has been duplicated (at line 44)  from master branch,  I always saw this warning message from my console, so.... I removed

app/models/report.rb
Polsonby: @ymao2 @willmcb 

Sorry, I'm now confused! Is this meant to mean background or warehouse? Should we change it?

ymao2: @Polsonby , it is warehouse thing,  there is another flag background_job is beneath it,  this flag is defined from report_type table, storing it in the report again is kind of duplication, let me remove it

app/models/report.rb
```diff
@@ -45,65 +55,64 @@ def run(report_guid: SecureRandom.uuid, **args)
...
+  def persist_results?
+    self.persist_results
   end
 
   # ETL (Extract Transform Load) based reports are generated using
   # the Warehouse, and therefore require processing before making
   # available for download
   def etl?
-    self.report_type&.etl?
+    self.etl
```
Polsonby: I think this could be an issue - should be left as-is?

Polsonby (approved):
I'm happy to approve - let's get it merged and QA it with real data 👍

Polsonby:
Great work @ymao2 and great reviewing @willmcb :)

willmcb (approved):
LGTM 👍

PR #1191, opened 2020-07-09
CT-2916 Add requester_reference

willmcb:
## Description

Adds requester_reference, and request_dated fields to DB.

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
**Before:**
Screenshot 2020-07-13 at 16 22 40

**After:**
Screenshot 2020-07-13 at 16 20 55

case_details
Screenshot 2020-07-13 at 16 03 05

### Related JIRA tickets
[CT-2916](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2916)

1) Login as Brian rix
2) fill in details
3) when at "What is the date on the request?" question fill in details - try a date in the future this should error.
4) click through to case_details references should be visible as above
5) click 'change' edit the details for the new fields
6) check that updates are reflected in case_details


:w
```diff
@@ -0,0 +1,139 @@
...
+      create :case_transition_waiting_for_data, case: kase
+      create :case_transition_ready_for_vetting, case: kase
+      create :case_transition_vetting_in_progress, case: kase
+      create :case_transition_ready_to_copy, case: kase
+      create :case_transition_ready_to_dispatch, case: kase
+      create :case_transition_close, case: kase
+      kase.reload
+    end
+  end
+end
```
Polsonby, line 139: I think you might have created this file accidentally on the way past... `:w` is a very vim-sounding filename :)

app/views/cases/offender_sar/_case_details.html.slim
```diff
@@ -23,10 +23,9 @@
           = render partial: 'cases/sar/third_party_relationship', locals: { case_details: case_details }
           = render partial: 'cases/offender_sar/requester_name', locals: { case_details: case_details }
           = render partial: 'cases/offender_sar/third_party_company_name', locals: { case_details: case_details }
-          = render partial: 'cases/offender_sar/requester_reference', locals: { case_details: case_details }
         = render partial: 'cases/sar/response_address', locals: { case_details: case_details }
-        = render partial: 'cases/offender_sar/recipient_details', locals: { case_details: case_details }
```
Polsonby, line 28: Did we mean to remove this?

spec/factories/case/offender_sars.rb
```diff
@@ -30,6 +30,8 @@
     other_subject_ids               { 'ABC 123 DEF' }
     case_reference_number           { '123 ABC 456' }
     subject_address                 { '22 Sample Address, Test Lane, Testingington, TE57ST' }
+    request_dated { Date.parse('13-07-2010') }
+    requester_reference { '456 ABC 123' }
```
Polsonby, line 34: I wouldn't hold up anything on account of it, but FWIW...

I don't normally like this approach (like on lines 30-32) aligning things with spaces in vertical rows as if they're tabbed columns  - wouldn't do it myself - but if they're lined up when we encounter the file, I'd maintain the arrangement

Wouldn't recommend changing the other ones either, to preserve easy inspection of line history with blame

spec/models/case/sar/offender_spec.rb
```diff
@@ -226,6 +226,24 @@
...
+  describe '#request_dated' do
+    context 'invalid value' do
+      it 'errors' do
+        expect {
+          build(:offender_sar_case, request_dated: 'wibble')
+        }.to raise_error ArgumentError
+      end
+    end
+
+    context 'received date cannot be in future' do
```
Polsonby, line 239:
```suggestion
    context 'request_dated cannot be in future' do
```
maybe worth amending the label here

Polsonby, line 239:
```suggestion
    context 'request_dated cannot be in future' do
```
maybe worth amending the label here

Polsonby (approved):
Good stuff! Couple tiny things but approving

:w
willmcb: lol thanks vim

app/views/cases/offender_sar/_case_details.html.slim
willmcb: I think so - as it has moved from third party recipient dropdown to a new screen it was on the case_details twice if the sending to a third party

app/models/case/sar/offender.rb
```diff
@@ -40,6 +41,8 @@ def searchable_fields_and_ranks
                  recipient: :string,
                  reply_method: :string,
                  subject_address: :string,
+                 request_dated: :date,
+                 requester_reference: :string,
```
Polsonby, line 45: Apologies again! We have duplicate key in here for `requester_reference` - duplicated on line 51 below. Puma complains about it on server startup. Maybe we could order the keys alphabetically again - seems to be have got out of sync a little

PR #1195, opened 2020-07-14
CT-2919 update letter templates

Polsonby:
## Description
The reason for doing all this was so that we could support cases where the requester and the recipient were different people and we needed to send the acknowledgement letter to one and the dispatch letter to the other. 

This PR adds support for `letter_recipient` which is a convenience method on the Letter object which calls either "requester_name" or "recipient_name" depending on whether it is an acknowledgement letter or a dispatch letter, and the same for `letter_address`. This allows us to keep logic out of the letter templates themselves and have simple method calls in the ERB templates for those.

There was a bug where it displayed the offender name for third party recipient, when it should have output either the representative name or nothing. Think this is fixed, but worth keeping an eye on

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots

For example, prisoner requests own data to send to solicitor

Acknowledgement letter addressed to prisoner
image

Dispatch letter addressed to solicitor
image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2919

### Deployment
nope

For each scenario, you need to make sure you choose the relevant template at the letter stage - solicirtor or offender template for example - but otherwise it should "do the right thing" in terms of putting in subject name/address or third party name/address as appropriate depending on subject/requester/recipient settings for the case

### DIFF
--- a/app/models/case/sar/offender.rb
+++ b/app/models/case/sar/offender.rb
@@ -217,7 +217,8 @@ class Case::SAR::Offender < Case::Base
   end
 
   def recipient_name
-    (!subject_recipient?) ? third_party_name : subject_name
+    return subject_name if subject_recipient?
+    third_party_name.present? ? third_party_name : ''
   end
 
   def recipient_address
--- a/app/models/letter.rb
+++ b/app/models/letter.rb
@@ -29,5 +29,23 @@ class Letter
   def template_name
     @letter_template&.name
   end
+
+  def letter_recipient
+    case @letter_template.template_type
+    when "dispatch"
+      values.recipient_name
+    when "acknowledgement"
+      values.requester_name
+    end
+  end
+
+  def letter_address
+    case @letter_template.template_type
+    when "dispatch"
+      values.recipient_address
+    when "acknowledgement"
+      values.requester_address
+    end
+  end
 end
 
--- a/app/models/letter.rb
+++ b/app/models/letter.rb
@@ -30,7 +30,7 @@ class Letter
     @letter_template&.name
   end
 
-  def letter_recipient
+  def name
     case @letter_template.template_type
     when "dispatch"
       values.recipient_name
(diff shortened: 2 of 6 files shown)


app/models/letter.rb
```diff
@@ -29,5 +29,23 @@ def letter_date
   def template_name
     @letter_template&.name
   end
+
+  def letter_recipient
```
willmcb, line 33: There is a tiny bit of duplicated logic here - I think it's fine but wonder if we could avoid this with something like a LetterRecipient class.

I don't think this is a blocker just a thought as I saw similar shapes in the code.

willmcb: tbh though I think this implementation is fine and not overly complicated 👍

app/models/letter.rb
```diff
@@ -29,5 +29,23 @@ def letter_date
...
+    case @letter_template.template_type
+    when "dispatch"
+      values.recipient_name
+    when "acknowledgement"
+      values.requester_name
```
willmcb, line 38: This is a hard one for naming. From the perspective of the models you'd always think that `recipient_name` is was who the letter was going to - but that's not the case in reality.

Thinking whether there is a level of abstraction we are missing that could make this clearer.

app/models/case/sar/offender.rb
```diff
@@ -185,7 +185,8 @@ def requester_address
   end
 
   def recipient_name
-    (!subject_recipient?) ? third_party_name : subject_name
+    return subject_name if subject_recipient?
```
willmcb, line 188: Is there anyway that both of these could be `nil`? I'm thinking that this would not be allowed be the model validations.

willmcb (approved):

app/models/letter.rb
Polsonby: Agreed  - I think when we get the next type of letter (probably chasing) we could look at refactoring the letter stuff

app/models/letter.rb
Polsonby: I like the idea :)

PR #1204, opened 2020-07-16
CT-2926 no case edit history if no changes

ymao2:
## Description

Solve the issue for adding new entry to case history even there is nothing being changed
## Self-review checklist

* [X] (3) Tests passing

### Screenshots

before any change 
Screenshot 2020-07-16 at 13 40 23

Choose a section and click 'change', and then click "continue" without changing anything
Screenshot 2020-07-16 at 13 40 38
Choose same section, then change something and click "continue" 
Screenshot 2020-07-16 at 13 40 52
Screenshot 2020-07-16 at 13 40 56

### Related JIRA tickets

https://dsdmoj.atlassian.net/browse/CT-2926
### Deployment

This PR is to fix the issue of adding new entry into case history even you didn't change anything

login as branson office
choose a offender case 
click a section from case detail part e.g. subject detail
Click the 'continue' button without changing anything 
you should see 'No change were made' and no adding record in case history
Click the same section again and change some detail 
you should see 'case updated' and a new entry in the case history


app/form_models/offender_sar_case_form.rb
```diff
@@ -150,22 +67,14 @@ def set_empty_value_if_unset(params, field)
     params.merge!(field => "") unless params[field].present?
   end
 
+  def set_empty_value_if_unset_for_date(params, field)
+    params.merge!(field => "") unless params["#{field}_yyyy"].present? && 
+                                params["#{field}_mm"].present? && 
+                                params["#{field}_dd"].present?
+  end
```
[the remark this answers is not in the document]
ymao2: how did you know that? yes -_-!,  I was trying to get the date_of_birth to work, somehow in the session data I can only get the values of 'virtual fields'. not the real field, then after params_for_step() call, it will insert a empty date_of_birth which cause the validation failure :(.

app/controllers/cases_controller.rb
```diff
@@ -93,13 +93,15 @@ def edit
   def update
     @case = Case::Base.find(params[:id])
     authorize @case
+    @case = @case.decorate
+    if self.class.method_defined?(:persist_session_data)
+      self.persist_session_data(params)
+    end
```
Polsonby, line 99: I'd rather we weren't editing cases controller like this - wondering if we can have the update method in offender_sar_controller.rb call "persist_session_data" on something there then call `super`

app/controllers/cases_controller.rb
```diff
@@ -110,7 +112,6 @@ def update
     end
 
     set_permitted_events
-    @case = @case.decorate
     @case_transitions = @case.transitions.case_history.order(id: :desc).decorate
```
Polsonby, line 115: Sorry, not sure why we're changing these lines - 101, 102, 113?

spec/form_models/offender_sar_case_form_spec.rb
```diff
@@ -1,125 +0,0 @@
-require 'rails_helper'
-
-RSpec.describe OffenderSARCaseForm, type: :model do
-  let(:case_form) { OffenderSARCaseForm.new({}) }
-
-  it 'can be created' do
```
willmcb, line 6: I still think it would be good to keep this spec rather than delete it completely - even if it is just to keep `it can be created`.

There are still two public methods on the `OffenderSarCaseForm.rb` which I think could be tested.

A wider point is that it makes sense to keep this file even with minimal tests as it serves as a scaffold for future changes to `OffenderSarCaseForm.rb`. The fact that there is a `spec` file present makes people thinking that changes they make should have associated tests. If no test file exists there is an 'effort' barrier to creating one again.

Polsonby (approved):
Great work :) 👍

willmcb (commented):
Good job 👍

PR #1205, opened 2020-07-17
CT-2917 update the case details page

njpreddy:
## Description
CT-2917 Update the case details section

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
Case 1 - Third party requests data to be sent to the data subject

Case1

Case 2 - Third party requests data for themselves

Case2

Case 3 - Data subject requesting their own data

Case3

Case 4 - Data subject requesting data to be sent to third party

Case4

### Related JIRA tickets
Jira ticket CT-2917
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=CT&modal=detail&selectedIssue=CT-2917

Create the following cases with the appropriate steps:

87925944-ebd96280-ca78-11ea-877f-41163baaeca1

### DIFF
--- a/app/helpers/cases_helper.rb
+++ b/app/helpers/cases_helper.rb
@@ -331,11 +331,11 @@ module CasesHelper #rubocop:disable Metrics/ModuleLength
 
   def get_sar_recipient_label(recipient)
     if recipient == 'requester_recipient'
-      'Requester'
+      t('helpers.label.offender_sar.recipient_type.recipient')
     elsif recipient == 'subject_recipient'
-      'The data subject'
+      t('helpers.label.offender_sar.recipient_type.data_subject')
     else
-      'Third party'
+      t('helpers.label.offender_sar.recipient_type.third_party')
     end
   end
 
--- a/app/helpers/cases_helper.rb
+++ b/app/helpers/cases_helper.rb
@@ -328,4 +328,15 @@ module CasesHelper #rubocop:disable Metrics/ModuleLength
       kase.has_attribute?(:escalation_deadline) &&
       kase.within_escalation_deadline?
   end
+
+  def get_sar_recipient_label(recipient)
+    if recipient == 'requester_recipient'
+      'Requester'
+    elsif recipient == 'subject_recipient'
+      'The data subject'
+    else
+      'Third party'
+    end
+  end
+
 end
--- a/app/views/cases/offender_sar/_case_details.html.slim
+++ b/app/views/cases/offender_sar/_case_details.html.slim
@@ -14,12 +14,12 @@
 
         = render partial: 'cases/offender_sar/requested_on_behalf', locals: { case_details: case_details }
 
-        - if case_details.third_party? && (case_details.recipient == 'requester_recipient' || case_details.recipient == 'subject_recipient')
+        - if case_details.third_party?
           = render partial: 'cases/offender_sar/third_party_details', locals: { case_details: case_details }
 
         = render partial: 'cases/offender_sar/request_recipient', locals: { case_details: case_details }
 
-        - if (not case_details.third_party) && case_details.recipient == 'third_party_recipient'
+        - if case_details.third_party_recipient?
           = render partial: 'cases/offender_sar/third_party_details', locals: { case_details: case_details }
 
         = render partial: 'cases/offender_sar/request_details', locals: { case_details: case_details }
--- a/app/views/cases/offender_sar/_case_details.html.slim
+++ b/app/views/cases/offender_sar/_case_details.html.slim
@@ -7,7 +7,10 @@
 
     table
       tbody.sar-basic-details
+        / 1st Section
         = render partial: 'cases/shared/case_type', locals: { case_details: case_details }
+
+        / 2nd Section
         = render partial: 'cases/offender_sar/data_subject', locals: { case_details: case_details }
         = render partial: 'cases/offender_sar/subject_aliases', locals: { case_details: case_details }
         = render partial: 'cases/offender_sar/date_of_birth', locals: { case_details: case_details }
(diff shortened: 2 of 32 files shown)


app/helpers/cases_helper.rb
```diff
@@ -328,4 +328,15 @@ def show_escalation_deadline?(kase)
       kase.has_attribute?(:escalation_deadline) &&
       kase.within_escalation_deadline?
   end
+
+  def get_sar_recipient_label(recipient)
```
willmcb, line 332: Might be worth extracting these out into the `en.yml`

app/views/cases/offender_sar/_case_details.html.slim
```diff
@@ -19,15 +19,17 @@
         = render partial: 'cases/offender_sar/subject_address', locals: { case_details: case_details }
         = render partial: 'cases/offender_sar/high_profile_subject', locals: { case_details: case_details }
         = render partial: 'cases/offender_sar/third_party', locals: { case_details: case_details }
-        - if case_details.third_party?
-          = render partial: 'cases/sar/third_party_relationship', locals: { case_details: case_details }
+        - unless case_details.third_party == false && case_details.recipient == 'subject_recipient'
```
willmcb, line 22: I wonder if it might be clearer to wrap this in a method that explains what this mean so we can have:

```ruby
unless data_subject_is_requesting_date_for_themselves?
```
I think this may make it easier to reason about 👍

willmcb (commented):
Looks good 👍 - I have a few suggestions - let me know if you'd like to chat about them

app/helpers/cases_helper.rb
njpreddy: Great spot Will, done :-)

app/views/cases/offender_sar/_case_details.html.slim
Polsonby: As we just discussed, Nick, wondering if we might just get away with simpler logic anyway?

``` 
- if case_details.third_party?
   = render partial 'third_party_requester_details'
```
... because "third_party" specifically means "third party requester" for Offender SAR cases and is slightly independent of the recipient - recipient will either be subject or requester if "third_party" is true but I don't _think_ we need this part of the page to check

And then in the next section about where to send the data, we can just use 

```
- if case_details.third_party_recipient?
          = render partial: 'cases/offender_sar/third_party_recipient_details'
```

app/views/cases/offender_sar/_request_details.html.slim
```diff
@@ -0,0 +1,13 @@
+tr.request_dated.section
+  th
+    = t('common.case.request_dated')
+  td
+    = l case_details.request_dated, format: :default
```
Polsonby, line 5: There's a crash on this partial that Alex reported if the optional "request_dated" date field is nil. Please could you amend this line to:

```suggestion
    = l(case_details.request_dated, format: :default) if case_details.request_dated.present?
```

Polsonby (commented):
Looks great! I wasn't sure about nuking all the "one liner" view partials for each field, since that's how the other correspondence types are displayed but I really like the way the offender sar view files have ended up when you look at the files side by side in the folder, I think we're in a nice place with this PR :) 

There are a few tiny things we could potentially consider tweaking, but not sure I'd hold up merging for them! 

Happy to approve either way - just wondering if you might be able to make one change to request dated as [per comment above](https://github.com/ministryofjustice/correspondence_tool_staff/pull/1205#discussion_r458413703) though?

app/views/cases/offender_sar/_case_details.html.slim
```diff
@@ -7,30 +7,32 @@
...
-          = render partial: 'cases/offender_sar/third_party_company_name', locals: { case_details: case_details }
-        = render partial: 'cases/offender_sar/response_address', locals: { case_details: case_details }
-        = render partial: 'cases/offender_sar/request_dated', locals: { case_details: case_details }
-        = render partial: 'cases/offender_sar/requester_reference', locals: { case_details: case_details }
+
+        = render partial: 'cases/offender_sar/subject_details', locals: { case_details: case_details }
+
+        = render partial: 'cases/offender_sar/requested_on_behalf', locals: { case_details: case_details }
+
+        - if case_details.third_party? && (case_details.recipient == 'requester_recipient' || case_details.recipient == 'subject_recipient')
```
Polsonby, line 17: I _think_ we can just do 
```suggestion
        - if case_details.third_party?
```
- you can't ever have third_party requester AND third party recipient so it's redundant to test for it here
image

app/views/cases/offender_sar/_case_details.html.slim
```diff
@@ -7,30 +7,32 @@
...
+          = render partial: 'cases/offender_sar/third_party_details', locals: { case_details: case_details }
+
+        = render partial: 'cases/offender_sar/request_recipient', locals: { case_details: case_details }
+
+        - if (not case_details.third_party) && case_details.recipient == 'third_party_recipient'
```
Polsonby, line 22: We don't need to check for third party requester here, I think? Also, the enums for the recipient generate free predicate methods, so we can write the above as simply 
```suggestion
        - if case_details.third_party_recipient?
```

app/views/cases/offender_sar/_case_details.html.slim
```diff
@@ -7,30 +7,32 @@
...
-          = render partial: 'cases/sar/third_party_relationship', locals: { case_details: case_details }
-          = render partial: 'cases/offender_sar/requester_name', locals: { case_details: case_details }
-          = render partial: 'cases/offender_sar/third_party_company_name', locals: { case_details: case_details }
-        = render partial: 'cases/offender_sar/response_address', locals: { case_details: case_details }
-        = render partial: 'cases/offender_sar/request_dated', locals: { case_details: case_details }
-        = render partial: 'cases/offender_sar/requester_reference', locals: { case_details: case_details }
+
+        = render partial: 'cases/offender_sar/subject_details', locals: { case_details: case_details }
+
+        = render partial: 'cases/offender_sar/requested_on_behalf', locals: { case_details: case_details }
```
Polsonby, line 15: I'm just wondering about the partial name here - is this the requester details section?  Or is it just part of it? This uncertainty is what I hoped we'd resolve by naming the partials using the same names as the steps they relate to :)

app/helpers/cases_helper.rb
```diff
@@ -328,4 +328,15 @@ def show_escalation_deadline?(kase)
...
+    if recipient == 'requester_recipient'
+      t('helpers.label.offender_sar.recipient_type.recipient')
+    elsif recipient == 'subject_recipient'
+      t('helpers.label.offender_sar.recipient_type.data_subject')
+    else
+      t('helpers.label.offender_sar.recipient_type.third_party')
+    end
+  end
+
```
Polsonby, line 341: Could we move this into the offender_decorator, maybe? 

Also, wondering if we can get rid of this conditional by using the same approach as we did in the offender_decorator for `subject_type_display` if we wire up the keys in the en.yml file using the same strings as the values for recipient here. 

Compare with 
image

So we'd have keys in the yaml file like `helpers.label.offender_sar.recipient.requester_recipient` or more simply `helpers.label.offender_sar.requester_recipient` and then the method becomes something like 

```suggestion
def get_sar_recipient_label(recipient)
  I18n.t('helpers.label.offender_sar.recipient.' + recipient)
end
```
(... but using the equivalent form for a decorator i.e. `object.recipient`)

app/views/cases/offender_sar/_request_details.html.slim
njpreddy: Done!

app/views/cases/offender_sar/_case_details.html.slim
njpreddy: Done!

app/views/cases/offender_sar/_case_details.html.slim
njpreddy: Done!

Polsonby (approved):
Nice work :)

PR #1208, opened 2020-07-21
CT-2938 Updated Dispatch letters

willmcb:
This task was failing due to an erroneous capital letter in the
`LetterTemplateSeeder.rb` for a template abbreviation.

Also updates the prod rake tasks so that `:letter_templates` can be run independently rather than bundled with other tasks under the `:misc` task as is the case currently.

## Description
Updates script so it can be run on demo to update the data there.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
[CT-2938](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2938)

- run `rake db:seed:dev:letter_templates` locally and see if the templates are correct in your local copy.

### DIFF
--- a/db/seeders/letter_template_seeder.rb
+++ b/db/seeders/letter_template_seeder.rb
@@ -259,7 +259,7 @@ class LetterTemplateSeeder
       rec = LetterTemplate.find_by(abbreviation: 'solicitor-disclosed-covid')
       rec = LetterTemplate.new if rec.nil?
       rec.update!(name: 'Solicitor disclosed letter (COVID-19)',
-                  abbreviation: 'Solicitor-disclosed-covid',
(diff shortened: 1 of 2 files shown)


Polsonby (approved):
LGTM :) - thanks for this

I'm actually not so sure having it under :misc was such a bad idea as I thought, looking at the code now, because the tasks in there are all in this sort of "utility data" category that both should be run, and should be safe to run, repeatedly as needed on any environment

But I'm really happy that we've made it a standalone task now on the prod group because although the other tasks _should_ be safe, that's not the same as us _knowing_ they're safe to run - just in case our databases have been manually altered in the between times.

PR #1224, opened 2020-07-30
CT-2368 closed cases report for offender sar

ymao2:
## Description

There has been a closed report template for all case types including FOI, SAR, etc. but the problem with that template is that there will be some columns meaningful to one of case type but not irrelevant to other case types, it becomes worse for Offender SAR type,  this PR is to 
- create a close report for Offender SAR specific
- have the flexibility to allow adding another close report easily in the future 
- solve the report categories permission issues to show up the reports which is allowed to be seen by the current user  

## Self-review checklist

* [X] (3) Tests passing

### Screenshots

- for other type
Screenshot 2020-07-30 at 17 27 20
Screenshot 2020-07-30 at 17 27 30

- for offender sar
Screenshot 2020-07-30 at 17 27 44
Screenshot 2020-07-30 at 17 27 54

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2368

1 - login as branson user
2 - choose reports tabl
3 - click "Create report " button
4 - Choose "Offender SAR" and the sub report of "Closed cases report"
5 - choose the time frame
9 - confirm to  "Create custom report"
10 - click the link to download it from flash message


app/models/warehouse/case_report.rb
```diff
@@ -163,6 +163,17 @@ def dequote_and_truncate(text)
       def humanize_boolean(boolean)
         boolean ? 'Yes' : nil
       end
+
+      private 
+      def get_requester_type(kase)
+        if kase.offender_sar?
+          kase.third_party? ? kase.third_party_relationship : 'data subject'
```
Polsonby, line 170: Tiny thing, but could we get rid of this check for offender_sar? 

I think if we added a `requester_type` method to the model offender.rb we could put line 51 back how it was

app/models/report_type.rb
```diff
@@ -54,12 +63,19 @@ def self.respond_to_missing?(meth, include_private = false)
...
+    all_reports = {
       foi: self.standard.foi.order(:full_name),
       sar: self.standard.sar.order(:full_name),
       offender_sar: self.standard.offender_sar.order(:full_name)
     }
+    allowed_reports = []
+    self.get_report_categories(correspondence_types).each do | report_category |
+      if !REPORT_TYPE_CORRESPONDENCE_TYPE_MAP[report_category].nil?
+        allowed_reports |= all_reports[REPORT_TYPE_CORRESPONDENCE_TYPE_MAP[report_category]]
+      end
```
willmcb, line 76: Is the intention here to ensure that `allowed_reports` has only unique values in it - the bitwise OR (`|=`) with assignment is not that common to see so just checking that its not meant to be `||=`?

willmcb (approved):
lgtm 👍

app/services/stats/etl/closed_cases.rb
```diff
@@ -2,58 +2,14 @@
...
-        #
-        # Ordering of generation matters, each function call creates a temp CSV
-        @_temp_files = generate_header_fragment + generate_data_fragments
-        self
-      end
 
-      def transform
-        self
+      def result_name
+        RESULTS_NAME
```
Polsonby, line 10: Super tiny thing, but could you clarify what this method's for?

The fact that it has both plural and singlular forms of "result" here caught my eye

ymao2: the method is return the file name of csv zip file (used in func: transform()),  the original code uses the "result_name",  I changed the constant name to keep consistent

Polsonby (approved):
Great stuff :)

PR #1230, opened 2020-08-06
CT-2974 Add ex-probation user to subject type

njpreddy:
## Description

CT-2974 Add "ex-probation user" subject type

## Self-review checklist

* [X] (3) Tests passing

### Screenshots

CT-2974 add ex-probation user to subject type

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=CT&modal=detail&selectedIssue=CT-2974

On the "Who is the subject?" page of the case creation journey you should see "Ex-probation service user" as a "Subject type" option.

### DIFF
--- a/app/models/case/sar/offender.rb
+++ b/app/models/case/sar/offender.rb
@@ -56,6 +56,7 @@ class Case::SAR::Offender < Case::Base
     detainee: 'detainee',
     ex_detainee: 'ex_detainee',
     probation_service_user: 'probation_service_user',
+    ex_probation_service_user: "ex_probation_service_user"
   }
 
   enum recipient: {
--- a/spec/models/case/sar/offender_spec.rb
+++ b/spec/models/case/sar/offender_spec.rb
@@ -59,6 +59,7 @@ describe Case::SAR::Offender do
         expect(build(:offender_sar_case, subject_type: 'ex_offender')).to be_valid
         expect(build(:offender_sar_case, subject_type: 'offender')).to be_valid
         expect(build(:offender_sar_case, subject_type: 'probation_service_user')).to be_valid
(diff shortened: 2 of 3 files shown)


app/models/case/sar/offender.rb
```diff
@@ -56,6 +56,7 @@ def searchable_fields_and_ranks
     detainee: 'detainee',
     ex_detainee: 'ex_detainee',
     probation_service_user: 'probation_service_user',
+    ex_probation_service_user: "ex_probation_service_user"
```
willmcb, line 59: Really small thing but I generally in ruby double quotes are only needed / used with string interpolation. I think it makes sense for these to be single quotes like the others as it is easier to read as well 👍

spec/models/case/sar/offender_spec.rb
```diff
@@ -218,15 +219,15 @@
...
         expect(kase).not_to be_valid
         expect(kase.errors[:third_party_name]).to eq ["can't be blank if company name not given"]
         expect(kase.errors[:third_party_company_name]).to eq ["can't be blank if representative name not given"]
       end
 
       it 'does not validate third_party names when ecipient is not third party too' do
-        kase = build :offender_sar_case, third_party: false, third_party_name: '', 
+        kase = build :offender_sar_case, third_party: false, third_party_name: '',
                       third_party_company_name: '', recipient: 'subject_recipient'
         expect(kase).to be_valid
```
willmcb, line 232: Going by the it statement I would have thought this should be an invalid test 🤔

willmcb (commented):
Could we have a quick chat? Not sure I understand what's happening in the tests here.

PR #1233, opened 2020-08-10
CT-2932 Update search fields

Polsonby:
## Description
We needed to make search work much better than it was working for Offender SAR. We've also had feedback concerning issues with search from the London Disclosure team. Although fixing those wasn't the aim of this ticket, we are making a couple of changes to the way searching works for all Correspondence Types in this PR (see 3. below), hoping these will be positive for all users, not just the Offender SAR team.

Search uses Postgres full text search for cases. Whenever an update is made, or the indexer is run, Postgres calculates tokenised values and weights that will be used to find this case during searches. This is saved in a field called `document_tsvector` on each row. The SQL to maintain this is mildly funky (to say the least) so the app uses a gem called `pg_search` to abstract away the messy stuff. 

The search feature is pretty complex and does a lot of things around filtering and saving searches, but the net result is we are left with a pretty simple set of controls for adjusting how it works.

1. In Case::Base we have the default fields and weighting for all correspondence types as a starting point. We haven't messed with this.
BEFORE and AFTER
image

2. In Case::SAR::Offender, we were adding to this list
BEFORE
image

This PR changes this so that we don't extend the set, but replace it with our own weightings because the defaults are not good for Offender SAR cases
AFTER
image

3. When searches are executed, we have some configuration in the Searchable concern 
BEFORE
image

In this PR we've made two key changes here
AFTER
image

First, we now use multiple search terms to _narrow_ the search, not extend it. "And", not "or" if you like. 

Second, we now support finding matches using stemming, so that entering the first few characters will match compatible target strings. This wasn't the case before, so imagine you had a case number `200727002` - you can find it using the whole number, but if you searched for `20072700` (missing off the final digit through a copy/paste error) nothing would be returned. Now, it returns 200727001 and 200727002 if they exist. 
 

## Self-review checklist

* [x] (3) Tests passing

### Screenshots

All the Davids, thanks...

image

Or just this one

image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2932

### Deployment

We need to run the reindex command in order to populate the tokens in the document_tsvector column on every environment after deploying there. 

`./cts cases reindex`

Before you try searching with the new branch, you need to run the reindex command in order to populate the tokens in the document_tsvector column, or the cases won't be findable. 

`./cts cases reindex`

Now you can experiment with finding cases you know are there. You should always be able to find a record using the prisoner number or the case number - this worked fine before. But if you are using e.g. last name and year of birth, it was really hard to  find a case even if you knew it was there. Suppose you search for "Smith", of which there are thousands of matches in the real data, and you try adding another term - like the year of birth '1980' - to narrow the search down.

Previously, this would make things _worse_ because it gave you union of the sets, when in fact we wanted the intersection. But now you it works to narrow down the search to just a few and it's easy to find the right match. 

You can search for date of birth using ISO 8601 date format - e.g. `1980-03-24` - or using just the year.

Locally, you can set up a couple of test cases with the same name but different year of birth, plus other cases using the same year of birth but different names. You should be able to find the correct cases using the name, then see the right one by adding the year of birth - and see other cases with the same year of birth.


app/models/concerns/searchable.rb
```diff
@@ -9,9 +9,10 @@ module Searchable
     pg_search_scope :search,
                     against: searchable_fields_and_ranks,
                     using: { tsearch: {
-                               any_word: true,
+                               any_word: false,
```
ymao2, line 12: Hi Ian, just curious for this change,  this change seems to me will affect other case types, is it any consequence outcome for this?

spec/models/case/sar/offender_spec.rb
```diff
@@ -218,15 +218,15 @@
...
       it 'validates third party names when recipient is third party' do
-        kase = build :offender_sar_case, third_party: false, third_party_name: '', 
+        kase = build :offender_sar_case, third_party: false, third_party_name: '',
                       third_party_company_name: '', recipient: 'third_party_recipient'
         expect(kase).not_to be_valid
         expect(kase.errors[:third_party_name]).to eq ["can't be blank if company name not given"]
         expect(kase.errors[:third_party_company_name]).to eq ["can't be blank if representative name not given"]
       end
 
       it 'does not validate third_party names when ecipient is not third party too' do
```
willmcb, line 228: Small thing, also pre-existing this PR but should we change the spelling of `ecipient` here? Could make a list of this sorts of things and combine them in a later PR in one fell swoop.

willmcb (approved):
Nice, LGTM 👍

app/models/concerns/searchable.rb
Polsonby: Sorry - missed this! Yes, we want to experiment with it on the London side also as we suspect it may give better results for them too

PR #1234, opened 2020-08-10
CT-2989 kilo users cases view

ymao2:
## Description

The issue was because when the restriction of responder role was moved from service level into policy level,  the way of construction sub_query was different from the original one due to the bad performance of old way, 
old way

-----
  case_ids = Assignment.with_teams(user.responding_teams).pluck(:case_id)
  open_scope.where(id: case_ids)

new way 
--------
          team_restriction = Assignment
              .joins('join teams_users_roles on assignments.team_id=teams_users_roles.team_id')
              .where('teams_users_roles': {user_id: @user.id, role: :responder.to_s})
              .select(:case_id).distinct       
          @scope.where(id: team_restriction)
there were some extra conditions from "with_teams" from old way,  which were related to the state of assignments,  the "rejected" has been excluded in the allowed state.  The new way didn't take such state condition into account, this fix is to add this state condition back

## Self-review checklist

* [X] (3) Tests passing

### Screenshots

I have created two cases and assigned them to and LAA responder - they see this when they log in:
Screenshot 2020-08-05 at 13 44 36

I reject the first case - go back to the cases page you should see this:
Screenshot 2020-08-05 at 13 39 58

Rather than this:
Screenshot 2020-08-05 at 13 46 55

1. Create a case as a dispatcher, say FOI
2. Assign to a business unit, say LAA
3. Log in as that BU's respoder
4. Click on the recently created case
5. Reject it
6. Click back to the 'Cases' view
7. The rejected case should no longer be visible for that user.

### DIFF
--- a/app/policies/case/foi/standard_policy.rb
+++ b/app/policies/case/foi/standard_policy.rb
@@ -13,6 +13,7 @@ class Case::FOI::StandardPolicy < Case::BasePolicy
           team_restriction = Assignment
               .joins('join teams_users_roles on assignments.team_id=teams_users_roles.team_id')
               .where('teams_users_roles': {user_id: @user.id, role: :responder.to_s})
+              .where.not(state: [ :rejected ] )
               .select(:case_id).distinct       
           @scope.where(id: team_restriction)
         else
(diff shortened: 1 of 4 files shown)


willmcb (approved):
Looks good 👍

Polsonby (approved):
Good work, both!

PR #1236, opened 2020-08-11
CT-2873 Changes to new data request form

Polsonby:
## Description
This PR makes significant changes to the new Data Requests creation flow

* Removes the support for creating multiple data requests in a single page
* Adds pre-determined options for request type (changing the field name from `data`)
* Requires a mandatory, supplemental "notes" field if the request_type is `other`
* Adds optional "from" and "to" dates

This saves the data in the database on success, display and editing are added in other PRs

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots

image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2873

### Deployment
This PR has database migrations that must be run using `rake db:migrate` to add and update columns on the DataRequest table

The database structure is not backwards compatible with master after running the migrations. Down migrations are supported, though, so if you need to run it and then switch back to master you'll need to do `rake db:rollback` for each database migration here; there are three. 

After running the database migrations, you should see the new screen when you click "record data request" on an offender SAR case. After saving the request, you should see the location and request type listed in the data requests table. You can check other values in the database.

Request type note should be mandatory if and only if "other" is chosen. 

The date fields are optional individually and together. The only validation is that the `from` date can't be after the `to` date


app/views/cases/data_requests/edit.html.slim
```diff
@@ -19,7 +19,7 @@ span.visually-hidden
   .column-full
 
     h2.data-request__location.lede= @data_request.location
-    p.data-request__data= @data_request.data
+    p.data-request__request_type= @data_request.request_type
```
ymao2, line 23: my feeling the old field 'data' is more closer to the new field of "request_type_note",  is it just temporary solution here? as the  "request_type_note" is missing from this page, is it going to be added in another ticket?

app/models/data_request.rb
```diff
@@ -30,13 +45,33 @@ def kase
...
+
   def update_cached_attributes(new_data_request_log)
     self.cached_date_received = new_data_request_log.date_received
     self.cached_num_pages = new_data_request_log.num_pages
   end
 
   def clean_attributes
-    [:location, :data]
+    [:location]
       .each { |f| self.send("#{f}=", self.send("#{f}")&.strip) }
```
ymao2, line 75: maybe add "request_type_note" here instead? although not full sure why here need some clean-up

app/models/data_request.rb
```diff
@@ -4,15 +4,30 @@ class DataRequest < ApplicationRecord
...
+  enum request_type: {
+    offender: 'offender',
+    all_prison_records: 'all_prison_records',
+    all_nomis_records: 'all_nomis_records',
+    nomis_contact_logs: 'nomis_contact_logs',
+    probation_records: 'probation_records',
+    prison_and_probation_records: 'prison_and_probation_records',
+    other: 'other'
+  }
+
```
ymao2, line 28: looked at the ticket,  there is no type called "offender",  do you need to remove it?

ymao2: hmm, just checked the rest of codes, seems you use "offender" for so many places..... but I didn't see it from Jira ticket

app/services/data_request_create_service.rb
```diff
@@ -1,18 +1,18 @@
...
     @user = user
     @case = kase
-    @new_data_requests = build_data_requests(data_requests)
+    @data_request = build_data_request(data_request_params)
   end
 
   def call
     ActiveRecord::Base.transaction do
       begin
         @result = :unprocessed
```
ymao2, line 14: saw this :unprocessed has been removed, maybe remove it here?

db/migrate/20200811222853_rename_data_request_note_to_request_type_note_for_data_request.rb
```diff
@@ -0,0 +1,5 @@
+class RenameDataRequestNoteToRequestTypeNoteForDataRequest < ActiveRecord::Migration[5.2]
+  def change
+    rename_column :data_requests, :data_request_note, :request_type_note
+  end
+end
```
ymao2, line 5: any particular reason we add "data_request_note" first, then rename it as "request_type_note"?

app/models/data_request.rb
Polsonby: Oh man - Thanks! I inherited that from somewhere else and hadn't checked properly, just saw there were 7 🤦 

I used it EVERYWHERE 😂 - changing

app/models/data_request.rb
```diff
@@ -30,13 +45,33 @@ def kase
 
   private
 
+  def validate_from_date_before_to_date
+    if date_from.present? && date_to.present? && date_from >= date_to
```
willmcb, line 49: I initially read this as if the date was the "present", as in now? Was thinking how can these dates be the the same haha 🤦

spec/site_prism/page_objects/pages/cases/data_request_page.rb
```diff
@@ -1,7 +1,7 @@
 module PageObjects
   module Pages
     module Cases
-      class DataRequestPage < SitePrism::Page
+      class DataRequestPage < PageObjects::Pages::Base
         set_url '/cases/{case_id}/data_requests{/new}'
```
willmcb, line 5: What is `{/new}` referring to here - this is not a syntax I am familiar with. Usually interpolated values come after the `/`

db/migrate/20200811222853_rename_data_request_note_to_request_type_note_for_data_request.rb
Polsonby: I messed up the name first time round; then I think I had intervening migrations in commits in git. Seemed easier to rename than unpick everything

app/models/data_request.rb
willmcb: Maybe this could be wrapped in a `dates_submitted?` method or something.

```ruby
def dates_submitted?
  dates_from.present? && date_to.present?
end

# then:

if dates_submitted? && date_from >= date_to
```

I think this is really just me misinterpreting things at first glance more than anything, but maybe something to think about.

Dates are odd in that `date.past?` (don't thing this is a real method) `date.present?` have sort of ambiguous names as they can relate to "existence" in the latter case, and a "place in time" for both.

willmcb:
Looks good to me - ran the manual test works well. Just checking - validations fail if `date_from` and `date_to` are the same day. This makes sense but I was wondering what if the thing that is being requested was on the same date? Not much of an issue as you can just put the next day, but I thought I'd ask the question 👍

willmcb (approved):
👍

ymao2 (approved):

app/models/data_request.rb
Polsonby: Good spot, thanks! It was annoying me about the same date thing too. I completely see your point about the English language of .present?, have done something along these lines, nice idea!

spec/site_prism/page_objects/pages/cases/data_request_page.rb
Polsonby: This was already in place - I'm not sure exactly why they did it this way 🤔 

I'll read up about it sometime!

willmcb (approved):

PR #1257, opened 2020-08-21
CT-2368 Fixed the bug of wrong subject name

ymao2:
## Description

Fixed the bug Alex found during QA:  showing subject type rather than required full name of data subject 

regarding whether a test spec is needed to covering testing specific column of a report,  IMO, there is no need, there have been a few test specs for the reports, mainly for covering the key features like the scope of report etc, whether a suitable task is raised or not or the stats are correctly or not,  and also the source of the reports (e.g. the source for close report is from warehouse table) has been covered on. model level separately to make sure the outcome is correct

## Self-review checklist

* [X] (3) Tests passing

### DIFF
--- a/app/services/stats/etl/offender_sar_closed_cases.rb
+++ b/app/services/stats/etl/offender_sar_closed_cases.rb
@@ -18,7 +18,7 @@ module Stats
 
       FIELD_COLUMNS = [
         'number', 
-        'sar_subject_type', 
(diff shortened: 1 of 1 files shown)


willmcb:
LGTM 👍 - I agree as well no need for a test spec.

willmcb (approved):
👍

PR #1259, opened 2020-08-24
CT-3014 - Change copy on page and related places in app.

willmcb:
## Description

Change "Pages for dispatch" to "Final page count" in copy.

Also updates app elsewhere (specs, models etc.) to reflect this.

## Self-review checklist

* [ ] (3) Tests passing

#### Before:
Screenshot 2020-08-24 at 13 27 36

**_Edit page_**
Screenshot 2020-08-24 at 13 27 11

#### After:
Screenshot 2020-08-24 at 13 25 59

_**Edit page:**_
Screenshot 2020-08-24 at 13 26 15

### Related JIRA tickets

[CT-3014](https://dsdmoj.atlassian.net/browse/CT-3014)

- log in as Branston user
- go to case show page check copy is as above
- click "Update final page count"
- check that copy is as above

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -693,4 +693,4 @@ DEPENDENCIES
   yard
 
 BUNDLED WITH
-   1.17.1
+   2.1.4
--- a/app/models/case/sar/offender.rb
+++ b/app/models/case/sar/offender.rb
@@ -60,10 +60,10 @@ class Case::SAR::Offender < Case::Base
                  third_party_company_name: :string,
                  late_team_id: :integer,
                  third_party_name: :string,
-                 number_dispatched_pages: :string,
+                 number_final_pages: :string,
                  number_exempt_pages: :string
 
-  attribute :number_dispatched_pages, :string, default: '0'
+  attribute :number_final_pages, :string, default: '0'
   attribute :number_exempt_pages, :string, default: '0'
 
   enum subject_type: {
@@ -109,7 +109,7 @@ class Case::SAR::Offender < Case::Base
   validate :validate_third_party_address
 
   validate :validate_request_dated
-  validate :validate_number_dispatched_pages
+  validate :validate_number_final_pages
   validate :validate_number_exempt_pages
 
   before_validation :reassign_gov_uk_dates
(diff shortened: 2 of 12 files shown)


app/models/case/sar/offender.rb
```diff
@@ -60,10 +60,10 @@ def searchable_fields_and_ranks
                  third_party_company_name: :string,
                  late_team_id: :integer,
                  third_party_name: :string,
-                 number_dispatched_pages: :string,
+                 number_final_pages: :string,
                  number_exempt_pages: :string
```
ymao2, line 64: @willmcb  should number_final_pages and number_exempt_pages be integer?  then you don't have to check other possible values like string, so the logic for validating those values are more simpler

willmcb: Good idea will change in separate PR

Polsonby:
LGTM! Happy to approve once spec passing - I wonder whether we should put the bundler version back though until we're all on the same version?

Gemfile.lock
```diff
@@ -693,4 +693,4 @@ DEPENDENCIES
   yard
 
 BUNDLED WITH
-   1.17.1
+   2.1.4
```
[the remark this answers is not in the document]
willmcb: Nope 🤦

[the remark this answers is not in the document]
willmcb: Vim asked me to run a bundle update to make my language server work I think this may have caused this.

app/models/case/sar/offender.rb
```diff
@@ -108,9 +108,15 @@ def searchable_fields_and_ranks
...
   validate :validate_third_party_address
 
+  validates :number_final_pages,
+            numericality: { only_integer: true, greater_than: 0,
+                            message: 'must be a positive whole number' }
+
+  validates :number_exempt_pages, 
+            numericality: { only_integer: true, greater_than: 0,
+                            message: 'must be a positive whole number' }
+
```
ymao2, line 118: hmm very interesting, never know rails can have such feature :),  only thing I am thinking  "only_integer" is necessary here as we have defined the type is integer,  seems to me it is redundant one.

spec/models/case/sar/offender_spec.rb
```diff
@@ -211,69 +211,69 @@
...
+  describe 'number_final_pages' do
     context 'invalid values' do
-      it 'errors when string represents a float' do
-        kase = build(:offender_sar_case, number_dispatched_pages: "24.45")
+      it 'errors when float is used' do
+        kase = build(:offender_sar_case, number_final_pages: 24.45)
         expect(kase).not_to be_valid
-        expect(kase.errors[:number_dispatched_pages]).to eq ['must be a positive whole number']
+        expect(kase.errors[:number_final_pages]).to eq(['must be a positive whole number'])
       end
```
ymao2, line 253: Those test specs do make me wonder... 
I removed those examples for testing those page fields as string in my PR,  as we don't test non-integer values on integer field, non-boolean value for bool type field,... and so on on model level usually ----> although it is definitely no harm if you want to check,  it may be more useful to test non-integer value from controller/view level. But anyway, this is only for my doubt rather than a issue

app/models/case/sar/offender.rb
willmcb: I think this will throw an error from to the form if numbers with decimals are submitted 👍

spec/models/case/sar/offender_spec.rb
willmcb: Yeah I saw that in your PR - was wondering what your thoughts were on this. I think you may be right regarding the controller / view tests, but I may leave it for now and then re-look at it later on.

PR #1273, opened 2020-08-27
CT-3034, CT-3036 Fixed the bugs related to offender sar cases

ymao2:
## Description

- Minor fixes for close/open reports like column name, order, and wrong logic
- Bug fix for performance report:  if the month is not within the range of the performance report, it will throw an error, 
- Changed the Redis.exists() to Redis.exists? due to upgrade of gem version 
- Added translation for even of "destroy"

## Self-review checklist

* [X] (3) Tests passing

### Screenshots

- Minor fixes for close/open reports like column name, order, and wrong logic
For open report
before 
open_case_before
after 
Screenshot 2020-08-27 at 18 22 33

For close report
before 
close_case_before
After 
close_case_after_fix

- Bug fix for performance report:  if the month is not within the range of the performance report, it will throw an error, 
This has been covered by new test spec I added in this PR

- Changed the Redis.exists() to Redis.exists? due to upgrade of gem version 
Fixed the error message 
```
Redis#exists(key)` will return an Integer in redis-rb 4.3. `exists?` returns a boolean, you should use it instead. To opt-in to the new behavior now you can set Redis.exists_returns_integer = true. To disable this message and keep the current (boolean) behaviour of 'exists' you can set `Redis.exists_returns_integer = false`, but this option will be removed in 5.0. (/usr/src/app/app/services/stats/base_monthly_performance_report.rb:94:in `block in report_details'
````
- Added translation for even of "destroy"
Before
destroy_before_add
After
destroy_after_add

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3036
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3034
### Deployment

If you try to run this branch from local, please make sure you have run sidekiq locally as well

Download the open case report
1 - Login as Branston user
2 - click "All open cases" table
3 - Click the "Download cases" link at right top of the case list
4 - A csv file should be downloadable, and check the columns and vlaues
Download the close cases report
1 - Login as Branson usesr
2 - Choose the "Reports" menu from top
3 - Click "Create report" button 
4 - Choose "Offender SAR" type
5 - Choose sub report "Closed case report"
6 - Type in the time period you want to produce 
7 - a Message should be popped from top which give the link for you to download it
8 - a zip file will be downloaded after you click that link
9 - unzip it and check the columns and values from the csv file 
Download the performance report
1 - repeat 1 to 4 as above
2- Choose sub report "Monthly report (Offender SARs)"
3 - a Message should be popped from top which give the link for you to download it
if the message tells you this report is not ready yet please wait for a few seconds to try again
4 - a excel file should be downloaded

### CODE REVIEW

File: app/services/stats/base_monthly_performance_report.rb [hunk 1]
```diff
@@ -172,20 +172,27 @@ def create_background_jobs
...
+    def merge_stats(data_collector)
+      merged_stats = init_merged_stats()
       data_collector.each do |data|
         data_object = JSON.parse(data, symbolize_names: true)
         data_object.each do |month, stats|
           month_key = month.to_s.to_i > 0 ? month.to_s.to_i : month
           stats.each do |stat_item, value|
-            if !merged_stats[month_key].key?(stat_item)
-              merged_stats[month_key][stat_item] = 0
+            if !merged_stats[month_key].nil?
```
Polsonby, line 190: Just a style thing, you can probably do 

```suggestion
            if merged_stats[month_key]
```
if you're expecting an object and not specifically trying to differentiate between `false` and `nil`, or 
```suggestion
            if merged_stats[month_key].present?
```
is just a slightly nicer expression than "not is it nil?"

Polsonby:
LGTM on reading, but if you have a moment, it would be really great to include screenshots showing the fixes or quick steps to test

app/services/stats/base_monthly_performance_report.rb
ymao2: @Polsonby ,   the change I made is necessary,  that line is not a change/replacement to the line 185, I added a new line to check whether the month is within the months the report required, then next step is to check whether the metric (such as non_triggered_open_in_time) is in that month or not which is the line 185 trying to do .

ymao2: If not adding this extra check,   an error will be thrown out saying Nil:class doesn't have key method which is the error happened when we demoed to Sarah

ymao2: thank you for your suggestion, actually remind me I should use merged_stats.key?(month_key)

Polsonby (approved):
Looks good to me :)

willmcb (approved):
👍

PR #1275, opened 2020-08-28
CT-3033 Update cover page to display request type properly

Polsonby:
## Description
Small change to output the request type consistently on the data requests table both on the case show page and on the cover page for printing. Split out a partial which is referenced from both places. 

We added a pages received column which is intended to be empty on the cover page even when there is a value for the data request - this is a space for the vetters to write in during the vetting process, as the cover page is printed off only once after the case and data requests have been logged on the system. 

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
BEFORE
image

AFTER
image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3033
### Deployment
None
On an Offender SAR case with some data requests that have from and/or to dates populated, view the cover page using the Preview Cover Page button and verify the appearance of the request type is the same as on the request table entry on the show page

### DIFF
--- a/app/views/cases/offender_sar/_data_request_request_type.html.slim
+++ b/app/views/cases/offender_sar/_data_request_request_type.html.slim
@@ -1,0 +1,20 @@
+                = t :"cases.data_requests.index.request_type.#{data_request.request_type}"
+                - if data_request.other? && data_request.request_type_note.present?
+                  = ": #{data_request.request_type_note}"
+                - if data_request.request_dates_either_present?
+                  span.data-detail<
+                    br
+                    - if data_request.date_from.present?
+                        time datetime="#{data_request.date_from}"
+                          = l data_request.date_from, format: :default
+                    - if data_request.request_dates_both_present?
+                      = " - "
+                    - if data_request.request_date_from_only?
+                      span<
+                        = t('cases.data_requests.onwards')
+                    - if data_request.request_date_to_only?
+                      span<
+                        = t('cases.data_requests.upto')
+                    - if data_request.date_to.present?
+                        time< datetime="#{data_request.date_to}"
+                          = l data_request.date_to, format: :default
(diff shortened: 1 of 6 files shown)


app/views/cases/offender_sar/_data_request_request_type.html.slim
```diff
@@ -0,0 +1,20 @@
+                = t :"cases.data_requests.index.request_type.#{data_request.request_type}"
+                - if data_request.other? && data_request.request_type_note.present?
+                  = ": #{data_request.request_type_note}"
+                - if data_request.request_dates_either_present?
+                  span.data-detail<
+                    br
+                    - if data_request.date_from.present?
```
willmcb, line 7: Just an idea - could we make these methods private and then just have a public wrapper method that outputs the right thing - then we don't have to repeat the spans? 🤔

spec/views/cases/cover_pages/show_html_slim_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        date_from: Date.new(2018, 8, 15),
+        date_to: Date.new(2019, 8, 15),
+        cached_num_pages: 32,
+        cached_date_received: Date.new(1972, 9, 25),
+      )
+    }
+
+    before do
+      assign(:data_request, data_request)
+      assign(:case, data_request.kase)
```
willmcb, line 19: I think here you may need to create a case with a known ID for the test to pass.

willmcb (approved):
One thought about a possible non-blocking refactor. I think the test is failing as the case numbers don't match.

spec/views/cases/cover_pages/show_html_slim_spec.rb
Polsonby: This is a pattern that we've use in existing specs - the factory for the data request creates the parent case I think - passed locally 🤔

PR #1277, opened 2020-08-30
CT-3044 fix time sensitive specs

Polsonby:
## Description
When we run our specs early in the morning, in the evening, or at the weekend, certain specs fail because they make assumptions about the date or time of day that the specs are to be run. 

This causes a great deal of friction and delay since we work flexible hours and often need to run tests outside the normal 9-5. Also, we have automated PRs generated by dependabot which suffer from the same problems if they're triggered in these time windows. We need to make our test suite so it can be run any time of day or on any day successfully without reporting false positives.

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
n/a

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3044

### Deployment
n/a

n/a

### DIFF
--- a/spec/features/cases/foi/case_listing/incoming_cases_spec.rb
+++ b/spec/features/cases/foi/case_listing/incoming_cases_spec.rb
@@ -5,44 +5,62 @@ feature 'listing incoming on the system' do
   given(:press_officer) { find_or_create :press_officer }
   given(:private_officer) { find_or_create :private_officer }
 
-  given(:assigned_case) { create :assigned_case,
-                                 created_at: 1.business_days.ago,
-                                 identifier: 'assigned_case' }
-  given(:fresh_assigned_case) { create :assigned_case,
-                                       identifier: 'fresh_assigned_case' }
+  given(:assigned_case) do
+    Timecop.freeze(Date.new(2020, 8, 19)) do
+      create :assigned_case,
+           created_at: 1.business_days.ago,
+           identifier: 'assigned_case'
+    end
+  end
+  given(:fresh_assigned_case) do
+    Timecop.freeze(Date.new(2020, 8, 19)) do
+      create :assigned_case,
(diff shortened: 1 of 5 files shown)


danielatdigital (approved):

spec/features/cases/foi/case_listing/incoming_cases_spec.rb
```diff
@@ -5,44 +5,62 @@
   given(:press_officer) { find_or_create :press_officer }
   given(:private_officer) { find_or_create :private_officer }
 
-  given(:assigned_case) { create :assigned_case,
-                                 created_at: 1.business_days.ago,
-                                 identifier: 'assigned_case' }
-  given(:fresh_assigned_case) { create :assigned_case,
-                                       identifier: 'fresh_assigned_case' }
+  given(:assigned_case) do
+    Timecop.freeze(Date.new(2020, 8, 19)) do
```
willmcb, line 9: Just thinking - we are making multiple calls with the same date, maybe we could wrap everything and then call it once like this using `before` and `after` calls:

```ruby
describe "some set of tests to mock" do
  before do
    Timecop.freeze(Time.local(1990))
  end

  after do
    Timecop.return
  end

  it "should do blah blah blah" do
  end
end
```

willmcb (approved):
Small comment that may make the tests a bit cleaner / faster 👍

spec/features/cases/foi/case_listing/incoming_cases_spec.rb
willmcb: More info: https://github.com/travisjeffery/timecop

Polsonby: I think we have that explicitly turned off 🤔 

Let's try it out sometime

PR #1294, opened 2020-09-06
CT-2994 Fixed the search feature for kilo users

ymao2:
## Description

This PR is to fix the bug related the scope of the users including the kilo users can see FOI cases, the detail for explaining the permission/expectation for those can be checked from Jira ticket. 
Basically 
- FOI and Internal Review cases can be discovered by any user on the tool but subject to the scope which case type the user is allowed to see --> this is the FOI scope policy,
- For responder: "Open cases tab" and “Closed cases” only show the cases being assigned to them ---> this is functional feature, not related to scope policy at all
- Other case types are restricted to the role's permission per case type

The reason about why this issue happens is that I changed the scope of FOI policy by moving the restriction which was under Open/closed cases tab to policy level by misunderstanding/not knowing the rules and workflow. The solution I did is to move the responder restriction for specific pages back to the level where it should be.  I was thinking to expand the policy level by taking extra factor of which page the user is on, but I found this method will make the policy more vague (as fundamentally the FOI is public available) and also will add the extra parameter for all other policies.

## Self-review checklist

* [] (1) Quick stakeholder demo done OR
* [X] (3) Tests passing

### Screenshots

None
### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2994
### Deployment

No extra steps

1 - Login as a non-admin user from one of london teams
2 - create a FOI case without being assigned to any team, write down the case number
3 - Login as another non-admin user from one of london teams
4 - search the case being created at step2,  
5 - the case at step2 should be shown, you can click it to view the detail but no other operations are allowed


spec/features/cases/offender_sar/case_search_spec.rb
```diff
@@ -100,9 +100,9 @@
     let(:searcher) { responder }
     let(:search_term) { "Common text" }
 
-    scenario 'finds offender_sar_case but not FOI' do
-      expect(cases_search_page.search_results_count.text).to eq "1 case found"
-      expect(cases_search_page.case_list.count).to eq 1
+    scenario 'finds offender_sar_case and FOI' do
```
willmcb, line 103: How does this find an FOI case when one is not defined in the spec?

willmcb: Ah I see its called `kase` - might be a bit more readable to change it from `kase` to `foi_kase` in the `given()` on line `7` so that it's clearer that its an FOI we are creating and then searching for later

willmcb (approved):
Looks good to me - but I think we should wait for @Polsonby to review as well 👍

app/policies/case/sar_policy.rb
```diff
@@ -19,19 +19,11 @@ def resolve
...
+          scopes << @scope.where(id: Assignment.team_restriction(@user.id, :responder))
         end
 
         if @user.approver?
-          team_restriction = Assignment
-              .joins('join teams_users_roles on assignments.team_id=teams_users_roles.team_id')
-              .where('teams_users_roles': {user_id: @user.id, role: :approver.to_s})
-              .select(:case_id).distinct      
-          scopes << @scope.where(id: team_restriction)
+          scopes << @scope.where(id: Assignment.team_restriction(@user.id, :approver))
```
Polsonby, line 26: Nice! 👍

Polsonby (approved):
Great work :) 

It's hard to verify the case visibility in the spec is correct in review given the way our specs are set up; maybe it's just me? Feels like I'd need to understand every single case setup in [that massive list](https://github.com/ministryofjustice/correspondence_tool_staff/pull/1294/files#diff-b3ae9a3ca7e3fb1274b4219a04a6b1aeL255)

But it would be good to ask PM to pay special attention to who can see what when QAing

Everything I could check locally in an almost-empty DB worked perfectly; tested FOI case visibility between kilos in search, verified SAR could not be seen in search, checked Branston user couldn't see FOI, etc

PR #1300, opened 2020-09-09
CT-3017 changed the conditions of showing general close report based on permission

ymao2:
## Description

This PR  is to "hide" the general-closed-cases-report for offender-sar user as Offender-SAR has own version of closed-cases-report which is under 'Offender-SAR' type.  The method I used is to hide them by checking the user's permission.   I also changed some flags for 'R007' in DB in order to achieve it

## Self-review checklist

* [X] (3) Tests passing

### Screenshots

Before 
Screenshot 2020-09-09 at 10 11 02

after
Screenshot 2020-09-09 at 10 14 24

for FOI / SAR type, nothing change 
Screenshot 2020-09-09 at 10 11 22

 
### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3017
### Deployment

Please run data migrate command 
rails data:migrate

1. login as branston user
2. click "Reports" tab
3. click "Create report"
4, you should only see 'Offender SAR'
5. sign out
6. login as london user
7. repeat 2, 3 step
8. you should see 'FOI', 'SAR' and 'General closed cases report'


app/controllers/stats_controller.rb
```diff
@@ -163,10 +163,17 @@ def create_spreadsheet(report)
     axlsx
   end
 
+  def is_general_close_report_present?
+    close_report_scope = Pundit.policy_scope(current_user, ReportType.closed_cases_report)
+    close_report_scope.count > 0
```
Polsonby, line 168: I think this could just be
```suggestion
    Pundit.policy_scope(current_user, ReportType.closed_cases_report).present?
```

Polsonby: Also a tiny thing, I'm a little confused by the name of the method here. Would it make sense to call the method `closed_cases_report_present?` or are we distinguishing it from another closed cases report?

ymao2: @Polsonby , yes there is another version of closed report for offender-sar specifically which is R401. 
R007 is the one I call it as "General" which is used by other case types like FOI SAR,  I am not sure how to name it

app/controllers/stats_controller.rb
```diff
@@ -163,10 +163,17 @@ def create_spreadsheet(report)
...
+  end 
+
   def user_permitted_custom_report_types
     # find out the scope of the custom report types the user can see via set intersection operation
     @correspondence_types = CorrespondenceType.custom_reporting_types & current_user.permitted_correspondence_types
-    @correspondence_types += [self.class.closed_cases_correspondence_type]
+    if is_general_close_report_present?
+      @correspondence_types += [self.class.closed_cases_correspondence_type]
```
Polsonby, line 175: I find the += [...] a bit confusing here. Are we always going to get an Array out of line 173? Then we should just be able to do

```suggestion
      @correspondence_types << self.class.closed_cases_correspondence_type
```

Polsonby (commented):
LGTM - just a couple tiny things to consider maybe?

app/controllers/stats_controller.rb
Polsonby: If @correspondence_types comes out as `nil`, the above code blows up anyway :) 

Could we make it the `<<` just for ease of reading?

Polsonby (approved):
Good work, LGTM :)

PR #1302, opened 2020-09-09
CT-2991 Added coverage for #create action and steps validation

Polsonby:
## Description
We encountered bugs in final testing because although we were testing the happy path through case creation for Offender SAR, we weren't testing the partial validations that it has to do on each step. This PR includes a bug fix described below but otherwise shouldn't change the behaviour of the validations

In adding test cases, I've tried to ensure that the logic in the private methods in the case form that do things like setting empty default values if a radio group is omitted or ensuring that date fields get validated on a given step are all tested, such that when you comment those lines out in the case form, the relevant specs then fail. Previously this wasn't the case! 

Bug - I've added in a check on step 5 to run the validation that the optional date requested isn't a future date there and then on that step - this validation was previously not being called during this step, and meant although the case was invalid with. future date, you didn't see the error on the step where you added the date, it just silently prevented the case from becoming valid and looped around to the start again when you got to the end. 

I discovered that one of the most hacky methods I'd added seems to be no longer required - possibly as a beneficial side effect of some of the refactoring that's happened along the way. So I've removed it - would be good to test that part thoroughly to see if I've missed something.

## Self-review checklist

* [x] (3) Tests passing

### Screenshots

Now, if you enter the optional `request dated` date on the request details step, and add a future date, you see the validation message there and then (as opposed to never 🤦): 
image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2991

### Deployment
n/a

Verify that case creation for offender SAR still works as intended
You can test the partial validations on each step of the offender SAR case creation journey submitting empty fields - the requester details and recipient details should require one of the two radios to be chosen, then require appropriate validations for the third party fields if that's one of the options selected. Please play around with the `requester-details` step in particular because I've removed my horribly- and incorrectly-named `clear_param_if_condition` methods :)


spec/controllers/cases/offender_sar_controller_spec.rb
```diff
@@ -68,6 +68,159 @@
...
+          }
+        end
+
+        it 'validates subject name and address' do
+          errors = assigns(:case).errors.messages
+          expect(errors[:subject_full_name]).to eq ["can't be blank"]
+          expect(errors[:subject_address]).to eq ["can't be blank"]
+        end
+
+        it 'sets empty values and validates' do
```
willmcb, line 97: I find this wording slightly confusing - is this not just testing that these things are validated? Or are you saying "sets" due to the controller work arounds for dates and things? :)

spec/controllers/cases/offender_sar_controller_spec.rb
```diff
@@ -68,6 +68,159 @@
...
+                third_party_name: '',
+                third_party_company_name: '',
+                postal_address: '',
+              }
+            }
+          end
+
+          it 'validates requester details' do
+            errors = assigns(:case).errors.messages
+            expect(errors[:third_party_name]).to eq ["can't be blank if company name not given"]
```
willmcb, line 141: Not a big thing but we could wrap these expects in a method as they are repeated on lines `208-211`. Method could be something like:

`expect_third_party_details_to_be_correct`

willmcb (commented):
I like this approach it means that we can test the validations specifically for each step partial, which have more baggage than the model validations as the rely on the 'step-mechanism'.

Two things:

1.
- it might be worth adding similar specs for the steps that were implemented outside of the main flow:
https://github.com/ministryofjustice/correspondence_tool_staff/blob/master/app/views/cases/offender_sar/_final_page_count_step.html.slim 

and

- https://github.com/ministryofjustice/correspondence_tool_staff/blob/master/app/views/cases/offender_sar/_date_responded_step.html.slim

Though they are not part of the 'steps' flow they basically do the same thing interms of validations so it might be worth testing them here. Happy to add these myself later if you'd rather just concentrate on the 'flow' steps for now.

2.
I wonder if there is a way to wrap these validations in a loop and expect things more dynamically? Haven't given it too much thought but by the shape of the code it looks like it could become "iterable", where we feed in the step value and it then checks in a loop? Not sure it would be any simpler but the thought crossed my mind...

Happy to chat about any of this if you'd like 👍

spec/controllers/cases/offender_sar_controller_spec.rb
Polsonby: thanks, had a go at clarifying as discussed, has it helped?

spec/controllers/cases/offender_sar_controller_spec.rb
Polsonby: Done, thanks!

danielatdigital (approved):

PR #1303, opened 2020-09-10
CT-2679  not be able to save the upload files for ico case type

ymao2:
## Description

This PR is to fix the bug the uploaded files cannot be saved for ico case type

The reason I found was that the name being used for collecting the files in form was 'case_ico' which is the key, 'ico',  being used in the controller, so the value was always skipped by the backend which caused this bug

## Self-review checklist

* [X] (3) Tests passing

### Screenshots

Before:  uploaded files cannot be saved 
After
Screenshot 2020-09-10 at 14 43 25

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2679
### Deployment

Not easy to test it locally (at least I couldn't mange to do it, I tested on staging env by deploying this version

1. Login as London use
2. Create a ICO case and upload some files 
3. Once the whole creation journey is completed, you should see those files at case detail page


Polsonby (approved):
Amazing 😮 - great detective work 👏 👏 👏 👏

PR #1306, opened 2020-09-11
CT-2982 Ensure third party states consistent

Polsonby:
Offender SAR should never have both `third_party` requester AND `third_party_recipient` but you can potentially get a case into this state. 

This PR adds validation to prevent a case being in this state but because the options are on two different screens, just
the validation alone results in a poor user experience, so if it happens we also implement a before filter to nudge recipient to the right option, `requester_recipient` and ensure consistency without breaking the user experience

## Description
THE PROBLEM
So, let's say you have an Offender SAR case with subject requester and third party recipient
image

Edit it to make it requested by someone else on step 2 - now you see two lots of details. THIS SHOULD NEVER HAPPEN

image

Step 1 - add validation so you can't change it to third party if third_party_recipient is selected... 

image

It's a terrible message and it's not obvious how to fix it - you can't even fix it on this screen! So we need to make sure people never see this message. Therefore - 

Step 2 - We also add a before_validation which silently nudges recipient to the right value, given the user's intention

image

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
see above!

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2982

### Deployment
n/a

As above!
Create an offender SAR case with subject requester and third party recipient
Edit it to select "Requested on someone elses' behalf"
Note that it works AND the recipient is now "requester", as it would be had they created it this way from the outset.


ymao2:
look good to me, I approve it

ymao2 (approved):

PR #1314, opened 2020-09-21
CT-2955  Circle ci slack integration

digitalali-moj:
## Description
Integrate Circle CI with Slack so that:
- When someone deploys it should send a message with the image tag in the product-specific channel
- When someone deploys to production it should send a message to the channel #staff-services-circleci with the tag.
 
**Note**: It's not possible to send failure notifications for deployment without significant refactoring of the code into an orb, or adding the WHEN clause to every job which will get messy. If it's something that we feel is missing then I can do that, but I'd recommend that it be a second iteration, as it will make the code go very long and might not even be needed in the end. At the moment, the lack of a success message means that there's a failure. The dev will also get an email. And the case of a deploy failing is very rare unless actually modifying the deployment script.

Also "mention:" key is putting @ with team name, but doesn't seem to knock knock on slack for me, maybe it's not really doing anything. Might remove.

Fake sample workflow: https://app.circleci.com/pipelines/github/ministryofjustice/correspondence_tool_staff/1754/workflows/839ced10-459a-4865-aecf-0de998b6d707 (comms without actual deployments)

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-2955

### Deployment
This affects builds and deployments, so should be tested before being merged. 

Do a build, we should get a message in Staff-services-tech channel with the build tag and workflow URL.


Polsonby (approved):
LGTM :)

PR #1315, opened 2020-09-22
Ct 2948 2949 filter issues

ymao2:
## Description

Firstly, don't be scared by the number of files changed,  nearly half of them are due to the location changes or removal
This is PR is to fix the fundamental issues below around filters side 
1 - Displayed irrelevant filter e.g. sensitivity flag and choices under some filters e.g.  offender sar states for London within each filter based on the user's permissions 
2 - Some of key filters didn't work e.g. status filter for offender SAR type
3 - The filter on "my open cases" doesn't match others
4 - requirement for adding some filters for "Closed cases" tab
5 - From coding practise, many codes are unnecessary duplicated

In order by fixing the above issues and giving possibility for improvment
- Introduced the permission check for filter level and choices level 
- Re-structured the filter classes ( didn't change the core part of each filter)
   -- Introduced 2 extra super classes for date filter and multi choices filter into filter classes (for OO practise)
   -- Moved all those filter classes files into separated folder 'case_filter' module
   -- Introduced 2 view templates to date filter and multi-choices filter and removed all those "duplicated" ones
   -- Linked the filters with the page tab to make adding new filter much easier 
- Read the permitted states for each case type from yaml file rather than hard code 
- Added a few filters into Closed reports: received_date, date_responded, case types, exemptions (only for non-offender-sar)
- A simple solution for downloading the closed report

## Self-review checklist

* [ ] (3) Tests passing
* [] (6) No TODO's without new ticket numbers

### Screenshots

**Before**  
**Branson user**
- Open cases 
Screenshot 2020-09-22 at 17 55 23
Screenshot 2020-09-22 at 17 55 28
- my Open
Screenshot 2020-09-22 at 17 55 36
- Closed cases 
Screenshot 2020-09-22 at 17 55 48

**After** 
- Open cases
Screenshot 2020-09-22 at 17 59 10
Screenshot 2020-09-22 at 17 59 41
- My open
Screenshot 2020-09-22 at 17 59 51
- Closed cases
Screenshot 2020-09-22 at 18 00 23
Screenshot 2020-09-22 at 18 00 30

**London User**
After
- Open
Screenshot 2020-09-22 at 18 05 20

Screenshot 2020-09-22 at 18 05 25

- Closed cases tab
Screenshot 2020-09-22 at 18 05 44

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2948
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2479


app/controllers/cases/filters_controller.rb
```diff
@@ -33,10 +33,11 @@ def closed
       service = call_search_service(unpaginated_cases)
       @query = service.query
 
+      @maximum_records_for_download = 1000
       if service.error?
         flash.now[:alert] = service.error_message
       else
-        if download_csv_request?
+        if download_csv_request? and service.result_set.count < @maximum_records_for_download
```
Polsonby, line 39: Tiny thing :) Probably should be `&&` instead of `and` here - unless you're intentionally using it for flow control e.g. https://www.synopsys.com/blogs/software-security/ruby-demystified-and-vs/

app/services/case_filter/exemption_filter.rb
```diff
@@ -0,0 +1,90 @@
+module CaseFilter
+  class ExemptionFilter < CaseMultiChoicesFilterBase
+
+    def self.identify
```
willmcb, line 4: Here it may be better to use `class << self` and bunch all the class methods together as there are a few of them in this class.

app/services/case_filter/case_date_range_filter_base.rb
```diff
@@ -0,0 +1,96 @@
+module CaseFilter
+  class CaseDateRangeFilterBase < CaseFilterBase
+
```
willmcb, line 3: As per the `CaseMultiChoicesFilterBase` would be good to discuss adding this via a mixin rather than a class :)

app/services/case_filter/case_type_filter.rb
```diff
@@ -0,0 +1,92 @@
...
+    }.with_indifferent_access.freeze
+
+    class << self
+      def filter_attributes
+        [:filter_case_type]
+      end    
+
+    end
+
+    def get_available_choices
```
willmcb, line 22: Here again it would be nice to match the interface of either `available_choices` or `get_available_choices`

app/services/case_filter/case_status_filter.rb
```diff
@@ -0,0 +1,59 @@
...
+          {'open'   => I18n.t('filters.statuses.open'),
+          'closed' => I18n.t('filters.statuses.closed')}
+      }
+    end
+
+    def call
+      filter_status(@records)
+    end
+
+    def crumbs
```
willmcb, line 20: I think we may be able to extract `def crumbs` to a class lower down the hierarchy. Seems to be of a similar shape in all the subclasses and relies on a method being called on "`@query`" - I think we could add some polymorphism here by passing in the differing method as a symbol. Or decorating the query object and returning the correct data via the same interface at a per class level.

app/state_machines/configurable_state_machine/machine.rb
```diff
@@ -8,6 +8,10 @@ def initialize(config:, kase:)
       @events = nil
     end
 
+    # Ideal there shouldn't be any other definitions of states other than state machine yaml file. but 
+    # tried to remove it with a method in manager and ended up with an error below 
+    # >> Circular dependency detected while autoloading constant in one of policy ruby code
+    # Couldn't think of easy way to fix it, so just added missing states here
```
willmcb, line 14: Interesting - I think this would be good to investigate - perhaps we should pair on it 🤔

app/views/cases/filters/index.html.slim
```diff
@@ -14,18 +14,14 @@
 = render partial: 'cases/shared/search_bar'
 
 = render partial: 'cases/shared/case_tabs'
-
 section.govuk-tabs__panel
-
-  - if request.path.include? '/cases/open'
+  - if !@query.available_filters(current_user, @current_tab_name).empty?
```
willmcb, line 18: I think we could loose the negation here if we instead used `.present?`. I find double negatives hard to parse.

app/views/cases/search_filters/_filters.html.slim
```diff
@@ -8,35 +8,11 @@
...
-          summary aria-controls="filter-deadline-content" aria-expanded="false" role="button"
-            span.case-filters__summary
-              = t('filters.headings.filter_by_deadline')
-          .case-filters__content#filter-deadline-content aria-hidden="true"
-            = render partial: 'cases/search_filters/final_deadline', locals: { f: f }
-        - if request.path.include? '/cases/search'
+        - @query.available_filters(current_user, @current_tab_name).each do | filter |
           details
-            summary aria-controls="filter-exemptions-content" aria-expanded="false" role="button"
+            summary aria-controls="filter-#{}-content" aria-expanded="false" role="button"
```
willmcb, line 13: Is something meant to be interpolated here in `filter-#{}-content`? looks like an object to interpolate is missing to me 🤔

willmcb (commented):
Good job I have concentrated on the application code rather than the specs at the moment as I know you said you are still working on the feature 👍 

Let me know if you'd like to discuss any of my comments here! Happy to re-review once it's ready :)

app/services/case_filter/case_multi_choices_filter_base.rb
```diff
@@ -0,0 +1,31 @@
+module CaseFilter
+  class CaseMultiChoicesFilterBase < CaseFilterBase
```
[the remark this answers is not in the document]
ymao2: IMO,  class (sub-class) is the thing I aim for,  the structure is
Filter Sets --- SearchQuery
  - filter element (primarily I would suggest based on the type of filter -- or I would say the nature of filter), such as 
      - Date type 
      - Multi choice
      - Integer
      - Single Choice
     etc.
The interface to outside world will be "same"
    - UI related part:  so we have template, element choices
    - Scope outcome: which is the scope thing
    - other bits...
my intention is to
 - keep each individual filter simply enough ( mixing Date and Multi-choice within one filter component seems unnecessary complicated to me especially once you consider permission level, plus if the filter is different type, then it must means different meaning, doesn't make sense to group them into one UI for any reason other than saving space for UI?)
 - include most common one -> so we don't need to duplicate codes
 - but also allow "special" being defined so easily within this framework

Mixins based on my understanding are more like the extended "feature"/"behaviour" (not core behaviour) of you would like to add to other objects ( no matter which type object is , like Audit, permission check etc.)---> which doesn't fit in the above framework

We can discuss more if you want to :)

app/views/cases/search_filters/_filters.html.slim
ymao2: mistake, didn't cause any issue for running the app, but caused the failure of test spec

app/services/case_filter/case_filter_base.rb
```diff
@@ -0,0 +1,65 @@
+module CaseFilter
+  class CaseFilterBase
```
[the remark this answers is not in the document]
ymao2: whether needing a pure abstract class is arguable,  depending on how those sub-classes are called from "ousider", it is more suitable if you want to make the "outsider" to only know interface, but has no responsibility for creating the instance of sub-classes and maintaining the lifecycle of the objects,  some design patterns like Adapter pattern , Poxy etc. ---->(actually I am quite like Adapter pattern, used it relevant often), But for our example,  there is no expose requirement, it is managed by SearchQuery (---> behave like FilterSet,  some confusion from previous designer for this part), I would think it is more close to the pattern being used quite often for WeForm as each filer is combination of look-and-feel plus + the behaviour  ( + outcome)
  FormFieldSet 
     --  BooleanField 
     -- MultipleChoiceField
but possible with higher granularity.

app/services/case_filter/case_multi_choices_filter_base.rb
```diff
@@ -6,7 +6,7 @@ def identify
         if filter_attributes.empty?
           raise '#dentify should be defined in sub-class of CaseMultiChoicesFilterBase'
         else
-          filter_attributes[0]
+          filter_attributes[0].to_s
         end
       end
```
Polsonby, line 12: Tiny typo here. "#dentify" instead of "#identify" :) 

there's a Ruby style thing about guard clauses (see https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/GuardClause) - 

Wondering is the message for the exception right? I'm probably missing something but it looks like maybe it should say something like 

```suggestion
def identify
  raise 'filter_attributes should be defined in sub-class of CaseMultiChoicesFilterBase' if filter_attributes.empty?
  filter_attributes.first.to_s
end
```

ymao2: haven't got used to 'ruby' feature --> offering the 'first' , 'second' which is quite useful

Polsonby (approved):
Great work :)

PR #1323, opened 2020-09-24
CT-3107 Add Complaint correspondence type

Polsonby:
## Description
Spike to implement first pass at new Correspondence Type for Complaints - this PR is experimenting with subclassing Case::SAR::Offender and related classes

It does the smallest amount of possible work to add the Correspondence Type model, state machine workflow, and get the model spec passing; other specs included where necessary to achieve this.

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3142

### Deployment
The correspondence type seeder has been updated but in the next PR we'll add a rake task to run it in different environments.

None really - as long as the specs pass and no side-effects elsewhere, should be good


spec/models/case/sar/offender_complaint_spec.rb
```diff
@@ -0,0 +1,703 @@
...
+        subject: 'B',
+        subject_address: 'B',
+        subject_aliases: 'B',
+        third_party_company_name: 'B',
+        third_party_name: 'B',
+      })
+    end
+  end
+
+  # describe 'deadline' do
```
willmcb, line 460: I guess we can lose this commented out stuff for this new type 😂

app/models/case/sar/offender_complaint.rb
```diff
@@ -0,0 +1,306 @@
...
+                            message: 'must be a positive whole number' }
+  validate :validate_third_party_states_consistent
+
+  before_validation :ensure_third_party_states_consistent
+  before_validation :reassign_gov_uk_dates
+  before_save :set_subject
+  before_save :use_subject_as_requester,
+              if: -> { name.blank? }
+
+  def validate_third_party_states_consistent
```
willmcb, line 125: I think a pragmatic thing we could do in this new type is to extract these validations out into `app/validators` using `ActiveModel::EachValidator` as a base class. Might be able to DRY things out a little with OffenderSars as well with this.

spec/factories/case/offender_sar_complaints.rb
```diff
@@ -0,0 +1,143 @@
+FactoryBot.define do
+
+  factory :offender_sar_complaint, class: Case::SAR::OffenderComplaint do
```
willmcb, line 3: This got me thinking - I wonder if it would be nicer to have `Case::SAR::Offender::Complaint` rather than the `OffenderComplaint` class. 

What got me thinking this way is that we have a variable `offender_sar_complaint` and the class `OffenderComplaint` and it seems like there is a bit of a word order jumble there. Not sure if this helps but I thought I'd ask now as we are just starting 👍

spec/factories/case/offender_sar_complaints.rb
```diff
@@ -0,0 +1,143 @@
...
+      third_party_relationship { 'Solicitor' }
+      requester_reference { 'FOOG1234' }
+      third_party_company_name { 'Foogle and Sons Solicitors at Law' }
+      third_party_name { 'Mr J. Smith' }
+      postal_address { '22 High Street' }
+      recipient { 'requester_recipient' }
+    end
+
+    trait :data_to_be_requested do
+      # Default state for a new offender_sar_case
```
willmcb, line 54: We should amend or remove this comment :)

spec/models/data_request_spec.rb
```diff
@@ -133,17 +133,6 @@
     end
   end
 
-  describe '#case' do
-    subject(:data_request) { build :data_request }
-
-    it { should be_valid }
-
-    it 'is restricted to Offender SAR at present' do
```
willmcb, line 141: Might be worth thinking about this in an additive way once we have the tests passing i.e. "It is restricted to Offender SAR and Offender Sar Complaints at present"

spec/models/case/sar/offender_complaint_spec.rb
```diff
@@ -0,0 +1,703 @@
...
+        kase = build :offender_sar_complaint, third_party: false, third_party_relationship: '',
+                      recipient: 'subject_recipient'
+        expect(kase).to be_valid
+      end
+    end
+  end
+
+  describe '#subject' do
+    it 'is the same as subject full name' do
+      kase = create :offender_sar_complaint
```
willmcb, line 265: Another possible small refactor here - I think a lot of these could be `build_stubbed` rather than create as we are testing in memory object validations.

spec/factories/case/offender_sar_complaints.rb
Polsonby: Great suggestion 👍

ymao2 (approved):

PR #1324, opened 2020-09-24
CT-3116 Add extra fields to the closed-report for FOI

ymao2:
## Description

This PR is to make sure the following fields to have correct value and will be shown on closed-reports for FOI (business days)
- num_days_taken
- num_days_late
- num_days_taken_after_extention
and those fields will map to the following keys in warehouse table 
- number_of_days_taken
- number_of_days_late
- number_of_days_taken_after_extension

And because there have been case being syned into warehouse table in the past,  in order to refresh the values in old cases
2 extra cts command have been added to allow generating the CaseSyncJob manually based on the options below
- id range of cases
- case number

## Self-review checklist

* [X] (3) Tests passing

### Screenshots

Before 
Screenshot 2020-09-25 at 12 29 38
After 
Screenshot 2020-09-25 at 12 18 22

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3116
### Deployment

please run db migration to add new fields into warehouse table
rails db:migrate

1 - Login as London user
2.- Click 'Reports' from top menu
3 - Click 'Create report' button
4 - Choose 'General closed cases report'
5 - Put proper time frame
6 - Click 'Create custom report'
7 - Download the report (in zip format, need to uncompress it) and check the data


app/models/case/base.rb
```diff
@@ -962,5 +974,21 @@ def validate_case_link(type, linked_case, attribute)
...
+      .where(event: ['extend_for_pit', 'extend_sar_deadline'])
+      .order(id: :desc)
+      .first
+      most_recent_extension.nil? ? nil : most_recent_extension.created_at.to_date
+    else
+      nil
+    end
+  end
+
+  def potential_include_extension?
```
Polsonby, line 989: I find this method name a bit opaque, not clear what the concept is here :) - wondering if we can consider changing it to make it a bit more self-documenting?

app/models/case/base.rb
```diff
@@ -612,15 +612,31 @@ def already_late?
   end
 
   def num_days_late
-    days = ((date_responded.nil? ? Date.today : date_responded) - external_deadline).to_i
+    days = @deadline_calculator.class.days_late(
+      external_deadline, 
+      (date_responded.nil? ? Date.today : date_responded))
```
Polsonby, line 617: This concept appears 5 times; wondering if it's a candidate for making into a wee method to name the concept and dry it up?

`(date_responded.nil? ? Date.today : date_responded)`

app/models/case/base.rb
ymao2: I am thinking I will remove this func and only check the flag of has_pit_extension as this is London user needs,  I was thinking maybe it is nice to check 'SAR' case too---> now I feel I shouldn't really think about SAR case type as it maybe be managed differently.

Polsonby (approved):
LGTM :) - great work 👍

PR #1338, opened 2020-10-06
CT-3039 showing deleted case in case transitions

ymao2:
## Description

This PR is to make the case-history displayable after a historic linked case is deleted, so changes I made is to make the whole story around the deletion side completed
- Check whether there are some linked cases before deletions
- Show different wording based on whether the historic linked case has been deleted or not
 
## Self-review checklist

* [X] (3) Tests passing

### Screenshots

Before 
Grey death screen

After  
1 - Extra checking when a case is deleted 
error_message_due_to_linked_to_a_case

2 - For the case which has a linked case, the case history will look like below if the linked case hasn't been deleted
updated_case_history_in_the_linked_case

3- After the linked case is deleted, the case history will look like
updated_case_history_after_deletion

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3039

1 - Login as London user
2 - Pick up an example FOI/SAR case 
3 - Pick up an case to link to the case in step 2 
4 - Load the case detail for the case in step 2
5 - Click "Link a case" in linked cases section and type the number of case from step 3, then click 'Add link'
6 - Check case details for the case in step 3 and will see there is new entry in case history from step 5
7 - Come back to the cases in step 2 and choose "Delete case"
8 - Click "Delete", an error message will pop up saying the related case links are not empty, you need to remove link first 
9 - Following the instruction, click 'Cancel' to come back to the case detail
10- Remove the linked case by clicking 'Remove link' from the linked cases section
11 - Check the case history again for case in step 3, and a new entry should be added from step 10 and it shows the number of case in step 2 
12 - Delete the case from step2 by following step 8 
13 - Check the case history again for case in step 3, and the wording for the entry from step 11 should be changed to show the id of case only plus some extra note


app/views/cases/confirm_destroy.html.slim
```diff
@@ -23,7 +23,7 @@
         = @case.subject
 
     p
-      = f.text_area :reason_for_deletion
+      = f.text_area :reason_for_deletion, @case.reason_for_deletion
```
Polsonby, line 26: I don't think you need to pass the current value for the reason on to the method here, it should automatically populate the field even without this change :)

Polsonby (approved):
Great PR, thanks!

app/views/cases/confirm_destroy.html.slim
ymao2: you are right tried 👍

PR #1342, opened 2020-10-07
CT-3041 Prevent double click error for state action button 

ymao2:
## Description

This PR is to 
- mainly prevent user from clicking state action button multiple times
- tidy up the code a little bit by removing unnecessary/wrong duplications of scope of allowed-actions, removing unused identical triggered flow for offender SAR

Initially, I was seeking to change the backend to ignore such multiple clicks, but found out no matter which way I chose (e.g.,   e.g. only filter out the errors if the current event is same as  most recent event one, or if current state is same as the target state ) it will make the rules of validating the event based on the permission of user and the definition of workflow compromised in wrong way,  so I think the better way to do this is to prevent such undesired user interaction from front-end

For testing, I tried so many way and did some research, couldn't find a way to simulate fast clicks from rspec,  so I suggest to leave it this out of test unless someone has better solution to test it. 

## Self-review checklist

* [X] (3) Tests passing

### Screenshots

Before 
double clicking a state action button causes grey death screen 
After
The second click will be ignore from front-end

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3041

1 - Login as Branson user
2 - Create a case
3 - fast click at any state action button multiple times
4 - the action should be performed correctly


Polsonby (approved):
Great solution 👍 nice work!

PR #1344, opened 2020-10-08
CT-3113 Split prison number in method (to show only one) and apply in letter templates

digitalali-moj:
## Description
Multiple prison numbers on cases need to show up on letter templates as only the first one in the list. This has been split by Comma and Spaces in between. (Commas converted to spaces and then split), then returning only the first one.

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
image
image
image
image

=== Another CASE, DISPATCH LETTERS ===

image
image
image
image
image
image
Word example:
image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3113

### Deployment
Need to run letter template seeder for each deployment manually

rake db:seed:dev:letter_templates
Create cases with multiple prison numbers and different address / recipient settings.
Go to each letter template
Ensure that wherever there is a Prison number, it only shows the first one


app/models/case/sar/offender.rb
```diff
@@ -283,6 +283,11 @@ def data_requests_completed?
     end
   end
 
+  def first_prison_number
+    return '' unless prison_number.present?
+    prison_number.gsub(/[,]/, ' ').split(' ').first
+  end
```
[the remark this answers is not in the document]
willmcb: Ah is it that the string can have spaces or commas? If so there are some other ways of doing this: 

for instance:

```ruby
prison_number.split(/[\s,]/)
```

More options here:
https://stackoverflow.com/questions/19509307/split-string-by-multiple-delimiters/19509391

db/seeders/letter_template_seeder.rb
```diff
@@ -92,7 +92,7 @@ def seed!
                   <p>
                   <br>
                   <br><strong>DATA PROTECTION ACT 2018: SUBJECT ACCESS REQUEST</strong>
-                  <br><strong><%= values.subject_full_name&.upcase %> - <%= values.prison_number&.upcase %><% if (values.prison_number.gsub(/[,]/, ' ').squeeze(' ').strip).match?(/[ ]/) %> [DELETE AS APPROPRIATE]<% end %></strong>
+                  <br><strong><%= values.subject_full_name&.upcase %> - <%= values.first_prison_number&.upcase %></strong>
```
willmcb, line 95: Why not `upcase` them in the method? would remove the need for the `&` syntax as we check for presence already there.

db/seeders/letter_template_seeder.rb
```diff
@@ -4,7 +4,7 @@ def seed!
     puts "---- Seeding Letter Templates ----"
     # Reciever for prision related string
     prison_receiver = <<~EOF
-        <%= letter.name %><% if values.prison_number.present? %> - <%= values.prison_number %><% if (values.prison_number.gsub(/[,]/, ' ').squeeze(' ').strip).match?(/[ ]/) %> [DELETE AS APPROPRIATE]<% end %><% end %>
+        <%= letter.name %><% if values.prison_number.present? %> - <%= values.first_prison_number %><% end %>
```
willmcb, line 7: Why do we `upcase` everywhere else other than here? Seems like it would be better to be uniform with this 🤔

spec/models/case/sar/offender_spec.rb
```diff
@@ -777,5 +777,43 @@
...
+    context 'When there is just one prison number' do
+      it 'returns the prison number' do
+        expect(kase.first_prison_number).to eq '123465'
+      end
+    end
+
+    context 'When there is more than one prison number' do
+      let(:kase)  { create :offender_sar_case, prison_number: 'A12345, B98765' }
+      let(:kase1)  { create :offender_sar_case, prison_number: '  A12345  , B98765 ' }
+      let(:kase2)  { create :offender_sar_case, prison_number: 'A12345 B98765' }
```
willmcb, line 793: What about `A12345, B98765 A234667`, is this a possible string? (i.e. mix of commas and spaces)

willmcb (commented):
Good job, please see comments 👍

app/models/case/sar/offender.rb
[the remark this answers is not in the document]
Polsonby: Yeah, nice - had a play though, none of those options do quite the right thing for our examples. I think it's fine :)

db/seeders/letter_template_seeder.rb
Polsonby: Good call, have amended the method

willmcb (approved):
👍

PR #1345, opened 2020-10-08
CT-3143 Case creation for offender SAR complaint

Polsonby:
## Description
This PR implements the case creation flow and display (show page) for Offender SAR complaint, including the controller, views, site prism files and so on 

It also adds a new task to run the correspondence type seeder in use; this should be safe to run on production environments as it does find or create but please pay extra attention to this to verify no unwanted side effects.

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots

image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3143

### Deployment
Run the correspondence type seeder (and restart the app pods? @ymao2 is this right?)

`rails db:seed:dev:correspondence_types`

Create a new Offender SAR case and verify that the details display as desired on the Case Show page


app/controllers/cases/offender_sar_complaint_controller.rb
```diff
@@ -0,0 +1,53 @@
+module Cases
+  class OffenderSarComplaintController < OffenderSarController
+    include OffenderSARComplaintCasesParams
+
+    def initialize
```
willmcb, line 5: May be more Railsy to use a before filter here rather than `initialize`:

https://stackoverflow.com/questions/18576150/ruby-on-rails-before-filter-vs-rubys-initialize

config/routes.rb
```diff
@@ -80,6 +81,16 @@
       end
     end
 
+    resources :offender_sar_complaints, only: only, controller: 'offender_sar_complaint', as: :case_sar_offender_complaint do
```
willmcb, line 84: ` controller: 'offender_sar_complaint', as: :case_sar_offender_complaint` why use two separate but similar names? I think it may be better to just stick to either one.

Polsonby: I see your point 🤔 

The `as: ... ` routes are set up for namespacing consistency across the correspondence types, by the look of things - but with offender_sar and now with the complaint we've not really followed the same pattern as the rest of them - see `rake routes | grep case_`  

Right now consistency with Offender SAR is the North Star for this work, but we can take a look at it in the next phase maybe?

app/controllers/cases/offender_sar_complaint_controller.rb
Polsonby: That was an interesting rabbit hole... 😁 It turns out you need to call `super` within the initialize method to make the view render inside the layout properly - not sure why! But all the correspondence types follow this pattern. You can omit the other instance vars in the initialize method and the new complaint tests still pass, but I've left it as-is for consistency with all the other controllers for now, maybe one day we'll find out why it's necessary.

willmcb (approved):
Nice work 👍

PR #1350, opened 2020-10-12
Change delete/create to apply for cronjob to avoid breaking prod

digitalali-moj:
## Description
Deleting Cronjob was forbidden on Circle CI - this can be investigated, but might not be needed.
Changed to Apply first, so Cronjob changes are applied directly to existing Cronjob. If we find this doesn't update the Cronjob correctly then we can investigate further on deleting.

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
(This was going to be part of another ticket, but that ticket got archived)

### Deployment
Requires deployment through to production.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -136,8 +136,7 @@ function _deploy() {
 
   if [ $environment == "production" ]
     then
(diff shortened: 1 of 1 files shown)


Polsonby (approved):
Cool! :) 👍 fingers crossed

PR #1364, opened 2020-10-23
CT-3215 Ensure assignments are present before trying to build reassign link

Polsonby:
## Description
After moving people/business units around in the CMS and deactivating the "Private Office" team this resulted in some users being unable to view a large number of cases.

On investigation, one problem seemed to be the deactivation of the private office team, which has a specific value for "Code" that had been amended as part of the deactivation. 

We restored the missing team, making no other changes to the database. This restored the ability to access some of the cases but other cases failed with a different problem now. 

[09120983cf00da9c7d7e94c6c5c62bff] ActionView::Template::Error (No route matches {:action=>"reassign_user", :case_id=>"20390", :controller=>"assignments", :id=>nil}, missing required keys: [:id]):

This happens in the cases helper for reassign user when the user has no assignments. The helper assumes there will always be at least one assignment or more, but in this case on production it seems there are none and the helper requires one. This PR adds a guard clause to ensure there is always at least one and return an empty string if not.

This doesn't address the root cause, and there may be further errors. Requires further investigation. 

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
n/a

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3215

### Deployment
n/a

n/a

### DIFF
--- a/app/helpers/cases_helper.rb
+++ b/app/helpers/cases_helper.rb
@@ -73,6 +73,7 @@ module CasesHelper #rubocop:disable Metrics/ModuleLength
               class: 'button'
     when :reassign_user
       path = nil
+      return '' unless @assignments.present?
       if @assignments.size > 1
         path = select_team_case_assignments_path(@case, assignment_ids: @assignments.map(&:id).join('+'))
       else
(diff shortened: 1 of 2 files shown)


willmcb (approved):
👍

PR #1365, opened 2020-10-26
CT-3216 Restore previous team members properly if there is a bad user role

Polsonby:
## Description
Carry on without breaking the transaction if there is a bad user role

When the team being moved has invalid team roles (e.g where user_id is nil in teams_users_roles) the team move service failed to properly restore the existing users to the deactivated team after those users are moved to the new incarnation. This causes authentication errors on cases wired up to the old team for any users whose team memberships weren't properly copied backwards. 

The service now ignores invalid rows in team_user_roles and restores all valid users for old team, where previously it stopped if one was invalid.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3216

### Deployment
We will need to run a rake task to fix affected users on production

n/a

### DIFF
--- a/app/services/team_move_service.rb
+++ b/app/services/team_move_service.rb
@@ -103,8 +103,12 @@ class TeamMoveService
   end
 
   def restore_users_for_old_team
-    @keep_user_roles.each do |ur|
-      TeamsUsersRole.create!(team: Team.find(ur[0]), user: User.find(ur[1]), role: ur[2])
+    @keep_user_roles.each do |team_id, user_id, role|
+      team =  Team.find_by_id(team_id)
+      user = User.find_by_id(user_id)
+      if team && user
+        TeamsUsersRole.create!(team: team, user: user, role: role)
+      end
     end
   end
 end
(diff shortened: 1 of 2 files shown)


app/services/team_move_service.rb
```diff
@@ -103,8 +103,12 @@ def restore_new_team_name_to_original_name
...
 
   def restore_users_for_old_team
-    @keep_user_roles.each do |ur|
-      TeamsUsersRole.create!(team: Team.find(ur[0]), user: User.find(ur[1]), role: ur[2])
+    @keep_user_roles.each do |team_id, user_id, role|
+      team =  Team.find_by_id(team_id)
+      user = User.find_by_id(user_id)
+      if team && user
+        TeamsUsersRole.create!(team: team, user: user, role: role)
+      end
```
ymao2, line 111: @Polsonby,  just being a little picky,  I was wondering whether we need to have a way to throw/log such errors,  but I am happy to leave it like the current way

ymao2:
LGTM 👍

ymao2 (approved):

app/services/team_move_service.rb
Polsonby: Great point... going to merge just now, but will have a think about this more generally - thanks!

PR #1372, opened 2020-10-29
CT-3220 excluded deactivated team from approving team

ymao2:
## Description

This PR is to fix the permission issue of Will and Kashan on triggered FOI cases as approver in Disclosure team due to the partial of failure of team movement. 
The issue we found was that the approving_team derived from those users returned the old team, the deactivated Disclosure team, not the new one, although they both are perfectly in the new disclosure team too.  we discovered the assumption for returning this approving team from user before this fix was 
1 - The approving team for a user should only have one 
2 - If the approving team gets moved (which will cause the users will be in 2 teams, one is the deactivated one, one is the new one, the new one is the copy of old new),  then top one should be the new team
Because of the partial failure and re-correction action happened afterwards,  the assumption is broken.  This fix is to exclude those deactivate ones which we believe is the correct logic it should be 

I added a new spec in user model to covering this change,   also added some extra checking in team_moving_service_spec to make sure the logic being fulfilled.  Those extra checking in the old logic will failure ( it seems to me we never really test this logic properly before)

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3220

Happy path ( not sure how to create a situation we had in live)
1 - Create a new Directorates
2 - move the disclosure team into new directorates
3 - login as approver in the disclosure team 
4 - check whether you can still see the cases under "New cases", you can take on case and assign another  team member

### DIFF
--- a/README.md
+++ b/README.md
@@ -503,7 +503,7 @@ control aspects of the build. The available build arguments are:
   to generate a PDF preview (libreoffice).
 
 #### Generating Documentation
-
+ 
 You can generate documentation for the project with:
 
 ```
--- a/app/models/teams_users_role.rb
+++ b/app/models/teams_users_role.rb
@@ -21,4 +21,5 @@ class TeamsUsersRole < ApplicationRecord
   scope :manager_roles,   -> { where(role: :manager)  }
   scope :responder_roles, -> { where(role: :responder) }
   scope :approver_roles,  -> { where(role: :approver) }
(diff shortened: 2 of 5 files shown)


Polsonby (approved):
Great job :)

PR #1376, opened 2020-11-10
CT-3172 complaints reporting

Polsonby:
## Description
This PR implements open and closed cases reports for Offender SAR Complaints as a baseline. In view of the uncertainty around MI and to get the PR into review I've ignored the monthly report for now.

It also includes some small refactoring and minor changes, including adding separate predicate methods for identifying _any_ offender sar case, a standard one, or a complaint: 
* offender_sar?                       # could be either - use this when we don't care which
* offender_sar_standard?
* offender_sar_complaint?

We also address some issues found with case creation for the new correspondence type, namely separating out the session data for in-progress complaints vs standard offender sars, also fixing a bug I found with the Back link, so that it now correctly remains within offender SAR complaint instead of "jumping tracks" back to standard offender SAR. 

## Self-review checklist

* [x] (3) Tests passing

### Screenshots

image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3172

### Deployment
Has both data and database migrations :) 

Run `rails db:migrate` and `rails data:migrate`

As Branston user: Viewing open cases list containing offender sar and offender sar complaint cases, "Download as CSV" should export the cases of both types according to the filters. (Filters bug on complaint fixed in Yikang's PR) 
Closed cases reports should contain only offender sar or offender sar complaints respectively


willmcb:
Interesting in regards to the predicate names. I think the method names make sense though I wonder if we may need to audit where `offender_sar?` is used currently so that we make sure we are not conflating Offender SARs with Offender SAR Complaints in places where we currently only want Offender SARS.

Part of me thinks we should just have `.offender_sar?` to mean only Offender SARs, and have `.offender_sar_complaint?` to mean Offender SAR Complaints only. If a condition with both we only need to add an `x.offender_sar? || x.offender_sar_complaint?`, it's not too much code to have to write.

What are your thoughts? @Polsonby

willmcb (approved):
👍

app/models/case/base.rb
```diff
@@ -814,6 +814,7 @@ def overturned_ico?;         false;  end
   def overturned_ico_sar?;     false;  end
   def overturned_ico_foi?;     false;  end
   def offender_sar?;           false;  end
+  def offender_sar_standard?;  false;  end
```
ymao2, line 817: maybe give more background why a new type of offender sar needs to be introduced?

app/models/case/sar/offender_complaint.rb
```diff
@@ -10,6 +10,10 @@ def offender_sar_complaint?
   end
 
   def offender_sar?
+    true
+  end
```
ymao2, line 14: that is even confusing than standard one

app/services/stats/r401_offender_sar_closed_cases_report.rb
```diff
@@ -15,13 +15,13 @@ def etl_handler
     end
 
     def case_scope
-      CaseFinderService.new(@user).closed_cases_scope.where(type: 'Case::SAR::Offender')
+      CaseFinderService.new(@user).closed_cases_scope.where("type LIKE 'Case::SAR::Offender%'")
```
ymao2, line 18: if a new report for complaint case report is introduced below,  I would think this one shouldn't be changed

app/services/stats/etl/offender_sar_complaint_closed_cases.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      def heading
+        @_heading ||= begin
+          CSV.generate_line(OffenderSarComplaintClosedCases::CSV_COLUMN_HEADINGS, force_quotes: true)
+        end
+      end
+
+    end
+  end
+end
```
ymao2, line 59: If this is just a copy of the OffenderSarClosedCases, I would suggest to simply inherit from OffenderSarClosedCases during cloning stage. 
`class OffenderSarComplaintClosedCases < OffenderSarClosedCases`

app/services/stats/r901_offender_sar_cases_report.rb
```diff
@@ -28,27 +28,27 @@ def self.description
...
     def initialize(**options)
       super(**options)
-      @case_scope = options[:case_scope] || Case::SAR::Offender.all
-    end 
+      @case_scope = options[:case_scope] || Case::Base.offender_sar.all
+    end
 
     def case_scope
-      @case_scope.where(type: 'Case::SAR::Offender')
+      @case_scope.where(type: 'Case::SAR::Offender').or(@case_scope.where(type: 'Case::SAR::OffenderComplaint'))
```
ymao2, line 35: It will cause potential problem if we use `or` in this way when the filter options are passed into this report 
I would suggest to use the following one 
`@case_scope.where(type: ["Case::SAR::Offender", "Case::SAR::OffenderComplaint"])`

app/services/stats/r901_offender_sar_cases_report.rb
```diff
@@ -28,27 +28,27 @@ def self.description
 
     def initialize(**options)
       super(**options)
-      @case_scope = options[:case_scope] || Case::SAR::Offender.all
-    end 
+      @case_scope = options[:case_scope] || Case::Base.offender_sar.all
```
ymao2, line 31: Suggest to use the `where(type: ["Case::SAR::Offender", "Case::SAR::OffenderComplaint"])`  instead of changing the scope of offender_sar flag in case base model

app/services/stats/r402_offender_sar_complaint_closed_cases_report.rb
```diff
@@ -0,0 +1,27 @@
...
+      CaseFinderService.new(@user).closed_cases_scope.where(type: 'Case::SAR::OffenderComplaint')
+    end
+
+    def report_type
+      ReportType.r402
+    end
+
+  end
+end
+
```
ymao2, line 27: @Polsonby In order to get this report displayable under custom report/<case type>, we need to update some flags and some fields in report_type and correspondence_types table but depending on what your thought about where you want to put this report, maybe let us have a chat?

ymao2 (changes requested):

app/views/cases/show.html.slim
```diff
@@ -80,15 +80,15 @@ div id="case-#{@correspondence_type_key}" class="case"
...
+  - if !@case.offender_sar?
     section#messages-section.case-info
       = render partial: 'cases/case_messages', locals: { case_details: @case }
 
   - if @case_transitions.any?
     section.case-info.has-page-break
       = render partial: 'cases/case_history', locals: { case_transitions: @case_transitions }
 
-  - if @case.offender_sar? || @case.offender_sar_complaint?
+  - if @case.offender_sar?
```
ymao2, line 91: @Polsonby ,  I am wondering this line, if the offender_sar still means standard one not complaint one, you may need to change this flag to the new flag e.g. `offender_sar_any?` otherwise, it may not work for complaint case?

willmcb:
I have a blank DB and created an Offender SAR and  two complaints - can see them fine in the report.

One thing though there is no column for the Offender SAR Type in the report - this may be more @ymao2's area of expertise but I would have thought this would be a useful thing to have for Branston. Otherwise there is no way of telling the two types apart for divvying out work etc.

willmcb (approved):
Few small things and a question re the report - otherwise looks good 👍

ymao2 (approved):

app/controllers/cases/offender_sar_controller.rb
```diff
@@ -149,13 +149,17 @@ def build_case_from_session(correspondence_type)
     end
 
     def session_persist_state(params)
-      session[:offender_sar_state] ||= {}
+      session[session_state] ||= {}
       params ||= {}
-      session[:offender_sar_state] = session[:offender_sar_state].merge params
+      session[session_state] = session[session_state].merge params
```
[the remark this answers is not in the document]
Polsonby: Great tip, thanks!

app/controllers/cases/offender_sar_complaint_controller.rb
```diff
@@ -61,5 +62,8 @@ def respond_offender_sar_params
       create_offender_sar_complaint_params
     end
 
+    def session_state
```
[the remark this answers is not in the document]
Polsonby: coincidence I think! Yikang has bravely volunteered to address any overlaps when she rebases

config/locales/en.yml
```diff
@@ -536,6 +536,45 @@ en:
         third_party_name: "Full name of the representative (optional)"
         third_party_relationship: Relationship to the subject
         total: Total
+      offender_sar_complaint:
+        case_reference_number: Case reference number (CRN)
+        external_reference: Requester reference (optional)
+        high_profile: High profile
+        message: Additional case information
+        number_dispatched_pages: "Number of pages"
```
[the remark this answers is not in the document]
willmcb: Haha no worries not a big deal in any case :)

PR #1388, opened 2020-11-24
CT-3237 Assign case to complaints officer

Polsonby:
## Description
This PR adds an assignment to the user who created the case. We also update the Offender SAR Complaint factory to default to a state of "To be assessed" now that this is the new initial state; this resulted in a few updates to specs which needed it.

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
Case assigned to creator - reflected on the Case Show page: 
image

"My cases" show up based on the self-assignment by user who created them:
image

Case history no longer shows approver assignment
image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3237

### Deployment
n/a

Sign in a a branston user and create a new Offender SAR Complaint - should be assigned to the user when you view the case show page and show up in "My open cases" list. There should be no change to the display of Offender SAR cases.

### DIFF
--- a/app/helpers/cases_helper.rb
+++ b/app/helpers/cases_helper.rb
@@ -51,6 +51,7 @@ module CasesHelper #rubocop:disable Metrics/ModuleLength
               class: 'button',
               method: 'post'
     when :assign_responder
+      return '' if @case.type_of_offender_sar?
       link_to I18n.t('common.case.assign'),
               new_case_assignment_path(@case),
               id: 'action--assign-to-responder',
--- a/app/services/case_self_assign_service.rb
+++ b/app/services/case_self_assign_service.rb
@@ -15,7 +15,6 @@ class CaseSelfAssignService
       if @assignment.valid?
         managing_team = @user.responding_teams.first
         @case.state_machine.assign_responder! acting_user: @user, acting_team: managing_team, target_team: target_team
-        @assignment.case.state_machine.accept_approver_assignment!(acting_user: @user, acting_team: target_team)
         @assignment.accepted!
         @assignment.save!
         @result = :ok
--- a/app/services/case_self_assign_service.rb
+++ b/app/services/case_self_assign_service.rb
@@ -1,0 +1,29 @@
+class CaseSelfAssignService
+  attr_reader :result, :assignment
+
+  def initialize(kase:, role:, user:)
+    @case = kase
+    @role = role
+    @user = user
+    @result = :incomplete
+  end
+
+  def call
+    Assignment.connection.transaction do
+      target_team = @user.responding_teams.first
+      @assignment = @case.assignments.new(team: target_team, role: @role, user: @user)
+      if @assignment.valid?
+        managing_team = @user.responding_teams.first
+        @case.state_machine.assign_responder! acting_user: @user, acting_team: managing_team, target_team: target_team
+        @assignment.case.state_machine.accept_approver_assignment!(acting_user: @user, acting_team: target_team)
+        @assignment.accepted!
+        @assignment.save!
+        @result = :ok
+      else
+        @result = :could_not_create_assignment
+      end
+    end
+    @result == :ok
+  end
+
+end
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -52,6 +52,12 @@ Rails.application.configure do
     Bullet.add_whitelist :type => :n_plus_one_query,
                          :class_name => 'BusinessUnit',
                          :association => :moved_to_unit
+    Bullet.add_whitelist :type => :n_plus_one_query,
+                         :class_name => 'Case::SAR::Offender',
+                         :association => :responder_assignment
+    Bullet.add_whitelist :type => :n_plus_one_query,
+                         :class_name => 'Case::SAR::OffenderComplaint',
+                         :association => :responder_assignment
 
     # Enable this to track down most of the N+1 query issues
     Bullet.raise         = true # raise an error if n+1 query occurs
--- a/spec/features/cases/offender_sar_complaint/case_creating_spec.rb
+++ b/spec/features/cases/offender_sar_complaint/case_creating_spec.rb
@@ -23,6 +23,7 @@ feature 'offender sar complaint case creation by a manager', js: true do
     then_expect_cases_show_page_to_be_correct_for_data_subject_requesting_own_record
     then_expect_linked_original_case_has_stamp_for_linkage
     then_expect_open_cases_page_to_be_correct
+    then_expect_case_in_my_open_cases
   end
 
   scenario '2 Data subject requesting data to be sent to third party' do
@@ -37,9 +38,10 @@ feature 'offender sar complaint case creation by a manager', js: true do
     then_expect_cases_show_page_to_be_correct_for_data_subject_sending_data_to_third_party
     then_expect_linked_original_case_has_stamp_for_linkage
     then_expect_open_cases_page_to_be_correct
+    then_expect_case_in_my_open_cases
   end
 
- scenario '3 Solicitor requesting data subject record' do
+  scenario '3 Solicitor requesting data subject record' do
     when_i_navigate_to_offender_sar_complaint_subject_page
     and_choose_original_offender_sar_case_and_confirm
     and_fill_in_requester_details_page(:third_party)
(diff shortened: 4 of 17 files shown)


willmcb:
Shall I update the migration script so that 'open' migrated cases default to `to_be_assessed` rather than `data_to_be_requested`?

config/environments/test.rb
```diff
@@ -52,6 +52,12 @@
     Bullet.add_whitelist :type => :n_plus_one_query,
                          :class_name => 'BusinessUnit',
                          :association => :moved_to_unit
+    Bullet.add_whitelist :type => :n_plus_one_query,
+                         :class_name => 'Case::SAR::Offender',
+                         :association => :responder_assignment
+    Bullet.add_whitelist :type => :n_plus_one_query,
+                         :class_name => 'Case::SAR::OffenderComplaint',
+                         :association => :responder_assignment
```
ymao2, line 60: just for my knowledge, what do the above do? :)

ymao2:
@Polsonby ,  tried it locally, all works!  just something in the case history IMO, it will cause some confusion 

Screenshot 2020-11-26 at 09 48 57
1 - the Clearance level seems not for complaint case?
2 - the second one a little bit duplication (no any useful information), but probably ok

And also those events I think are linked to the 2 actions you use in the code `assign_responder ` and `accept_approver_assignment `,  maybe consider to create a new event which is only used for complaint like "self-assign"-> the service name you suggested, then generate a history message more proper to complaint?,  but just a suggestion, if too much hassle, no need to change it , just find a way to remove the clearance level.

ymao2:
sorry, I was thinking to pick up the ticket for changing the team member ticket, then think about the action `assign responder` will be useful for the complaints which don't have the responder e.g. all those open migrated cases from access db,  same thing for  `accept_approver_assignment`

app/services/case_self_assign_service.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def call
+    Assignment.connection.transaction do
+      target_team = @user.responding_teams.first
+      @assignment = @case.assignments.new(team: target_team, role: @role, user: @user)
+      if @assignment.valid?
+        managing_team = @user.responding_teams.first
+        @case.state_machine.assign_responder! acting_user: @user, acting_team: managing_team, target_team: target_team
+        @assignment.case.state_machine.accept_approver_assignment!(acting_user: @user, acting_team: target_team)
+        @assignment.accepted!
```
willmcb, line 19: Out of interest what does this `.accepted!` method do?

app/helpers/cases_helper.rb
```diff
@@ -51,6 +51,7 @@ def action_button_for(event)
               class: 'button',
               method: 'post'
     when :assign_responder
+      return '' if @case.type_of_offender_sar?
```
willmcb, line 54: We use `.offender_sar?` in `_case_status.html.slim` as a replacement for `.type_of_offender_sar?` should we use it here for consistency as well?

config/environments/test.rb
willmcb: Is this allowing n+1 queries? Seems like this would be a bad thing to do... 🤔

spec/features/cases/offender_sar_complaint/case_creating_spec.rb
```diff
@@ -248,7 +248,7 @@ def then_basic_details_of_show_page_are_correct(offender_sar_case: nil)
   def then_expect_open_cases_page_to_be_correct(offender_sar_case: nil)
     click_on "Cases"
     expect(open_cases_page).to be_displayed
-    expect(cases_page).to have_content "Branston Registry"
+    expect(cases_page).to have_content "branston registry responding user"
```
willmcb, line 251: Should this be capitalised as well like the the text it replaced?

spec/services/case_self_assign_service_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  let(:team) { find_or_create :team_branston }
+  let(:unassigned_case) { find_or_create :offender_sar_complaint }
+  let(:responder) { find_or_create :branston_user }
+  let(:service)           { CaseSelfAssignService
+                              .new kase: unassigned_case,
+                                   role: 'responding',
+                                   user: responder }
+  let(:new_assignment) { instance_double Assignment }
+
+describe '#initialize' do
```
willmcb, line 13: Indentation is a little off here 👍

spec/services/case_self_assign_service_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'rails_helper'
+
+describe CaseSelfAssignService, type: :service do
+  let(:team) { find_or_create :team_branston }
+  let(:unassigned_case) { find_or_create :offender_sar_complaint }
```
willmcb, line 5: Given that most things are passed in to this service as dependencies - I think we could do more mocking here possibly rather than actually hit the DB? What do you think? 🤔

spec/views/cases/_case_status_html_slim_spec.rb
```diff
@@ -68,7 +70,8 @@
       current_state: 'drafting',
       type_abbreviation: 'FOI',
       who_its_with: 'DACU',
-      type_of_offender_sar?: false
+      type_of_offender_sar?: false,
```
willmcb, line 73: Why have both of these methods? They seem to essentially be duplicates of each other.

willmcb (approved):
Looks good - few non-blocking things to think about - happy to discuss if you'd like :)

app/services/case_self_assign_service.rb
Polsonby: Part of the Rails built in ActiveRecord support for enum - it enables magic bang and predicate methods to call on the model instance that are derived from the enum names and set the relevant field to the corresponding value
https://api.rubyonrails.org/v5.2.3/classes/ActiveRecord/Enum.html

app/helpers/cases_helper.rb
Polsonby: I don't think that's where we ended up... 🤔 sorry for the confusion

My original proposal was to use `.offender_sar?` in that sense but we changed it back to mean Offender::Standard

any place you see it it means it - any place you see `.type_of_offender_sar?` it means either standard or complaint

The above is intentional to hide the button that would show up because of the new event in the state machine config

spec/features/cases/offender_sar_complaint/case_creating_spec.rb
Polsonby: It's a different field - I've just used what the factory provided here 🤔

spec/services/case_self_assign_service_spec.rb
Polsonby: Thanks, good spot :)

spec/views/cases/_case_status_html_slim_spec.rb
Polsonby: offender_sar? is standard ones, type_of_offender_sar? is either standard or complaint

config/environments/test.rb
Polsonby: So, Bullet is a gem that we added to the project to track down slow queries  - it makes suggestions about how to fix n+1 query issues that it detects. The problem is that when we have complex ActiveRecord scopes like the ones in this app, its recommendations about how to fix it (typically using `include`) aren't actionable. This seems to have been consistent with previous developers, because there are a bunch of whitelisted cases in the app. I did spend some time experimenting to see if I could make it happy but couldn't find the magic incantation, so gave up and added two more to the list above. IMO this is not going to cause any discernable real world issues

spec/services/case_self_assign_service_spec.rb
Polsonby: Dunno! I've just tried to follow the same pattern as similar specs here and this is how their setup works - don't have an opinion about it, really. If anything, I'm worried there's _too much_ mocking in this test, actually, it's the main one I used to drive out the feature

ymao2:
Nice work!!,  LGTM :)

ymao2 (approved):

config/environments/test.rb
willmcb: Great no probs 👍

willmcb (approved):
Thanks for your responses! 👍

PR #1390, opened 2020-11-25
CT-3280 Changed the scope of business unit performance report to received date only

ymao2:
## Description

This PR is to change the scope of the business unit report to be based on the received_date as the basic scope

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3280

1 - Login as London user
2- Choose the "reports" link from top menu
3 - Click "Create report"
4 - Choose the FOI / SAR to generate business unit performance report
5 - Check the figure

### DIFF
--- a/spec/services/stats/r003_business_unit_performance_report_spec.rb
+++ b/spec/services/stats/r003_business_unit_performance_report_spec.rb
@@ -214,7 +214,9 @@ module Stats
           Timecop.freeze Time.new(2017, 6, 30, 12, 0, 0) do
             d1 = Date.new(2017, 6, 4)
             d2 = Date.new(2017, 6, 30)  
-            report = R003BusinessUnitPerformanceReport.new(period_start: d1, period_end: d2); report.run; report.to_csv
+            report = R003BusinessUnitPerformanceReport.new(period_start: d1, period_end: d2)
+            report.run 
+            report.to_csv
           end
         end
 
--- a/spec/services/stats/r003_business_unit_performance_report_spec.rb
+++ b/spec/services/stats/r003_business_unit_performance_report_spec.rb
@@ -240,8 +240,7 @@ module Stats
               [[6, :green], [18, :green]],
               [[6, :red], [12, :red], [18, :red]],
             ])
-
-          end
+        end
 
         it 'outputs results as a csv lines' do
           super_header = %q{"","","","","","",} +
--- a/spec/services/stats/r003_business_unit_performance_report_spec.rb
+++ b/spec/services/stats/r003_business_unit_performance_report_spec.rb
@@ -212,7 +212,9 @@ module Stats
       describe '#to_csv' do
         let(:report_csv) do
           Timecop.freeze Time.new(2017, 6, 30, 12, 0, 0) do
-            report = R003BusinessUnitPerformanceReport.new; report.run; report.to_csv
+            d1 = Date.new(2017, 6, 4)
+            d2 = Date.new(2017, 6, 30)  
+            report = R003BusinessUnitPerformanceReport.new(period_start: d1, period_end: d2); report.run; report.to_csv
           end
         end
 
@@ -220,14 +222,13 @@ module Stats
           rag_ratings = report_csv.map do |row|
             row.map.with_index { |item, index| [index, item.rag_rating] if item.rag_rating }.compact
           end
-
           expect(rag_ratings).to eq([
               [],
               (0..23).map { |x| [x, :blue] },
               (0..23).map { |x| [x, :grey] },
               [[6, :red], [12, :red], [18, :red]],
-              [[6, :red], [12, :red], [18, :red]],
-              [[6, :red], [12, :red], [18, :red]],
+              [[6, :red], [12, :green], [18, :red]],
+              [[6, :red], [12, :green], [18, :red]],
               [[6, :red], [12, :red], [18, :red]],
               [[6, :red], [12, :red], [18, :red]],
               [[6, :red], [18, :red]],
@@ -240,7 +241,7 @@ module Stats
               [[6, :red], [12, :red], [18, :red]],
             ])
 
-        end
+          end
 
         it 'outputs results as a csv lines' do
           super_header = %q{"","","","","","",} +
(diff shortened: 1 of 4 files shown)


spec/services/stats/r103_sar_business_unit_performace_report_spec.rb
```diff
@@ -115,38 +123,214 @@ module Stats
...
+        end
+      end
+    end
 
-
-        if type.present?
-          refusal_reason = CaseClosure::RefusalReason.__send__(type)
-          kase.refusal_reason = refusal_reason
+    #rubocop:disable Metrics/ParameterLists
+    def create_case(received:, responded:, deadline:, team:, responder:, ident:, flagged: false, type: nil)
```
willmcb, line 305: Not saying that you need to do this but we could use a builder to stop having to pass so many things into this method - just a thought 👍 

Builder pattern: https://medium.com/kkempin/builder-design-pattern-in-ruby-dfa2d557ff1b

spec/services/stats/r003_business_unit_performance_report_spec.rb
```diff
@@ -212,22 +212,23 @@ module Stats
       describe '#to_csv' do
         let(:report_csv) do
           Timecop.freeze Time.new(2017, 6, 30, 12, 0, 0) do
-            report = R003BusinessUnitPerformanceReport.new; report.run; report.to_csv
+            d1 = Date.new(2017, 6, 4)
+            d2 = Date.new(2017, 6, 30)  
+            report = R003BusinessUnitPerformanceReport.new(period_start: d1, period_end: d2); report.run; report.to_csv
```
willmcb, line 217: I think its better to have statements on multiple lines if we can - easier to read and using the `;` syntax to brake up statements is not really very common in Ruby code 👍

spec/services/stats/r003_business_unit_performance_report_spec.rb
```diff
@@ -296,8 +296,7 @@ module Stats
       describe '#results' do
         before do
           Timecop.freeze Time.new(2017, 6, 30, 12, 0, 0) do
-            # report = R003BusinessUnitPerformanceReport.new
-            report = R003BusinessUnitPerformanceReport.new(period_start: Date.today.beginning_of_year, period_end: Date.today, generate_bu_columns: true)
+            report = R003BusinessUnitPerformanceReport.new(period_start: Date.new(2017, 6, 2), period_end: Date.today, generate_bu_columns: true)
```
willmcb, line 299: Why does this need to be 'today'? - I always prefer to have defined dates in tests as it makes them more deterministic.

spec/services/stats/r003_business_unit_performance_report_spec.rb
```diff
@@ -430,24 +429,24 @@ module Stats
               %q{Performance %,Total received,Responded - in time,Responded - late,Open - in time,Open - late,} +
               %q{Performance %,Total received,Responded - in time,Responded - late,Open - in time,Open - late}
             expected_text = <<~EOCSV
-              Business unit report (FOIs) - 1 Jan 2017 to 30 Jun 2017
+              Business unit report (FOIs) - 5 Jun 2017 to 30 Jun 2017
```
willmcb, line 432: Why did we change this date from Jan to June? Does it just make the test faster or something? :)

spec/services/stats/r103_sar_business_unit_performace_report_spec.rb
```diff
@@ -1,9 +1,62 @@
...
+
+      @responder_a = create :responder, responding_teams: [@team_a]
+      @responder_b = create :responder, responding_teams: [@team_b]
+      @responder_c = create :responder, responding_teams: [@team_c]
+      @responder_d = create :responder, responding_teams: [@team_d]
+
+      @sar_tmm = create :refusal_reason, :sar_tmm
+      @vex = create :refusal_reason, :vex
+
+      # create cases based on today's date of 30/6/2017
```
willmcb, line 30: What does 'today's date' mean in this context - it was today's date three years ago but not today 🤔

spec/services/stats/r103_sar_business_unit_performace_report_spec.rb
```diff
@@ -115,38 +123,214 @@ module Stats
...
-        kase
       end
-      #rubocop:enable Metrics/ParameterLists
 
 
+      if type.present?
+        refusal_reason = CaseClosure::RefusalReason.__send__(type)
+        kase.refusal_reason = refusal_reason
+      end
+      kase.save!
```
willmcb, line 329: I think `kase.save!` will return the case anyway. Not sure we need a subsequent case.

spec/services/stats/r103_sar_business_unit_performace_report_spec.rb
```diff
@@ -115,38 +123,214 @@ module Stats
...
+    #rubocop:disable Metrics/ParameterLists
+    def create_case(received:, responded:, deadline:, team:, responder:, ident:, flagged: false, type: nil)
+      received_date = Date.parse(received)
+      responded_date = responded.nil? ? nil : Date.parse(responded)
+      kase = nil
+      Timecop.freeze(received_date + 10.hours) do
+        factory = :accepted_sar
+        kase = create factory, responding_team: team, responder: responder, identifier: ident, received_date: received_date
+        kase.external_deadline = Date.parse(deadline)
+        if flagged == true
```
willmcb, line 313: Think this can just be `if flagged` as flagged will be `true` or `false` there is no need to check it 👍

willmcb (approved):
Some small things here if you'd like to have a look at the comments - nothing blocking though 👍

spec/services/stats/r003_business_unit_performance_report_spec.rb
ymao2: @willmcb, for this one,  the time has been freeze at ` Timecop.freeze Time.new(2017, 6, 30, 12, 0, 0)`, so the today won't be changed, just a short cut for using `2017, 6, 30, 12, 0, 0`

spec/services/stats/r003_business_unit_performance_report_spec.rb
ymao2: @willmcb , doesn't make it faster, but it will affect the outcome (the scope of cases), previous one doesn't test the scope properly which means the tests don't cover the important part of testing report, ( we need to count right scope of the cases based on the time frame)

spec/services/stats/r103_sar_business_unit_performace_report_spec.rb
ymao2: @willmcb,  today here is same meaning as the above I explain, normally in all those tests related report, time is quite sensitive and important factor for the outcome (statistic figures), in order to avoid uncertain, we normally will use absolute timing like using Time.freeze, not sue relative in order to get rid of which day is weekend, or bank holiday or Feb thing),  so `today` here means the date the time is freeze at

spec/services/stats/r003_business_unit_performance_report_spec.rb
ymao2: thank for this, I didn't really pay attention to this line other than changing the time frame,  good to know `;` usage in Ruby :)

spec/services/stats/r103_sar_business_unit_performace_report_spec.rb
ymao2: @willmcb, 👍 ,   had a read the link, it is definitely much clearer way to build the instance  could I do this next time, I will assume we will do more report testing when we do complaint---> sounds a execute 😊 ,  but I ensure I have put this idea in my mind

willmcb (approved):
Great work 👍

PR #1391, opened 2020-11-26
CT-3165 specify complaint type, subtype, priority

Polsonby:
## Description
This is a small PR adding the fields and values for the complaint type, subtype and priority. There's a hack in the offender_complaint to pre-fill the fields since there is no UI to do this yet, this will be removed in the follow up PR

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
n/a
### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3165

### Deployment
Adds a data migration task to add default values for the new fields so that any existing complaint cases can still be saved and worked with

none

### DIFF
--- a/app/models/case/sar/offender_complaint.rb
+++ b/app/models/case/sar/offender_complaint.rb
@@ -7,12 +7,20 @@ class Case::SAR::OffenderComplaint < Case::SAR::Offender
 
   jsonb_accessor :properties,
                  complaint_type: :string,
+                 priority: :string
+
   validates :complaint_type, presence: true
+  validates :priority, presence: true
 
   enum complaint_type: {
     standard:  'standard',
     ico: 'ico',
     litigation: 'litigation',
+  }
+
+  enum priority: {
+    normal:  'normal',
+    high: 'high',
   }
 
   class << self
@@ -33,14 +41,14 @@ class Case::SAR::OffenderComplaint < Case::SAR::Offender
     false
   end
 
-  private 
+  private
 
   def stamp_on_original_case
     self.original_case.state_machine.add_note_to_case!(
       acting_user: self.creator,
       acting_team: self.creator.case_team(self.original_case),
       message: I18n.t(
-        'common.case/offender_sar.complaint_case_link_message', 
+        'common.case/offender_sar.complaint_case_link_message',
         creation_date: self.created_at.to_date))
   end
 
--- a/app/models/case/sar/offender_complaint.rb
+++ b/app/models/case/sar/offender_complaint.rb
@@ -32,6 +32,16 @@ class Case::SAR::OffenderComplaint < Case::SAR::Offender
     high: 'high',
   }
 
+  # CT-3165 WIP REQUIRED FOR VALIDATIONS
+  #         REMOVE ONCE UX IS COMPLETED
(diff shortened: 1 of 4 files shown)


app/models/case/sar/offender_complaint.rb
```diff
@@ -5,6 +5,43 @@ class Case::SAR::OffenderComplaint < Case::SAR::Offender
...
+
+  enum complaint_subtype: {
+    missing_data:  'missing_data',
+    inaccurate_data: 'inaccurate_data',
+    redacted_data: 'redacted_data',
+    timeliness: 'timeliness',
+  }
+
+  enum priority: {
+    normal:  'normal',
```
willmcb, line 31: Doesn't priority already exist? Not sure there is a need to have 'normal' at all given that is the default. Is it not just a boolean flag to say something is high priority 🤔

app/models/case/sar/offender_complaint.rb
```diff
@@ -5,6 +5,43 @@ class Case::SAR::OffenderComplaint < Case::SAR::Offender
...
+    high: 'high',
+  }
+
+  # CT-3165 WIP REQUIRED FOR VALIDATIONS
+  #         REMOVE ONCE UX IS COMPLETED
+  before_validation :set_types
```
willmcb, line 37: 👍

db/data_migrations/20201126094009_set_types_for_complaints.rb
```diff
@@ -0,0 +1,10 @@
+class SetTypesForComplaints < ActiveRecord::DataMigration
+  def up
+    kases = Case::Base.offender_sar_complaint.all
```
willmcb, line 3: Is this best as a migration or a rake task? Just thinking about the sequencing of the migration. Maybe I have the wrong end of the stick, but could you explain your reasoning to me :)

spec/models/case/sar/offender_complaint_spec.rb
```diff
@@ -44,6 +44,87 @@
...
+          build(:offender_sar_complaint, complaint_type: 'wibble')
+        }.to raise_error ArgumentError
+      end
+    end
+  end
+
+  describe '#priority' do
+    context 'validates that priority is not blank' do
+      it 'errors' do
+        pending 'removal of the "set_type" validation hack'
```
willmcb, line 77: Nice

willmcb (commented):
Can we have a quick chat?

ymao2 (approved):

app/models/case/sar/offender_complaint.rb
Polsonby: Great point... 🤔 I guess I've approached it from the point of view that there are really three priorities but we're collapsing them into two. Using the enum even with a boolean automatically gives us a full set of methods we can use to query and set the value, and using the strings makes it easier if we end up extending this beyond two states one day.

db/data_migrations/20201126094009_set_types_for_complaints.rb
Polsonby: You're quite right - I did consider making it a rake task but we never remember to run them. Happy to change if you like?

PR #1403, opened 2020-12-09
CT-3166 specify contact name

Polsonby:
## Description
This PR adds support for specifying contact names for the ICO and Litigation complaint types, as well as validations and the conditional display of the values on the case details page.

## Self-review checklist

* [x] (3) Tests passing

### Screenshots

image
image
image
image

image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3166

### Deployment
n/a

As Branston Responder, create a complaint case
Specify ICO - verify that name, reference and a contact means (either email or phone) are mandatory
Verify the details display in the relevant section of the case details page
Same for Litigation


spec/features/cases/offender_sar_complaint/case_creating_spec.rb
```diff
@@ -4,19 +4,18 @@
   given(:manager)         { find_or_create :branston_user }
   given(:managing_team)   { create :managing_team, managers: [manager] }
   given(:offender_sar) { create :offender_sar_case, :third_party, :closed}
+  given(:complaint_type) { 'standard_complaint' }
+  given(:complaint_subtype) { Case::SAR::OffenderComplaint.complaint_subtypes.keys().sample }
+  given(:complaint_priority) { Case::SAR::OffenderComplaint.priorities.keys().sample }
+  given(:default_params) { { complaint_type: complaint_type, complaint_subtype: complaint_subtype, priority: complaint_priority } }
+  given(:target_case) { build(:offender_sar_complaint, default_params).decorate }
```
ymao2, line 12: @Polsonby,  if we setup the basic types like complaint_type at the beginning of test spec, then it will lose the intention of practising the possible combination of those types cross different tests within this test spec, the coverage of possible scenarios will be significantly less.   I would still prefer to generate the choices dynamically among this test spec unless there is reason for not being able to do so.

spec/features/cases/offender_sar_complaint/case_creating_spec.rb
```diff
@@ -286,55 +348,38 @@ def and_confirm_offender_sar_case
...
-    @chosen_complaint_subtype = actual_params[:complaint_subtype] || 
-                Case::SAR::OffenderComplaint.complaint_subtypes.keys().sample(1)[0]
-    @chosen_complaint_priority = actual_params[:priority] || 
-                Case::SAR::OffenderComplaint.priorities.keys().sample(1)[0]
-    actual_params[:complaint_type] = @chosen_complaint_type
-    actual_params[:complaint_subtype] = @chosen_complaint_subtype
-    actual_params[:priority] = @chosen_complaint_priority
-    actual_params
-  end
-
```
ymao2, line 311: as I mentioned in my previous comments,  I would prefer to generate the choices of types within the test spec, maybe I misunderstood your intention here,  discuss further?

app/models/case/sar/offender_complaint.rb
```diff
@@ -8,17 +8,28 @@ class Case::SAR::OffenderComplaint < Case::SAR::Offender
...
   validate :validate_external_deadline
 
   enum complaint_type: {
-    standard_complaint: 'Standard',
-    ico_complaint: 'ICO',
-    litigation_complaint: 'Litigation',
+    standard_complaint: 'standard_complaint',
+    ico_complaint: 'ico_complaint',
+    litigation_complaint: 'litigation',
   }
```
ymao2, line 34: sounds like you still choose to define the wording in yaml file :),   in that case,  I need to change the coding in the filters otherwise the wording for filters won't be correct.

Polsonby: Yeah, I think there's value in following the same pattern unless there's an overwhelming reason to depart from it, which is why I've changed this back to work like other enums in the app - felt like a lot of fiddling, but I think this is the right way to go :)

ymao2: @Polsonby  ok for me 👍 🤓

spec/features/cases/offender_sar_complaint/case_creating_spec.rb
Polsonby: I think there were a couple of reasons for this - one was that when I added in the ability to set and check for the contacts, the tests started failing randomly because there's a dependency between complaint types and expectations of what the contact details part of the page should contain. I'm struggling to remember properly now after the gap, but I think it was something along those lines. Was conscious that this reduced the test variations but  I'm not sure how I feel about randomised tests and expectations; I'd much rather tests were deterministic and always create the same expectations and test setup on every run. Not sure how much value the random coverage actually provides in practice?

spec/features/cases/offender_sar_complaint/case_creating_spec.rb
Polsonby: Yeah, sorry! I should have talked to you about this before PR. See other reply though :)  

Another wrinkle is I wanted to use let rather than instance variables and check the expectations based on an actual case object rather than fine-grained values; the main problem is this code would need to set contact details as well as setting the complaint type, subtype and priority. On the whole, I think we should use factories and traits to cover variations and have specific named scenarios that we care about, rather than dynamically generating the coverage

spec/features/cases/offender_sar_complaint/case_creating_spec.rb
Polsonby: However, I'm sure there's a better way to structure the tests than what I've done, maybe it's something to iterate on

ymao2 (approved):
Nice work!!, LGTM 👍

PR #1406, opened 2020-12-15
CT-3076 Fixed issue of setting up deadline for non-standard type

ymao2:
## Description

Fixed the bug of pre-populating the deadline for non-standard complaint

right now the majority of feature tests for complaints are all for standard type by default,  once tackle other complaint type like ico and litigation, new test specs need to be added to cover those missing cases. 

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3076


app/form_models/offender_sar_complaint_case_form.rb
```diff
@@ -77,7 +77,7 @@ def params_after_step_confirm_offender_sar(params)
   end
 
   def params_after_step_date_received(params)
-    if [nil, "standard", "ico"].include? params["complaint_type"]
+    if [nil, "standard"].include? object["complaint_type"]
```
willmcb, line 80: What is `object` in this context? Could we give it a less generic name?

willmcb (approved):
One small comment otherwise LGTM 👍

willmcb:
Should we add a ticket to the backlog for the new complaint types?

app/form_models/offender_sar_complaint_case_form.rb
ymao2: it is indeed confusing,  actually this module is included in `OffenderBaseDecorator`, the `object` is the way designed from Draper::Decorator (super super class of `OffenderBaseDecorator`),  it is designed as generic name by purpose I guess

ymao2: sorry for not explaining the `object,` the object is the actual activerecord model instance.  e.g. 
for `OffenderBaseDecorator`,  the object would be offender_sar case

PR #1412, opened 2020-12-23
CT-3322 Use only active teams for managing teams

Polsonby:
## Description
After reorganising teams, some users in BMT have been left with inconsistent privileges because the system checks for the manager role against teams that have been deactivated - we need to do the same for managing teams as we did for approving teams and ensure the system always checks for valid team memberships when deciding on manager role permissions

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots

Before - a user in BMT whose managing team roles return a deactivated version of BMT cannot create new cases  because all the correspondence types moved to the new incarnation of BMT
image

After - managing_teams only looks at active teams
image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3322

### Deployment
none

If you have an anonymised copy of the DB, you can reset the password and log in as user id 722, and go to Create Case. On master, you should see no available correspondence types. With this branch, you should see available correspondence types.


ymao2 (approved):
Nice work!, LGTM :)

PR #1421, opened 2021-01-12
CT-3338 add "not applicable"

Polsonby:
## Description
In order to facilitate migration of existing cases from Access, we need a new complaint_subtype of "None" so that migrated cases will validate and be able to be edited/saved once they're imported. 

Turns out we can’t use “none” because this will generate a class method "none", which is already defined by Active Record so using “Not applicable” instead

Can't find any tests to update, apologies for one-liner PR 😹 

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots

image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3338


willmcb:
Nice! Yeah I really should have thought about this a bit more 'none' is possibly the worst word I could have used - but like the classic example of someone with the surname 'null' causing all sorts of issues in systems...

willmcb (approved):
LGTM 👍

PR #1427, opened 2021-01-18
CT-3250 Add extra columns for complaint closed cases report

ymao2:
## Description

This PR is to add the extra columns required by Complaint's closed cases report mainly, in order to add costs and outcome for ico complaint case, I added 2 new fields and 1 closesure outcome metadata. 

## Self-review checklist

* [X] (3) Tests passing

### Screenshots

Screenshot 2021-01-18 at 14 04 11

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3250

### DIFF
--- a/app/models/case_closure/offender_complaint_appeal_outcome.rb
+++ b/app/models/case_closure/offender_complaint_appeal_outcome.rb
@@ -1,0 +1,45 @@
+# == Schema Information
+#
+# Table name: case_closure_metadata
+#
+#  id                      :integer          not null, primary key
+#  type                    :string
+#  subtype                 :string
+#  name                    :string
+#  abbreviation            :string
+#  sequence_id             :integer
+#  created_at              :datetime         not null
+#  updated_at              :datetime         not null
+#  requires_refusal_reason :boolean          default(FALSE)
+#  requires_exemption      :boolean          default(FALSE)
+#  active                  :boolean          default(TRUE)
+#  omit_for_part_refused   :boolean          default(FALSE)
+#
+
+module CaseClosure
+  class OffenderComplaintAppealOutcome < Metadatum
+    def self.upheld
+      where(abbreviation: 'upheld').singular
+    end
+
+    def self.not_upheld
+      where(abbreviation: 'not_upheld').singular
+    end
+
+    def self.not_response_received
+      where(abbreviation: 'not_response_received').singular
+    end
+
+    def upheld?
+      abbreviation == 'upheld'
+    end
+
+    def not_upheld?
+      abbreviation == 'not_upheld'
+    end
+
+    def not_response_received?
+      abbreviation == 'not_response_received'
+    end
+  end
+end
--- a/app/services/stats/etl/offender_sar_complaint_closed_cases.rb
+++ b/app/services/stats/etl/offender_sar_complaint_closed_cases.rb
@@ -42,8 +42,8 @@ module Stats
         'number_of_exempt_pages',
         'number_of_final_pages', 
         'appeal_outcome', 
-        'cost_paid', 
-        'settlement_cost_paid'
+        'totoal_cost', 
+        'settlement_cost'
       ]
 
       def result_name
--- a/app/services/stats/etl/offender_sar_complaint_closed_cases.rb
+++ b/app/services/stats/etl/offender_sar_complaint_closed_cases.rb
@@ -42,7 +42,7 @@ module Stats
         'number_of_exempt_pages',
         'number_of_final_pages', 
         'appeal_outcome', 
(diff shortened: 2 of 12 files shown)


app/models/case_closure/offender_complaint_appeal_outcome.rb
```diff
@@ -0,0 +1,45 @@
...
+#  subtype                 :string
+#  name                    :string
+#  abbreviation            :string
+#  sequence_id             :integer
+#  created_at              :datetime         not null
+#  updated_at              :datetime         not null
+#  requires_refusal_reason :boolean          default(FALSE)
+#  requires_exemption      :boolean          default(FALSE)
+#  active                  :boolean          default(TRUE)
+#  omit_for_part_refused   :boolean          default(FALSE)
```
Polsonby, line 16: Are these comments correct for this class? Just wondering if these are left behind from reusing an existing file :)

app/models/case_closure/offender_complaint_appeal_outcome.rb
```diff
@@ -0,0 +1,45 @@
...
+    def upheld?
+      abbreviation == 'upheld'
+    end
+
+    def not_upheld?
+      abbreviation == 'not_upheld'
+    end
+
+    def not_response_received?
+      abbreviation == 'not_response_received'
```
Polsonby, line 42: It's probably a "me problem" rather than a problem with the implementation, but just fwiw I find it surprising that "abbreviation" is storing these values. Is that how it's meant to be used?

app/models/case_closure/offender_complaint_appeal_outcome.rb
ymao2: The schema for this table case_closure_metadata:  are the same cross all those classes which make sense,   although some flags in this table may only be relevant some of them.   I am happy to remove them, just feel it is redundant or misleading :)

ymao2: but the good things to place them in metadata table rather than enum type, is to have active flag and other possible flags

app/models/case_closure/offender_complaint_appeal_outcome.rb
ymao2: I agreed, I found this a while ago which is puzzled me as well,  but don't want to change the way right now

ymao2: it may have other thought around this field initially e.g. the possibility of sub-group (2 level structure)? but somehow end up with majority of metadata classes  is one level there, and also it explain why it use `.singular`

app/models/case_closure/offender_complaint_appeal_outcome.rb
Polsonby: Oh, I see! Thanks for the explanation 👍

app/models/case_closure/offender_complaint_appeal_outcome.rb
Polsonby: Cool, thanks - sounds sensible! 😁

app/services/stats/etl/offender_sar_complaint_closed_cases.rb
```diff
@@ -33,7 +40,10 @@ class OffenderSarComplaintClosedCases < BaseClosedCases
         'number_of_days_taken',
         'number_of_final_pages::integer - number_of_exempt_pages::integer',
         'number_of_exempt_pages',
-        'number_of_final_pages'
+        'number_of_final_pages', 
+        'appeal_outcome', 
+        'cost_paid', 
+        'settlement_cost_paid'
```
Polsonby, line 46: Need to be updated if we rename the fields :)

spec/services/stats/r402_offender_sar_complaint_closed_cases_report_spec.rb
```diff
@@ -58,7 +63,10 @@ module Stats
 
       context '#case_scope' do
         it 'ignores any selected periods' do
-          expected = [@closed_offender_sar.name, @closed_offender_sar1.name]
+          expected = [
+            @closed_standard_complaint.name, 
+            @closed_ico_compliant.name, 
+            @closed_litgation_complaint.name]
```
Polsonby, line 69: I don't think we should hold up this PR, but fwiw as a general approach it makes me nervous seeing instance variables in specs - I think these should almost always be `let` rather than instance variables - maybe something to discuss as a group?

Polsonby (approved):
LGTM :) nice one 👍

PR #1430, opened 2021-01-21
CT-3384 Changed rule of complaint case number

ymao2:
## Description

This PR is to fix the issue of duplicate case number if multi-complaints are raised against the same offender sar case.  The rule is below 
Original Offender SAR case number: 210108002
New Complaint case number:  Q210108002
2nd Complaint case number:   Q210108002 - 001 
3rd Complaint case number:    Q210108002 - 002 and so on.

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/browse/CT-3384

### DIFF
--- a/app/models/case/sar/offender_complaint.rb
+++ b/app/models/case/sar/offender_complaint.rb
@@ -194,9 +194,29 @@ class Case::SAR::OffenderComplaint < Case::SAR::Offender
 
   def set_number
     if self.original_case.present?
-      self.number = "Q#{self.original_case.number}"
+      self.number = next_number_from_original_case
     else
       next_number
     end
   end
+
+  private 
+  
+  def next_number_from_original_case
+    # It should be rarely multiple persons are tryniig to create a new complaint 
+    # against the same original case and submit nearly at the same time. So IMO (yikang)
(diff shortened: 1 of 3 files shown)


app/models/case/sar/offender_complaint.rb
```diff
@@ -194,9 +194,27 @@ def update_deadlines
...
     else
       next_number
     end
   end
+  
+  def next_number_from_original_case
+    # It should be rarely multiple persons are tryniig to create a new complaint 
+    # against the same original case and submit nearly at the same time. So IMO (yikang)
+    # it is not worthying to track the counter per cases level at DB like case number for other types, 
+    # simple appoach here is to try 2 times only if the case number somehow is duplicate by any chance
```
Polsonby, line 207:
```suggestion
    # It should be rare that multiple persons are trying to create a new complaint 
    # against the same original case and submit nearly at the same time. So IMO (yikang)
    # it is not worth trying to track the counter per cases level at DB like case number for other types, 
    # simple appoach here is to try 2 times only if the case number somehow is duplicated by any chance
```

Polsonby (approved):
LGTM :) Nice - I like the thoughtful but pragmatic approach you've taken :)

PR #1440, opened 2021-02-02
CT-3406-fix-reopen-issues

ymao2:
## Description

This PR is to fix the issue of reopen found during QA

## Self-review checklist

* [X] (3) Tests passing


config/routes.rb
```diff
@@ -86,7 +86,7 @@
       get '/(:step)', on: :collection, to: 'offender_sar_complaint#new', as: 'step'
       post '/offender_sar/(:number)', on: :collection, to: 'offender_sar_complaint#start_complaint', as: 'start_complaint'
       get '/reopen', on: :member, to: 'offender_sar_complaint#reopen', as: 'reopen'
-      patch '/confirm_reopen', on: :member, to: 'offender_sar_complaint#confirm_reopen', as: 'confirm_reopen'
+      patch '/reopen', on: :member, to: 'offender_sar_complaint#reopen', as: 'confirm_reopen'
```
Polsonby, line 89: There's something about this that's making my spidey-senses tingle :)  

We don't usually have branches inside controller actions that filter on the HTTP method, like on app/controllers/cases/offender_sar_complaint_controller.rb:27 

I'd expect separate controller actions, as before your change. Just wondering what the reason is for overloading one controller action with both responsibilities 🤔 - otherwise everything looks good :)

ymao2: @Polsonby  I was trying to fix the issue of reloading the page if there is validation error after clicking the 'confirm' button,  then I encountered a issue of not having the 'get routine' to confirm_open url.  Hmm,  but now I think about it again 🤔 , maybe not related to the route, let me change it back.

ymao2: Changed it back as I found out I can re-use the same url but define a different action handler to handle it :)

Polsonby (approved):
Nice work, LGTM :)

PR #1441, opened 2021-02-02
CT-3225 change complaint case type

willmcb:
## Description
makes the case status reset when the complaint_case type changes, also shows multiple contact details for different complaint types if they're present.

**Notes for reviewer:**
- Do you now if there is a cleaner / more API way to reset the case status? I just amend it on the loaded ActiveRecord before its re-saved but I wonder if there are better ways to do this? 🤔 

## Self-review checklist

* [x] (3) Tests passing

Try to change the complaint case_type to prove the following
- changing between case types should reset the status to `to_be_assessed`
- changing the contact details of the complaint type but not changing the actual case type should not reset the status
- changing sub types or priorities should not change the status.
- changing the status twice (i.e. standard --> ico --> litigation) should mean that all of the contact details for ICO and litigation are visible on the summary screen.
- contact details should only appear if they are `.present?` no blank fields on the summary page 👍


app/services/complaint_case_updater_service.rb
```diff
@@ -0,0 +1,37 @@
...
+    complaint_type_mapping(case_type) != params_case_type && params_case_type.present?
+  end
+
+  def complaint_type_mapping(case_complaint_type)
+    {
+      'Litigation': 'litigation_complaint',
+      'ICO': 'ico_complaint',
+      'Standard': 'standard_complaint'
+    }[case_complaint_type.to_sym]
+  end
```
ymao2, line 36: This mapping is not necessary and also a little risky as the wording may changes and it is maintained in en.yaml,  if the @kase is here is decorator class of OffenderComplaint model,  you can retrieve the value of complaint_type by using @kase.object.complaint_type.

app/views/cases/offender_sar_complaint/_complaint_type.html.slim
```diff
@@ -15,45 +15,60 @@ tr.complaint-type.section-complaint-priority
...
-      = t('common.case.gld_contact_phone')
-    td
-      = case_details.gld_contact_phone
-  tr.complaint-type.section-gld-reference
-    th
-      = t('common.case.gld_reference')
-    td
-      = case_details.gld_reference
+
+- if case_details.litigation_complaint? || case_details.ico_complaint?
```
ymao2, line 19: I think the ico and gld information still needs to be shown up even the complaint type is standard because a case may be changed from ico/litigation to standard which had such informations in the past.

spec/controllers/cases/offender_sar_complaint_controller_spec.rb
```diff
@@ -363,8 +363,8 @@
   end
 
   describe '#edit' do
-    let(:offender_sar_case) { create(:offender_sar_case).decorate }
-    let(:params) {{ id: offender_sar_case.id }}
+    let(:offender_sar_complaint) { create(:offender_sar_complaint).decorate }
+    let(:params) {{ id: offender_sar_complaint.id }}
```
ymao2, line 368: 👍 ,   nice spot,  seems like we tested wrong case for a while 😁

app/services/complaint_case_updater_service.rb
```diff
@@ -0,0 +1,37 @@
+class ComplaintCaseUpdaterService < CaseUpdaterService
+  attr_reader :result
+
+  INITAL_CASE_STATUS = 'to_be_assessed'
+
+  def call
+    reset_status_on_complaint_type_change
+    super
```
ymao2, line 8: The idea of having different update services for different situations is very good,  for this particular one,  if the current state is reset back to the initial one ( even different one), IMO the change needs to be reflected in the case-transitions as well otherwise the state from most recent record in case-transitions won't be matched, we can discuss it further if need

ymao2 (approved):
LGTM :)

PR #1451, opened 2021-02-04
CT-3255 closure detail for complaint case type

ymao2:
## Description

This PR is to add the following features 
- Add/Edit outcome for complaint
- Add/Edit approval flags
- Add/Edit costs for litigation'
for complaint cases under "Requires response" and "closed" state

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3255
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3401
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3273
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3400
https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3402


app/controllers/cases/offender_sar_complaint_controller.rb
```diff
@@ -79,5 +79,15 @@ def process_date_responded_params
     def session_state
       "#{@correspondence_type_key}_state".to_sym
     end
+
+    private 
+
+    def clean_empty_approval_flags(permitted_params)
+      if permitted_params["approval_flag_ids"].present?
+        permitted_params["approval_flag_ids"].delete("")
```
willmcb, line 87: Not sure I follow this. It's saying `if value is present then delete the value with the key of "", or return nil if not found` is this right? Does the sub array have a key which is an empty string?

spec/factories/case_closure_metadata.rb
```diff
@@ -420,4 +443,38 @@
...
+
+    trait :second_approval do
+      name                        { 'Has this had deputy director information services approval?' }
+      abbreviation                { 'second_approval' }
+      sequence_id                 { 1010 }
+    end
+
+    trait :not_approval_required do
+      name                      { 'No approval needed' }
+      abbreviation              { 'not_approval_required' }
```
Polsonby, line 463: This is such a tiny thing, sorry, but I wonder if either "`approval_not_required`" or "`no_approval_required`" might read better?

Polsonby (commented):
Good work! LGTM

spec/models/case_closure/offender_complaint_approval_flag_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it { should validate_presence_of(:abbreviation) }
+    it { should validate_presence_of(:sequence_id) }
+
+    describe '.id_from_name' do
+      it 'returns the id when name exists in the database' do
+        complaint_first_approval = create :offender_ico_complaint_approval_flag, :first_approval
+        complaint_second_approval = create :offender_ico_complaint_approval_flag, :second_approval
+        complaint_not_approval_required = create :offender_ico_complaint_approval_flag, :not_approval_required
+
+        expect(OffenderComplaintApprovalFlag::ICOApprovalFlag.id_from_name(complaint_first_approval.name))
```
willmcb, line 16: Small thing but it might be easier to read if we were to store the class in a variable and refer to that rather than calling it directly multiple times. Only that the class names are so long it's a lot to look at.

willmcb (approved):
Nice work - a few really small non-blocking things to possibly think about but nothing major! 👍

app/models/case_closure/offender_complaint_approval_flag.rb
```diff
@@ -0,0 +1,43 @@
+
+module CaseClosure
+
+  class OffenderComplaintApprovalFlag
+
+    class ICOApprovalFlag < Metadatum
```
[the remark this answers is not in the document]
ymao2: The intention is to group the 2 flag classes together as it will be used in the same field , let me re-organise them by using module instead

app/controllers/cases/offender_sar_complaint_controller.rb
ymao2: When params returned from frontend contain the "approval_flag_ids",  it will include a empty "" in the array (not sure why),  this func is to remove this empty item from the array, otherwise a nil value will be stored although it won't cause any trouble, but just want to remove it

willmcb (approved):
Great work well done 👍

PR #1479, opened 2021-02-22
CT-3474 Added rake task for validating migrated complaints

ymao2:
## Description

This PR is to add a rake task for running validation against each complaint cases migrated from access DB, the result with the validation errors will be logged into csv file

The method I used is not the one we normally do in the rails app, it is because if I load the data through Case::SAR::OffenderComplaint,  it is highly likely that the whole process will be aborted because of some "bad" data, and there is no chance to get the whole picture of the problems we have. 

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3474

### DIFF
--- a/lib/tasks/complaint_cases_data.rake
+++ b/lib/tasks/complaint_cases_data.rake
@@ -1,4 +1,5 @@
 require 'csv'
+require 'json'
 
 namespace :complaints do
   namespace :links do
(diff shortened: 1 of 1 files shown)


lib/tasks/complaint_cases_data.rake
```diff
@@ -61,4 +62,56 @@ namespace :complaints do
...
+      end 
+      puts "Totally #{counter} complaints cases couldn't pass the validation."
+
+    end
+
+    private 
+    
+    def fake_offender_complaint_data_object(record)
+      record.delete("id")
+      record.delete("document_tsvector")
```
willmcb, line 104: Does this not delete the values on the original record?

I think we need to have a check on this - either not to run in prod or some user interaction to confirm 🤔

willmcb (approved):
LGTM following conversation 👍

PR #1489, opened 2021-03-06
CT-3425 Add 'High' flag on case list pages for high priority complaint case

ymao2:
## Description

This PR is to add 'High' flag for high priority complaint case at 'All open cases' and 'My open cases' tabs

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3425

### DIFF
--- a/app/decorators/case/base_decorator.rb
+++ b/app/decorators/case/base_decorator.rb
@@ -78,7 +78,7 @@ class Case::BaseDecorator < Draper::Decorator
     end
   end
 
-  def highted_flag
+  def highlight_flag
     trigger_case_marker
   end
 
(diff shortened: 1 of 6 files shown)


app/decorators/case/base_decorator.rb
```diff
@@ -78,6 +78,10 @@ def trigger_case_marker
     end
   end
 
+  def highted_flag
```
willmcb, line 81: I think "highted" is not that clear - is the intention that this method can check if the case is 'high' priority? 

If so maybe just have `.high_priority?`

willmcb (commented):
I'm not sure about the word 'highted' happy to chat through it with you 👍

willmcb (approved):
Thanks for clarifying that with me :)

PR #1490, opened 2021-03-08
CT-3290 Added extra backlink and fixed the bug

ymao2:
## Description

This PR is to 
- Add an extra back-link for the step of confirming offender SAR detail during the complaint-creation journey
- If the complaint is created via "Start complaint",  when a user clicks back-link, it should return back the SAR detail.

Sounds simple change but I found I cannot simply use 'back' behaviour as it is a wizard for the creation journey, so I need to add an extra flag if the starting point is from the related Offender SAR page, and also found some of unnecessary 'current-step' flag in a few pages, plus moved back_link() from decorator to controller level which I think is more appropriate. 

All in all,  somehow I ended up modifying a few more files, but they are still simple changes

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3290

### DIFF
--- a/app/controllers/cases/offender_sar_controller.rb
+++ b/app/controllers/cases/offender_sar_controller.rb
@@ -193,7 +193,7 @@ module Cases
     end
 
     def load_optional_flags_from_params
-      @creation_optional_flags.each do |key, value|
+      @creation_optional_flags.each do |key, _|
         @creation_optional_flags[key] = params[key]
       end
     end
--- a/app/controllers/cases/offender_sar_controller.rb
+++ b/app/controllers/cases/offender_sar_controller.rb
@@ -13,6 +13,7 @@ module Cases
       super
       @correspondence_type = CorrespondenceType.offender_sar
       @correspondence_type_key = 'offender_sar'
+      @creation_optional_flags = {}
     end
 
     def new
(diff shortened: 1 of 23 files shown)


app/controllers/cases/offender_sar_controller.rb
```diff
@@ -187,5 +191,28 @@ def preserve_step_state
...
+    end
+
+    def has_optional_flags?
+      @creation_optional_flags.present? && @creation_optional_flags.values.all? {|x| x.present?}
+    end
+  
+    def back_link_url
+      "#{@case.case_route_path}/#{@case.get_previous_step}#{build_url_params_from_flags}"
+    end
+
```
Polsonby, line 216: Wondering if some of this ought to be in a decorator? Perhaps OffenderBaseDecorator?

Polsonby (approved):
LGTM :) nice  - just wondered if some things could be in decorator but looks like you've already thought this through so mentioning in passing

app/controllers/cases/offender_sar_controller.rb
ymao2: @Polsonby,  actually I just moved it out of the decorator class 😳 as in order to construct the proper link I need to a few variables e.g. case_route_path and flags which are only available on a Controller level.   In a sense,  I prefer to have this link URL defined at controller level rather than decorator level, for me decorator is mainly used to enrich the ActiveRecord Model,  but anything to do with URL and behavior are in the controller

PR #1491, opened 2021-03-08
CT-3505 Fixed the wrong msg when no change made from user

ymao2:
## Description

This PR is to fix the wrong case update message even there is no change made from user 

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3505

### DIFF
--- a/app/services/case_updater_service.rb
+++ b/app/services/case_updater_service.rb
@@ -13,7 +13,13 @@ class CaseUpdaterService
     ActiveRecord::Base.transaction do
       begin
         @kase.assign_attributes(@params)
-        if @kase.changed?
+        # properties is JSON object, if one of keys was not included 
+        # but later this key with value of nil is added, 
+        # it will be treated as changed which will give a misleading message 
+        # as from user's point of view, he/she doesn't change anything
+        # Each key within this JSON object will be tracked individually, 
+        # no need for tracking properties as the whole.
+        if (@kase.changed_attributes.keys - ["properties"]).present?
           @kase.save!
           @kase.state_machine.edit_case!(acting_user: @user, acting_team: @team)
           @result = :ok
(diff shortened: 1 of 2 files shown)


app/services/case_updater_service.rb
```diff
@@ -13,7 +13,13 @@ def call
...
       begin
         @kase.assign_attributes(@params)
-        if @kase.changed?
+        # properties is JSON object, if one of keys was not included 
+        # but later this key with value of nil is added, 
+        # it will be treated as changed which will give a misleading message 
+        # as from user's point of view, he/she doesn't change anything
+        # Each key within this JSON object will be tracked individually, 
+        # no need for tracking properties as the whole.
+        if (@kase.changed_attributes.keys - ["properties"]).present?
```
willmcb, line 22: What is this line doing? If there are keys other than 'properties' would this not return a false positive? e.g.

```
irb(main):001:0> a = ['a', 'b', 'c']
irb(main):002:0> a - ['c']
=> ["a", "b"] # this is still .present?


# whereas this wouldn't be:
irb(main):005:0> a = ['a']
irb(main):006:0> a - ['a']
=> []
```

Is this okay? :)

willmcb (commented):
Small comment - likely me just misunderstanding but would be good to check with you 👍

app/services/case_updater_service.rb
ymao2: @willmcb , yes that is my intention :),   so if c is the 'properties, for the following cases, 1, 2 will be considered as changed 
1 - ['a', 'b', 'c'] - ['c'] 
2 - ['a', 'b'] - ['c']
3 - []-['c']
4 - ['c'] - ['c']

willmcb: Great thanks I thought it may have been but I thought I'd check :)

willmcb (approved):
Nice work 👍

PR #1497, opened 2021-03-14
CT-2303 Added a feature of uploading the requests files

ymao2:
## Description

This PR is to add a feature for London users to upload new request files after the FOI/SAR case is create

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2303


app/views/cases/attachments/new.html.slim
```diff
@@ -0,0 +1,37 @@
...
+      .column-two-thirds
+        p.form-hint
+          = t('cases.attachments.upload_hint_copy')
+    .grid-row
+      section.case-main
+        = render partial: 'shared/dropzone_form',
+              locals: { s3_direct_post: @s3_direct_post,
+                        file_input_name: 'uploaded_request_files[]' }
+        .form-group
+          label.form_label
```
digitalali-moj, line 33: Does this automatically add "for" attribute for the textarea?

app/views/cases/attachments/new.html.slim
```diff
@@ -0,0 +1,37 @@
...
+
+= form_for @case, as: :case_attachment, url: case_attachments_path(@case), method: :post do |f|
+  .form-group
+    .grid-row
+      .column-two-thirds
+        p.form-hint
+          = t('cases.attachments.upload_hint_copy')
+    .grid-row
+      section.case-main
+        = render partial: 'shared/dropzone_form',
```
digitalali-moj, line 29: Is this the same upload form used everywhere or are there others, just asking as we have some keyboard accessibility issues. If it's the same we can fix that later as there is a ticket for it.

app/controllers/cases/attachments_controller.rb
```diff
@@ -27,6 +29,34 @@ def destroy
...
+        @case = @case.decorate
+        flash.now[:alert] = service.error_message
+        render :new
+      else
+        if service.result == :ok
+          flash[:notice] = t('notices.request_uploaded')
+        end
+        redirect_to case_path @case
+      end
+    end
```
Polsonby, line 58: Wondering if the inner conditional on line 53 ought to be on the same level as the outer conditional from line 47? 

Or is it like this intentionally to handle the `:blank` case from the service? 🤔 

Just looks a little unusual - I think normally for these we do 

```
    if service.result == :ok
      happy things
    else
      render or redirect with error
    end
```

config/locales/page_titles.en.yml
```diff
@@ -42,3 +42,4 @@ en:
     stats: Performance Reports - Track-a-query
     upload_responses_page: Uploading a response - %{case_number} - Track-a-query
     move_team_page: Moving a business unit - %{team_name} - Track-a-query
+    add_request_attachment_page: Upload new files for this request
```
Polsonby, line 45: Sorry, nitpick! Could we add the missing newline here?

Polsonby (commented):
LGTM 👍 😁 - one question, one nitpick - happy to approve though

app/views/cases/attachments/new.html.slim
ymao2: @digitalali-moj  yes, it is the same upload form used everywhere

app/views/cases/attachments/new.html.slim
ymao2: I don't know, let me check

ymao2: no 'for' is added, I added it

app/controllers/cases/attachments_controller.rb
ymao2: I changed it to case statement with extra part for handling ':blank' value

app/views/cases/attachments/new.html.slim
digitalali-moj: Is it matching the ID of the textarea? I can't see the ID here so hard to tell.

ymao2: 🤦‍♀️ ,   no, sorry! I corrected it "again", should it be right this time? 😳 
screenshot below
Screenshot 2021-03-15 at 21 33 15

Polsonby (approved):
LGTM :) nice

PR #1514, opened 2021-03-26
CT-3564 correct calculation of performance percentage in business unit report

ymao2:
## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3564

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -61,6 +61,7 @@ gem 'slim-rails', '~> 3.2'
 gem 'shell-spinner'
 gem 'schema_plus_enums', '~> 0.1'
 gem 'sidekiq', '~> 6.1.3'
+gem 'mimemagic', '~> 0.3.9'
 
 gem 'table_print'
 # gem 'thor-rails'
--- a/app/services/stats/calculations.rb
+++ b/app/services/stats/calculations.rb
@@ -72,9 +72,18 @@ module Stats
     end
 
     def self.calculate_bu_overall_performance(row)
-      value = row[:trigger_responded_in_time] + row[:non_trigger_responded_in_time]
-      total = row[:non_trigger_responded_in_time] + row[:non_trigger_responded_late] + row[:non_trigger_open_late] +
-        row[:trigger_responded_in_time] + row[:trigger_responded_late] + row[:trigger_open_late]
+      value = row[:trigger_responded_in_time] + 
+              row[:trigger_open_in_time] + 
+              row[:non_trigger_responded_in_time] + 
+              row[:non_trigger_open_in_time]
+      total = row[:non_trigger_responded_in_time] + 
+              row[:non_trigger_responded_late] + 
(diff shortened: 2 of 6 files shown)


Gemfile
```diff
@@ -61,6 +61,7 @@ gem 'slim-rails', '~> 3.2'
 gem 'shell-spinner'
 gem 'schema_plus_enums', '~> 0.1'
 gem 'sidekiq', '~> 6.1.3'
+gem 'mimemagic', '~> 0.3.9'
```
willmcb, line 64: Nice :)

app/services/stats/calculations.rb
```diff
@@ -72,9 +72,18 @@ def self.calculate_overall_performance(row)
     end
 
     def self.calculate_bu_overall_performance(row)
-      value = row[:trigger_responded_in_time] + row[:non_trigger_responded_in_time]
-      total = row[:non_trigger_responded_in_time] + row[:non_trigger_responded_late] + row[:non_trigger_open_late] +
-        row[:trigger_responded_in_time] + row[:trigger_responded_late] + row[:trigger_open_late]
+      value = row[:trigger_responded_in_time] +
```
willmcb, line 75: Much clearer to read like this 👍

app/services/stats/calculations.rb
```diff
@@ -100,15 +109,21 @@ def self.calculate_trigger(row)
...
     def self.calculate_bu_non_trigger(row)
-      calculate_percentage(row[:non_trigger_responded_in_time], row[:non_trigger_responded_in_time] + row[:non_trigger_responded_late] + row[:non_trigger_open_late])
+      calculate_percentage(
+        row[:non_trigger_responded_in_time] + row[:non_trigger_open_in_time], 
+        row[:non_trigger_responded_in_time] + row[:non_trigger_responded_late] + row[:non_trigger_open_late] + row[:non_trigger_open_in_time])
     end
 
     def self.calculate_bu_trigger(row)
-      calculate_percentage(row[:trigger_responded_in_time], row[:trigger_responded_in_time] + row[:trigger_responded_late] + row[:trigger_open_late])
+      calculate_percentage(
```
willmcb, line 118: Really small thing but it might be good to have these additions stored in a variable and the fed into the `calculate_percentage` function like is done in `calculate_bu_overall_performance`

Up to you though not a big thing 👍

Gemfile
willmcb: Just wondering is it worth having this upgrade in a separate PR just so its clear that we introduced it? Rather than having it in the same PR as this report change. Just thinking it may be good to know... 🤔

willmcb (commented):
LGTM - see one comment about  mimemagic upgrade. what are your thoughts?

willmcb (approved):
👍

PR #1534, opened 2021-04-19
Add fonts to dockerfile

willmcb:
## Description
When word docs were converted to PDFs for preview purposes, the necessary fonts were not available at the point of conversion. Meaning that alpha chars were replaced with the  “Replacement Glyph” backstop character as specified in the [unicode standard](http://unicode.org/glossary/#replacement_glyph). 

This PR adds some more fonts to the Docker build so that fonts are available at the point of conversion from MS word format to PDF. 

Unfortunately, docs that have been converted previously will need to be re-uploaded as previews are cached rather than generated on the fly. 

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
Sample file in Arial font demonstrating issue:
Screenshot 2021-04-19 at 12 10 05

With fonts available:
Screenshot 2021-04-19 at 12 11 59

### Related JIRA tickets
[CT-35-63](https://dsdmoj.atlassian.net/browse/CT-3563)

1. Create a new case that utilise doc upload (SAR for instance)
2. Upload various sample files with differing fonts (see Jira ticket for sample files)
3. Refresh page until 'view' option is available (can take a few seconds for background job to run)
4. Test that all alpha characters in preview docs look correct 👍

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,8 @@ RUN addgroup --gid 1000 --system appgroup && \
 # Some app dependencies
 RUN apk add libreoffice clamav clamav-daemon freshclam
 
+RUN apk add libreoffice ttf-freefont ttf-opensans ttf-ubuntu-font-family ttf-inconsolata ttf-liberation ttf-dejavu
+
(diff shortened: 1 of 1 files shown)


Polsonby (approved):
Great investigation :) - nice one!

PR #1537, opened 2021-04-22
return nil eather than 0 if total is zero

willmcb:
## Description
Removes zero from % columns in reports when there are no values present.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
[CT-2453](https://dsdmoj.atlassian.net/browse/CT-2453)

### DIFF
--- a/app/services/stats/calculations.rb
+++ b/app/services/stats/calculations.rb
@@ -128,7 +128,7 @@ module Stats
 
     def self.calculate_percentage(value, total)
       if total == 0
-        0.0
+        nil
       else
         ((value / total.to_f) * 100).round(1)
       end
--- a/spec/features/admin/business_units_spec.rb
+++ b/spec/features/admin/business_units_spec.rb
@@ -39,7 +39,7 @@ feature 'editing teams' do
 
     teams_areas_page.add_area_field.set 'This is another area'
     teams_areas_page.add_button.click
-    teams_areas_page.wait_until_existing_areas_visible count: 2
+    sleep(2)
     expect(teams_areas_page.descriptions).to include 'This is another area'
 
     teams_areas_page.create.click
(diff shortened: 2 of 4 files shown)


app/services/stats/calculations.rb
```diff
@@ -128,7 +128,7 @@ def self.calculate_bu(row)
 
     def self.calculate_percentage(value, total)
       if total == 0
-        0.0
+        nil
```
ymao2, line 131: @willmcb,  changing here seems working with xlsx,   initially, I was thinking of not changing here as calculations.rb is not only used by this report, but for monthly performance report too (from another side, it maybe a benefit),  and also it is sort of arguable thing for returning nil rather 0 for math calculation 🤔 

another way for this ticket is to change the function `to_csv` in `base_business_unit_performance_report.rb`  between line 102 and 105,  let me know what you think, sorry my suggestion maybe is too late 😳

willmcb: I have no strong opinions - only really that if this is okay and _works_ I'd rather just merge it 👍

spec/features/admin/business_units_spec.rb
```diff
@@ -39,7 +39,7 @@
 
     teams_areas_page.add_area_field.set 'This is another area'
     teams_areas_page.add_button.click
-    teams_areas_page.wait_until_existing_areas_visible count: 2
+    sleep(2)
```
ymao2, line 42: IMO not remove the line and move the sleep(2) above this line, the count of 2 means 2 elements not 2 seconds?

ymao2 (approved):
Happy to approve 🤓 ! only just some minor comments which shouldn't hold this PR

spec/features/admin/business_units_spec.rb
willmcb: Yes I think you're correct - though as it is not an expectation and it passes with the sleep I think it is okay.

PR #1540, opened 2021-04-27
CT-2176 Allow London team to send the response back if there is something wrong being found out

ymao2:
## Description

This PR is to allow London User to send back a response due to something being found related to responses 
-  Allow the manager to send a ready-to-close case back to drafting status for a non-trigger foi case
-  Allow the manager to send a ready-to-close case back to drafting status for a trigger foi case 
-  Allow the approver to send a ready-to-dispatch case back to drafting status for a trigger foi case 
-  All those sending-back actions will trigger a re-drafting-notice email to responders 

For triggering email parts,  as we allow the user to add a message when they send back a case,  sending-back action and adding-message both will trigger emails, 
 - one for re-drafting notice email 
 - one of the message being received
 it seems the second one is redundant, I added an extra flag that allows the email-trigger to be disabled for the particular scenario 

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2176


Polsonby (approved):
Looks good to me :) nice work!

PR #1549, opened 2021-04-29
Add less cmd tool into build

willmcb:
Adds less into Dockerfile. Rails console "rails c" depends on pry which
in turn depends on less. Without this it means we can't use a lot of
rails console stuff in environments.

For instance Model.where(something: "thing") does not work.

Adding back this into the build to try to fix this issue.

See: https://github.com/pry/pry/issues/1248#issuecomment-366056015

## Self-review checklist

* [ ] (3) Tests passing

1. try to run `Model.where(something: "thing")` in rails console in dev - you should get an `less: unrecognized option: X ....` error.
2. build and deploy this branch to dev
3. run the same command in the rails console - it should be okay now.


Polsonby (approved):
thanks for taking the trouble to fix 😁

PR #1550, opened 2021-04-29
CT-2558 Fixed a few issues for admin page

ymao2:
## Description

This PR is to fix a few issues 
 - potential random problem from deactivated responding teams
 - remove unsupported case type from admin ( if we keep them in, we need to spent time to develop relevant parts and pages which I don't think it is worth doing those) 

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-2558

### DIFF
--- a/lib/cts/cases/create.rb
+++ b/lib/cts/cases/create.rb
@@ -450,7 +450,7 @@ module CTS::Cases
         elsif options.key?(:responder)
           responder.responding_teams.first
         else
-          BusinessUnit.includes(:responders).responding.sample
+          BusinessUnit.includes(:responders).responding.active.sample
         end
       end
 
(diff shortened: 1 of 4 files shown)


lib/cts/cases/create.rb
```diff
@@ -450,7 +450,7 @@ def responding_team
         elsif options.key?(:responder)
           responder.responding_teams.first
         else
-          BusinessUnit.includes(:responders).responding.sample
+          BusinessUnit.includes(:responders).responding.active.sample
```
Polsonby, line 453: Mildly surprising this didn't break an existing spec - not sure this is part of the app we spend much time in!

I wonder if it might be an idea to add some coverage for this change anyway?

ymao2: Changed the existing admin spec to add the same number of deactivated teams as the active ones to cover this scenario (50% chance),  although it doesn't guarantee to test fully, but should be fair chance.

Polsonby: Nice! Thanks for adding 👍

Polsonby (approved):
LGTM :) - great to fix things like this, thank you!

PR #1559, opened 2021-05-06
Add previous BU IDs to BU reports

willmcb:
## Description
This PR adds two extra columns to the business unit reports to add the ids of each business unit, and the previous ID of the Unit if it has been moved.

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
[CT-2822](https://dsdmoj.atlassian.net/browse/CT-2822)

1. if testing locally you might want to move some teams around so that the previous BU ids are populated.
2. down load some BU reports
3. should see two extra columns next to "Business unit" on the right: "Business unit ID" and "Previous business unit ID"


app/services/stats/calculations.rb
```diff
@@ -160,32 +160,58 @@ def self.populate_team_details_callback(stats)
...
-        result_set[:moved] = team.moved
+        default_results_setter(result_set, team)
       end
+      stats_total_setter(stats)
+    end
+
+    def self.business_unit_result_setter(result_set, team)
+      result_set[:business_unit] = team.name
+      result_set[:business_unit_id] = team.id
+      result_set[:previous_business_unit_id] = team.moved_to_unit_id
```
ymao2, line 179: I think `moved_to_unit_id` is the new unit the current unit moved to, not the previous one,  it should be next one?

app/services/stats/base_business_unit_performance_report.rb
```diff
@@ -5,6 +5,8 @@ class BaseBusinessUnitPerformanceReport < BaseReport
       business_group:                  'Business group',
       directorate:                     'Directorate',
       business_unit:                   'Business unit',
+      business_unit_id:                'Business unit ID',
+      new_business_unit_id:       'New business unit ID',
```
ymao2, line 9: maybe make the line in line with the above?

ymao2 (approved):
LGTM:).  because you changed the columns from 'previsou ..' to 'next ...'  which sort of changed the criteria of the ticket, you may double-check with Shakeel and Deepika, and update the ticket?

PR #1567, opened 2021-05-13
Default disclosure flag to true for non-standard FOIs

willmcb:
Defaults to flag for disclosure specialist to true, and then hides these
options from the user when the case is non standard.

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
[CT-2317](https://dsdmoj.atlassian.net/browse/CT-2317)

1. create case > foi
2. for type select either of the non-standard options, you should see that the "Flag for disclosure specialist" radio option at the bottom of the page should hide itself. If you click standard case type it should show itself again.
3. create the non-standard case. It should have automatically been flagged for a disclosure specialist.


digitalali-moj (approved):
JS looks good to me - haven't been able to test but I'm sure you've done that. It but make sense logically. Nice work. Just a reminder to please test on a few browsers, which you probably doing anyway.

With regards to the hiding of elements - as long as it's not essential since some modes may still show it if it's CSS hidden. e.g. non style view for Accessibility. So there is a small chance of changing it in that view - but not sure if it's an  issue, maybe worth double checking.

willmcb:
@ymao2 could you review this again just to make sure that my reasoning behind deleting  `spec/features/user_journeys/foi/compliance_review_non_trigger_e2e_spec.rb` makes sense to you? Thanks 😺

ymao2:
> @ymao2 could you review this again just to make sure that my reasoning behind deleting `spec/features/user_journeys/foi/compliance_review_non_trigger_e2e_spec.rb` makes sense to you? Thanks 😺

Yes, I agree this spec can be removed based on the requirement from this ticket

ymao2 (approved):

PR #1568, opened 2021-05-13
CT-3600 added complaint cases into anonymised script

ymao2:
## Description

This PR is to 
- add complaint case type into anonymized cases
- preserve admin user's credential 

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3600

### DIFF
--- a/lib/db/database_anonymizer.rb
+++ b/lib/db/database_anonymizer.rb
@@ -171,15 +171,16 @@ class DatabaseAnonymizer
   #
 
   def anonymize_users(user)
-    unless allow_to_be_reserved?(user.email)
+    unless allow_to_be_reserved?(user)
       user.full_name = Faker::Name.unique.name
       user.email = Faker::Internet.email(name: user.full_name)
     end
     user
   end
 
-  def allow_to_be_reserved?(email)
-    email =~ /@digital.justice.gov.uk$/ || email =~ /@justice.gov.uk$/
+  def allow_to_be_reserved?(user)
+    (user.email =~ /@digital.justice.gov.uk$/ || user.email =~ /@justice.gov.uk$/) &&
+      user.roles.include?("admin")
   end
 
   # Anonymize Cases table including all those case types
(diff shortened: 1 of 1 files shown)


lib/db/database_anonymizer.rb
```diff
@@ -171,15 +171,16 @@ def raw_sql_from_record(record)
...
+    unless allow_to_be_reserved?(user)
       user.full_name = Faker::Name.unique.name
       user.email = Faker::Internet.email(name: user.full_name)
     end
     user
   end
 
-  def allow_to_be_reserved?(email)
-    email =~ /@digital.justice.gov.uk$/ || email =~ /@justice.gov.uk$/
+  def allow_to_be_reserved?(user)
```
willmcb, line 181: This is a nicer interface now as well 👍

lib/db/database_anonymizer.rb
```diff
@@ -217,6 +218,20 @@ def anonymize_case_sar_offender(kase)
   end
   # rubocop:enable Metrics/CyclomaticComplexity
 
+  # rubocop:disable Metrics/CyclomaticComplexity
+  def anonymize_case_sar_offendercomplaint(kase)
+    anonymize_case_sar_offender(kase)
+    kase.ico_contact_name = Faker::Name.name unless kase.ico_contact_name.blank?
+    kase.ico_contact_email = Faker::Internet.email(name:kase.name) unless kase.ico_contact_email.blank?
```
willmcb, line 225: Does this mean that the non-anonymised case name is used in the email? We should use a fake name if this is the case 👍

lib/db/database_anonymizer.rb
```diff
@@ -217,6 +218,20 @@ def anonymize_case_sar_offender(kase)
...
+    kase.ico_contact_phone = Faker::PhoneNumber.cell_phone unless kase.ico_contact_phone.blank?
+    kase.ico_reference = "ICOF" + Faker::Invoice.reference unless kase.ico_reference.blank?
+    kase.gld_contact_name = Faker::Name.name unless kase.gld_contact_name.blank?
+    kase.gld_contact_email = Faker::Internet.email(name:kase.name) unless kase.gld_contact_email.blank?
```
willmcb, line 229: Sam here possibly?

willmcb (commented):
Just confirm the names used in the emails are anonymised and this is good to go 👍

lib/db/database_anonymizer.rb
ymao2: @willmcb , the kase.name has been anonymized in the previous steps `anonymize_cases()` (this fun will be applied to all case types),   the way here for generating email by using name is just to make the email more sensible in a way

willmcb: Brill I thought that would be the case but best to confirm with this 👍

willmcb (approved):
LGTM 👍

PR #1577, opened 2021-05-26
CT-2293 move directorate to different business group

ymao2:
## Description

This PR is to provide the function of allowing the manager to manage a directorate into different business group, this action involvs the following substeps
1 - create a copy of the directorate user wants to move under target business group
2 - move the business units under this directorate into the new copy of directorate from step one by one.
    each movement of the business unit follows the steps defined by TeamMoveService we have implemented already 
3 - deactivate the old directorate
4 - link deactivated the directorate with the new one. 

All those steps will be grouped under one transation, any failure from above will revert all those changes back. only when all those steps are succeeded, the changes will be reflected into DB

The potential problem I can think of is that the time of moving a big directorate will be a little bit slow which may cause some trouble (I am not sure) like beyond 30 seconds and timeout? I tried to move the biggest trunk of directorate, it was completed within 30 seconds, let us monitor this in the future, if it does happen, may need to introduce a background task along with other small things. 
## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3585


app/views/teams/move_to_business_group.html.slim
```diff
@@ -0,0 +1,13 @@
+- content_for :page_title do
+  = t('page_title.move_directorate_page', team_name: @team.name)
```
digitalali-moj, line 2: Can we please be specific with the page titles as it's one of the things we have tickets for in the Offender SAR journey for Accessibility. Let's maybe specify 'Move..' and 'Confirm Move...' or something.

config/locales/en.yml
```diff
@@ -332,6 +332,7 @@ en:
     add_to_case_history: Add to case history
     assign: Assign to %{business_unit_name}
     move: Move to this directorate
```
digitalali-moj, line 334: Again for accessiblity in these can you please say "Move xxx" or "Move to yyy" or "Move xxx to yyy" - basically the links need to make sense in isolation if possible. I think personally that "Move to yyy" is enough because the page is already of xxx so the user knows.

config/locales/en.yml
```diff
@@ -1293,6 +1294,24 @@ en:
       will_be_moved: "%{team_name}'s users and open cases will be moved to %{target_directorate_name}"
     team: Business unit
     unreversible: This action cannot be reversed
+  directorates:
+    move:
+      confirm_heading: Confirm move
+      heading: Move directorate
+      intro:
+        bullet1: all business unites will move with it
```
digitalali-moj, line 1302: minor spelling error I think should be units

config/locales/en.yml
```diff
@@ -1293,6 +1294,24 @@ en:
...
+        bullet3: all members under each business unit will be able to view closed cases
+        bullet4: performance reports for the unit both before and after the move are available
+        paragraph: "Moving a directorate means:"
+      moved_successfully: "%{team_name} has been moved to %{destination_business_group_name}"
+      moving_to: Moving to
+      subhead: "Directorate: %{team_name}"
+      team_currently_located: This is where the team is currently located
+      will_be_moved: "The users and open cases assigned to the business units under %{team_name} will be moved to %{target_business_group_name}"
+      error: "Failed to move this directorate due to reason: #{reason}"
+    unreversible: This action cannot be reversed
```
digitalali-moj, line 1313: Was wondering how true this is if I can just move it back to the one before

digitalali-moj (changes requested):
Fantastic work - Checked front end - some minor changes that we can discuss if needed.

app/views/teams/move_to_business_group.html.slim
ymao2: The current page title is like e.g. the name of directorate  is ABC
`Moving a directorate - ABC - Track-a-query`
so you suggest to change it to 
`Move ABC - Track-a-query` ?

config/locales/en.yml
ymao2: Not very sure whether we need to redesign if we change the name of the button, the following are the screenshot for the buttons 
e.g. - move business unit
Screenshot 2021-05-28 at 23 45 05
- move directorate
Screenshot 2021-05-28 at 23 43 34

config/locales/en.yml
ymao2: I think the 'cannot be reversed' here is more from technique point of view, means we cannot simply undo the movement action as there are lots of records which will be created and updated on database level, and those changes actually affect those open cases.

config/locales/en.yml
digitalali-moj: We need to change the wording only e.g. "Move to Private Office" or "Move to testing-group" - so the blind person can hear this.

config/locales/en.yml
digitalali-moj: OK, thanks

config/locales/en.yml
digitalali-moj: thanks

config/locales/en.yml
ymao2: done

app/views/teams/move_to_business_group.html.slim
digitalali-moj: Sorry I mean Move a directorate - ABC - Track-a-query 
and then Confirm Move directorate - ABC - Track-a-query

ymao2: I see,  thank you!, done

Polsonby (approved):
LGTM - nice work! Just one thing I was wondering about, does it fail the whole thing cleanly as it's moving the teams within a directorate if it hits a problem there? I think it does... 🤔

ymao2:
> LGTM - nice work! Just one thing I was wondering about, does it fail the whole thing cleanly as it's moving the teams within a directorate if it hits a problem there? I think it does... 🤔

I would think so, at least from my local running result, this is `ActiveRecord::Base.transaction ` does as long as we do not HIDE the exception,   the exceptions within `TeamMoveSevice` has been rescued within its transaction,  but I raise a different exception if the result from TeamMoveService failed  like below within `move_sub_teams`
`raise FailedMoveTeamError.new(team) unless result == :ok`

willmcb (approved):
Nice work @ymao2 LGTM 👍

app/views/teams/move_to_directorate.html.slim
```diff
@@ -11,3 +13,5 @@
 
 = render partial: 'browse_directorates'
 
+br
+= link_to "Back", team_path(@team), class: 'govuk-back-link'
```
digitalali-moj, line 17: Do we need this here? Just thinking does that mean all processes have to have a back button at the bottom?

digitalali-moj (approved):
Awesome work - couple of minor picky things only.

app/views/teams/move_to_directorate.html.slim
ymao2: @digitalali-moj ,  in this case, let me remove this bottom one ( I just thought it would be convenient 😳 ) to make this page in line with others

PR #1589, opened 2021-05-28
CT-3630 Fixed bugs for editing ICO detail page

ymao2:
## Description

Initial intention is to fix the bug of not being able to link a case to ICO case via editing the case detail's page, but later found out there were a quite few issues around this editing page:=
1 - Not being able to link to another case 
2 - Not being able to remove a related case 
3 - When relinking to a different original case or related case, the relevant reverse link was not created 
4 - If any inputs from this page fail the validation, it will raise exception and show the grey page
5 - "begin rescure" block is within transation in CaseUpdateService which make the transation lost its point 
6 - Even the related cases have been changed under editing page, the message still show 'No changes made' after saving and return back to case detail page

So basially, the editing page only works on most standard and happpy path. So the ticket ended up making the whole editing page works.

## Self-review checklist

* [X] (3) Tests passing


app/models/case/base.rb
```diff
@@ -348,7 +348,18 @@ def create!(_attributes = {})
...
+
+  has_many :original_case_links,
+           -> { original },
+           class_name: 'LinkedCase',
+           foreign_key: :case_id
+
+  has_many :original_cases,
+           through: :original_case_links,
+           source: :linked_case,
+           :after_remove => :delete_reverse_links
```
willmcb, line 362: Do you need the old style arrow syntax here? Could it not just be (same on line 352):

```
after_remove: :delete_reverse_links
```

🤔

app/services/case_updater_service.rb
```diff
@@ -1,35 +1,61 @@
...
+            !two_arrays_is_equal?(@kase.original_case_ids, @params[:original_case_ids])
+  end
+
+  def has_related_cases_changed?
+    return (@kase.respond_to? :related_case_ids) &&
+            @params[:related_case_ids].present? && 
+            !two_arrays_is_equal?(@kase.related_case_ids, @params[:related_case_ids])
+  end
+
+  def two_arrays_is_equal?(array1, array2)
```
willmcb, line 58: this might read better as `two_arrays_are_equal?` as is is singular and this refers to two things 👍

app/services/case_updater_service.rb
```diff
@@ -1,35 +1,61 @@
...
-
+    rescue StandardError => error
+      @error_message = error.message
+      @result = :error
     end
   end
+
+  private
+
+  def has_linked_cases_changed?
```
willmcb, line 42: This might read better as `have_linked_cases_changed` 👍

app/services/case_updater_service.rb
```diff
@@ -1,35 +1,61 @@
...
+    return has_original_case_changed? || has_related_cases_changed?
+  end
+
+  def has_original_case_changed?
+    return (@kase.respond_to? :original_case_ids) && 
+            @params[:original_case_ids].present? && 
+            !two_arrays_is_equal?(@kase.original_case_ids, @params[:original_case_ids])
+  end
+
+  def has_related_cases_changed?
```
willmcb, line 52: This might read better as `have_related_cases_changed?` 👍

willmcb (approved):
Really nice work well done for fixing all those issues you identified 😄 

Few non-blocking comments from me but otherwise looks great 👍

app/models/case/base.rb
ymao2: tried,  doesn't need to use the arrow syntax, I removed them

PR #1607, opened 2021-06-14
CT-3690 Add "your reference" to word letter template

willmcb:
Fills in the "Your Reference:" field in the letter templates.

------
**Note for changing this in future:**

This is a change to the actual word docx template rather than the html
body content which is defined in the letter seeders.

To get this to work you have to add a new 'merge field' through the word
GUI and not just copy an existing reference in the doc as this won't be
picked up.

In the doc there are references like «=letter_address» which are merge
fields, if you copy paste this and just change the name it won't work.

You need to go to on word for Mac: insert > field, then select 'Mail
Merge' > 'MergeField' then in the input on the dialog that has
'MERGEFIELD' add '=name_of_your_new_merge_feild', so it looks like:

'MERGEFIELD =name_of_your_new_merge_feild'

then > click 'OK'.

Then save the doc.

You then pass the new param to the template_data hash passed to the
Sablon template in LettersController and it should render on the word doc download.

If you don't follow these steps it won't work as word sets up some meta data to control these fields you need to go through the GUI...

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots

Before:
Untitled

After:
Screenshot 2021-06-14 at 17 40 55

### Related JIRA tickets
[CT-3690](https://dsdmoj.atlassian.net/browse/CT-3690)

- download a cases ack letter - should now have data in the 'Your Reference:' field


ymao2 (approved):
The code and template LGTM 👍 !  I didn't test, will leave them to  Shakeel or Deepika :)

PR #1608, opened 2021-06-15
CT-3610 Add newlines to letter address

willmcb:
## Description
Adds formatting to the addresses in the letter templates so they match what is described on the ticket.

Works if there are commas separating the fields. If addresses are just ended in a string with spacing, there is no way that I can think of to simply space them out over multiple lines. For instance:

"23 Fake Road Old Town Durham DU35GH"

So if addresses are entered like this without some sort of look up to a service that sends back a better formatted address I don't think they can be split over multiple lines accurately. 

Mostly addresses are separated by commas or newlines, which in the latter case are all ready correctly formatted - for instance:

23 Fake Road 
Old Town 
Durham 
DU35GH
 

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
Before:
Screenshot 2021-06-15 at 13 50 34

After:
Screenshot 2021-06-15 at 13 49 36

### Related JIRA tickets
[CT-3601](https://dsdmoj.atlassian.net/browse/CT-3610)

Look at a report and check that the address is on newlines.

### DIFF
--- a/app/models/letter.rb
+++ b/app/models/letter.rb
@@ -44,7 +44,7 @@ class Letter
     when "dispatch"
       values.recipient_address
     when "acknowledgement"
-      values.requester_address
+      format_address(values.requester_address)
     end
   end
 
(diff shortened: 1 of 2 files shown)


ymao2 (approved):
LGTM :),  I think we just need to mention to them about this rule

PR #1629, opened 2021-07-01
CT-3705 Amending reseed script to be runable on cloud env

ymao2:
## Description

This PR is to make the reseed script runnable on cloud env, then we can use it to reset the data on database on non-prod env.

The major issue I found was data:migrate part, the original code uses data:migrate as one of steps of reseeding,  which IMO, it is not ideal,  reseeding is to init the database to a suitable initial state,  but data:migrate is historic foot steps how the data were changes with time passed. Both are for different purpose.  even worse, the data:migrate uses different seeders,  so the consequence is if we change  the seeder to keep the data in line with latest app, then we actually changed the history,  such situation is ok when you just run data:migration, the  old file won't be executed again,  But it will cause trouble if we use this to do reseeding. In addition  reseeding is for non-prod env,  but data:migrate is for every env including prod.  Based on the situation, I make a few changes :-

- Removed one migration file which was used for seeding dev-team-and-users as it is dangerous one and shouldn't as part of our data migration steps, the reason being put there was purely for reseeding purpose
- changed 2 migration files to make the code more robust. 
- Updated the dev_team_users seed. 
 
## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3705


lib/tasks/db.rake
```diff
@@ -28,19 +45,32 @@ namespace :db do
...
 
+  def is_on_production?
+    ENV['ENV'].present? && ENV['ENV'] == 'prod'
+  end
+
+  def safeguard_question
+    if ENV['ENV'].present?
+      print "Are you sure to reset the database into initial state: no cases with default users and teams? Y/n "
+      input = STDIN.gets.chomp
+      exit unless(input.start_with?('Y')) || input.nil?
```
willmcb, line 71: Do the brackets here need to be inclusive of `input.nil?` here? 🤔

willmcb (approved):
LGTM - one small comment though.

PR #1638, opened 2021-07-08
Ct 3693 - xss brakeman

willmcb:
## Description
Ameliorate some xss issues + update brakeman ignore - see Spreadsheet attached to ticket for explanations.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
[CT-3693](https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3693)

- do you agree with my reasoning?

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -101,6 +101,37 @@
       "note": ""
     },
     {
+      "warning_type": "Cross-Site Scripting",
+      "warning_code": 2,
+      "fingerprint": "4c88cf1040891ad952164e56bfe504a90419114c72d7993038e40bafc05ba933",
+      "check_name": "CrossSiteScripting",
+      "message": "Unescaped model attribute",
+      "file": "app/views/cases/letters/show.html.slim",
+      "line": 29,
+      "link": "https://brakemanscanner.org/docs/warning_types/cross_site_scripting",
+      "code": "Letter.new(letter_params[:letter_template_id], Case::Base.find(params[:case_id]).decorate).body",
+      "render_path": [
+        {
+          "type": "controller",
+          "class": "Cases::LettersController",
+          "method": "show",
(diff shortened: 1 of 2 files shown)


config/brakeman.ignore
```diff
@@ -180,6 +242,37 @@
...
+        }
+      ],
+      "location": {
+        "type": "template",
+        "template": "cases/letters/show"
+      },
+      "user_input": null,
+      "confidence": "High",
+      "note": "Cloned object in Letter.rb with relevant escaped vals now passed to view"
+    },
```
ymao2, line 275: it seems like even we have a solution for escaping the value in cases, the above warning couldn't be eliminated? 🤔

config/brakeman.ignore
```diff
@@ -140,6 +171,37 @@
...
+          }
+        }
+      ],
+      "location": {
+        "type": "template",
+        "template": "cases/show"
+      },
+      "user_input": "params[:id]",
+      "confidence": "Weak",
+      "note": "False positive - vals passed aren't from user input"
```
ymao2, line 203: just minor comment to the note here,  IMO, we can mark this as false positive, it is because we will use it params[:id] to find the case first which through ActiveRecord.find() --> which prevent any potential bad scripting. 

Based on my understanding, anything in params is sort of 'user input' to backend, no matter which way, doesn't have to be inputed via user typing

willmcb: Yep, I think the same

config/brakeman.ignore
willmcb: did you run it with `bundle exec brakeman`? It's not showing up for me when I run it locally

ymao2 (approved):
LGTM :)

PR #1672, opened 2021-08-04
Ct 3372 case validation links

digitalali-moj:
## Description
1. There should be a dramatic improvement on the links at the top, which should work in most places. There are some instances, e.g. file uploads where they would still fail, but these would have to be fixed by the back end in a separate ticket.
2. There should no longer be two validation boxes on the page when errors happen
3. The error messages at the top should be in most cases consistent with the bottom, or the correct version of the two.
4. This will not be 100% perfect, as the back end approach varies slightly, but I’ve tried to fix most of the issues that are possible to fix using f/e code.

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
Affects most validation - but here is an example...
before:
image
after: 
image
before
image
after
image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3372

### Deployment
n/a

Check validation on all pages - should be improved - nothing should be broken any more than it was before.
So that's the messages at the top, and the links.

### DIFF
--- a/app/assets/javascripts/modules/ValidationLinks.js
+++ b/app/assets/javascripts/modules/ValidationLinks.js
@@ -1,0 +1,57 @@
+// This is to fix the mismatch between the prefixes for validation # links that break after
+// we re-use ruby code. i.e. when we use "as ..." in form generation, context breaks links
+moj.Modules.fixValidationLinks = {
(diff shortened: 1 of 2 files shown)


willmcb (approved):
LGTM nice work 👍

PR #1674, opened 2021-08-04
Ct 3372 fix broken main

digitalali-moj:
## Description
Main branch broken due to ES6 - remove ES6 features to allow for compilation - Uglify doesn't like ES6 - there is another version but not that reliable. Or we can change to terser. Since it's not used much anyway, easier to just remove ES6.


willmcb (approved):
LGTM 👍

PR #1681, opened 2021-08-11
CT-3778 Fix issue ico-decision-comment contains \n

ymao2:
## Description

This PR is to fix the displaying issue when the ico-decision-comment field of a ICO appeal case contains '\n'
Contenting having '\n' won't be issue if we put the comment in slim page directly,  but it will cause issue when we use Slim template to generate html.  After searching through the backend,  I couldn't see any reason we should construct a html codes and string from our backend.  So I moved the logic back to the slim page level, and removed all those redundant codes. 

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/browse/CT-3778


willmcb (approved):
LGTM 👍

PR #1682, opened 2021-08-17
CT-3750 move jobs into separated pod on qa

ymao2:
## Description

This PR is to move jobs container from main pod (webapp + upload) into separated pod which give more memory usage to jobs and also separate the issue of jobs from our main app.  --> which means any issues causing restart of jobs pod won't affect user daily use on our webapp

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=CT&modal=detail&selectedIssue=CT-3750


config/kubernetes/production/deployment_sidekiq.yaml
```diff
@@ -0,0 +1,71 @@
...
+  template:
+    metadata:
+      labels:
+        app: track-a-query-jobs
+    spec:
+      containers:
+        - name: jobs
+          image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/correspondence/track-a-query-ecr:latest
+          command: ["./config/docker/entrypoint-background-jobs.sh"]
+          resources:
```
digitalali-moj, line 24: Just wondering what happens when the limits here are breached? I guess the website keeps running but would we lose any data / would we get alerted?

Do you think we need to do this on other namespaces for consistency?

Does this clash with the resource limits in the 02-limitrange.yaml file on the CPE side? Sorry not sure how this works.

digitalali-moj (approved):
Looks great, just one question - maybe we can discuss some time, or just slack me. Thanks.

config/kubernetes/production/deployment_sidekiq.yaml
ymao2: @digitalali-moj ,  thank you for reviewing my PR
For you question.  the pod will  restart only when the maximum limit is breached, not the request limit. In theory,  it won't affect anything,  as sidekiq communicates with redis to get the task from different queues and then connect to DB for executing the task,  pod clashing won't cause the trouble of redis,  only thing I can think of is the clash is happening at middle of handling a task, but IMO,  as long as the mark of completing the  task is not reflected back to redis,  once the pod is live again, this task will be run again.  

The limit defined in 02-limitrange is only default value for all those containers, it will be overwritten by the  ones defined on container level. 

Yes, I would think we should keep this consistency cross different namespaces.  I done for PQ, not for others

PR #1715, opened 2021-09-02
CT-3046 remove hyphen from letter if no prison number

ymao2:
## Description

This PR is to remove the hyphen from the letter templates if there is no prison number for case
## Self-review checklist

* [X] (3) Tests passing

### Screenshots

before 
Screenshot 2021-09-02 at 10 41 52
after
Screenshot 2021-09-02 at 10 41 33

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3046


db/seeders/letter_template_seeder.rb
```diff
@@ -101,7 +101,7 @@ def seed!
                   <br><br>Dear Colleague
                   <br>
                   <br><strong>DATA PROTECTION ACT 2018: SUBJECT ACCESS REQUEST</strong>
-                  <br><strong><%= values.subject_full_name&.upcase %> - <%= values.first_prison_number %></strong>
+                  <br><strong><%= values.subject_full_name&.upcase %><% if values.prison_number.present? %> - <%= values.first_prison_number %></strong><% end %>
```
digitalali-moj, line 104: Should the end be before the </strong> ?

db/data_migrations/20210902085638_hyphen_when_no_prison_number_letter_template.rb
```diff
@@ -0,0 +1,7 @@
+class HyphenWhenNoPrisonNumberLetterTemplate < ActiveRecord::DataMigration
```
digitalali-moj, line 1: Not sure what this does tbh :), looks different from 20190930145607_seed_initial_letter_templates.rb. I'm assuming it's just reseeding templates but do we need that? Just a question.

digitalali-moj (changes requested):
Nicely done, just a couple of questions from me then good to go. Thanks :)

db/data_migrations/20210902085638_hyphen_when_no_prison_number_letter_template.rb
ymao2: @digitalali-moj,  if just for this step here, yes, I am trying to reseeding the whole templates as I change quite a few letters.    The reason why I use this way to  update the templates is blow 
https://dsdmoj.atlassian.net/browse/CT-3777

Basically,  in the past, we often (at least me) forgot to run the reseeding command for letter when we deploy the change into different envs which is not ideal,  the main reason why we keep forgetting is that the step of updating the template is not part of auto-deployment, if we want to update, we have to login the pod and run it manually

but I have my concerns (I wrote in the above ticket) to auto it as part of deployment job, so I use data-migration to auto it,  if you think about it again,  data-migration is just right for the purpose of changing data, and also it gives your flexibility to do something more like deleting some templates not just for reseeding
let me know, if you have other thoughts, thank you!

db/seeders/letter_template_seeder.rb
ymao2: @digitalali-moj ,  no,  it is syntax basically it means the prison-number with hyphen will be displayed only if the case has the prison_number

digitalali-moj (approved):
Thanks for explaining! :)

db/seeders/letter_template_seeder.rb
ymao2: @digitalali-moj ,  thank you for spotting this,  corrected them

PR #1717, opened 2021-09-03
Ct 3175 data reference table

willmcb:
## Description
Adds `CategoryReference` table to replace the use of enums and adds an implementation for it to be used with the Cotacts / address book.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
[CT-3783](https://dsdmoj.atlassian.net/browse/CT-3783)

Run the data migrations
Run a regular db migration

go to `/contacts` 
click 'Add new address'
Fill in form
Submit

Also test that validations are correct especially for the Contact_types

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -1,5 +1,6 @@
 class ContactsController < ApplicationController
   before_action :set_contact, only: %i[ edit update destroy ]
+  before_action :set_address_types, only: %i[ edit new ]
 
   def index
     @contacts = Contact.all
@@ -44,6 +45,10 @@ class ContactsController < ApplicationController
   private
     def set_contact
       @contact = Contact.find(params[:id])
+    end
+
+    def set_address_types
+      @address_types = CategoryReference.list_by_category(:address_type)
     end
 
     def contact_params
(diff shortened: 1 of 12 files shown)


app/controllers/contacts_controller.rb
```diff
@@ -1,8 +1,14 @@
...
   before_action :set_contact, only: %i[ edit update destroy ]
+  before_action :set_contact_types, only: %i[ create edit new update ]
 
   def index
     @contacts = Contact.all
+    @contacts.each do |contact|
+      contact.contact_type_display_value = CategoryReference.display_value_by_category_and_code(
+        :contact_type, contact.contact_type
+      )
+    end
```
ymao2, line 11: I have my concern about this approach to get the display name of contact type for each contact because the whole address book will be loaded, it may cause issue when the address book gets big,  is it possible to create relationship (join) to get this value, so only first page will be linked up to the reference table.     For this relationship, I think it won't cause trouble for creating a new contact type or a new contact, ?

app/models/contact.rb
```diff
@@ -27,4 +29,18 @@ def format_address(separator)
      .reject(&:empty?)
      .join(separator)
   end
+
+  def validate_contact_type
+    address_type_codes = CategoryReference.list_by_category(:contact_type).map do |category|
```
ymao2, line 34: Not sure whether we should make the name consist each  for address type  or contact type?

PR #1719, opened 2021-09-07
CT-3055 add view spec for cases list

ymao2:
## Description

This PR is to add view spec for open-cases-list page, mainly for testing the logic of request-detail columns
## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&modal=detail&selectedIssue=CT-3055


spec/views/cases/filters/index_html_slim_spec.rb
```diff
@@ -0,0 +1,273 @@
...
+    policy = double('Pundit::Policy', policy_name => true)
+    allow(view).to receive(:policy).with(:case).and_return(policy)
+  end
+
+  def disallow_case_policy(policy_name)
+    policy = double('Pundit::Policy', policy_name => false)
+    allow(view).to receive(:policy).with(:case).and_return(policy)
+  end
+
+  let(:responder)       { find_or_create :foi_responder }
```
willmcb, line 14: Could just run a linter on these variables to get them aligned more 👍

spec/views/cases/filters/index_html_slim_spec.rb
```diff
@@ -0,0 +1,273 @@
...
+    assign(:action_url, '/cases/open')
+    assign(:current_tab_name, 'open')
+
+    disallow_case_policy :can_add_case?
+
+    render
+    cases_page.load(rendered)
+
+    zeroth_case = cases_page.case_list[0]
+    expect(zeroth_case.number.text).to eq "Case number #{ovt_foi_trigger_case.original_case.number}"
```
willmcb, line 81: Might be nice to add these parts to named methods just so it's clearer to read, something like:
```ruby
def case_expectations(case, values)
    expect(case.number.text).to eq values[:case_num]
    expect(case.type.text).to eq values[:type_text]
    expect(case.flag.text).to eq values[:flag_text]
    expect(case.request_detail.text) values[:request_text]
    expect(case.draft_deadline.text).to eq values[:request_deadline_text]
   # etc...
end
```

and then just look up each case and create a hash for each - would save on some of the logic repetition 👍

spec/views/cases/filters/index_html_slim_spec.rb
```diff
@@ -0,0 +1,273 @@
...
+    assign(:query, search_query)
+    assign(:action_url, '/cases/open')
+    assign(:current_tab_name, 'open')
+
+    disallow_case_policy :can_add_case?
+
+    render
+    cases_page.load(rendered)
+
+    zeroth_case = cases_page.case_list[0]
```
willmcb, line 165: same here would be able to reuse that `case_expectations` method

willmcb (approved):
Nice - I think there is scope for making this test a bit DRYer please see comment. Its nice work though 👍

spec/views/cases/filters/index_html_slim_spec.rb
ymao2: @willmcb  thanks for this advice 👍 ,  I added 2 function to compare those fields,

willmcb (commented):
Nice 👍

PR #1722, opened 2021-09-09
CT-3805 bug: deactivating user who has some live cases and some responded cases (London team)

ymao2:
## Description

This PR is to fix the bug when a user having some live cases and some responded cases linked with is deactivated.  The bug was the inconsistency for handing the responded cases.  When user is identified as having live cases linked with by using `has_live_cases_for_team?(@team)` ( the logic for this check is checking whether there is any case whose state is not responded ),  then all those cases linking to this users will be unassigned and be reset back to the "awaiting_responder" state,  even the case is responded which conflict with the checking rule. 

The fix is to filter out those responded cases from the `unassign_from_user` process. 

## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/secure/RapidBoard.jspa?rapidView=25&projectKey=CT&modal=detail&selectedIssue=CT-3805

### DIFF
--- a/spec/features/admin/deactivating_users_spec.rb
+++ b/spec/features/admin/deactivating_users_spec.rb
@@ -1,12 +1,16 @@
 require 'rails_helper'
 
 feature 'deactivating users' do
-  given!(:team_admin)        { find_or_create :team_admin }
-  given!(:manager)        { find_or_create :disclosure_bmt_user }
-  given!(:responder)      { create :responder, responding_teams: [bu] }
-  given!(:user_with_cases){ find_or_create :foi_responder }
-  given!(:live_case)      { create :accepted_case }
-  given!(:bu)             { find_or_create :foi_responding_team }
+  given!(:team_admin)       { find_or_create :team_admin }
+  given!(:manager)          { find_or_create :disclosure_bmt_user }
+  given!(:responder)        { create :responder, responding_teams: [bu] }
+  given!(:user_with_cases)  { find_or_create :foi_responder }
+  given!(:live_case)        { create :accepted_case }
+  given!(:responded_case)   { create :responded_case,
+                                responder: user_with_cases,
+                                responding_team: bu,
+                                received_date: 5.days.ago }
+  given!(:bu)               { find_or_create :foi_responding_team }
 
   before do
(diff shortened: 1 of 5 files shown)


spec/features/admin/deactivating_users_spec.rb
```diff
@@ -1,12 +1,16 @@
...
+  given!(:team_admin)       { find_or_create :team_admin }
+  given!(:manager)          { find_or_create :disclosure_bmt_user }
+  given!(:responder)        { create :responder, responding_teams: [bu] }
+  given!(:user_with_cases)  { find_or_create :foi_responder }
+  given!(:live_case)        { create :accepted_case }
+  given!(:responded_case)   { create :responded_case,
+                                responder: user_with_cases,
+                                responding_team: bu,
+                                received_date: 5.days.ago }
+  given!(:bu)               { find_or_create :foi_responding_team }
```
willmcb, line 13: Really small thing and I know this wasn't your choice as it existed in the code already - but could we think of a better name than `bu`? Doesn't really mean anything - I guess "business user" but I'm not sure. Should probably just use `business_user` then as it is unambiguous.

Might be worth changing while we're editing this code, like the Boy Scout's rule: https://www.oreilly.com/library/view/97-things-every/9780596809515/ch08.html

spec/services/user_deletion_service_spec.rb
```diff
@@ -79,14 +79,37 @@
...
+        it 'unassigns the live cases from user without touching responded cases' do
           kase = create :accepted_case,
                         responder: responder,
                         responding_team: team
+
+          responded_case = create :responded_case,
+                                  responder: responder,
+                                  responding_team: team,
+                                  received_date: 5.days.ago
+          expect(kase.responder_assignment.state).to eq "accepted"
```
willmcb, line 91: We could wrap these expectation blocks in methods to make it a bit more readable:

`responder_has_accepted_initial_draft_case`
`responder_has_accepted_responded_case`

etc...

willmcb (approved):
Two minor non-blocking comments you might want to look at. Otherwise looks good 👍

PR #1740, opened 2021-09-24
CT-3841 change SearchUpdateIndex job to be on case level

ymao2:
## Description

This PR is to change the way of running SearchUpdateIndex job. 
The old way is to scan the whole cases table with dirty value of "true",  then run the updating-process on each one,  this scanning process will be triggered by any update on any cases, which cause the following issues
- Any one of the case has issue with the updating process,  all those cases newer than the failed case won't get the chance to run the task until someone spots it ( but quite often, we don't keep our eyes on those failures in sentry)
- Unnecessary overload the tasks.
I changed the task to make it being triggered and being run on case level, so decouple problematic case from "good" once but still can be visible in other monitor tools such sentry

## Self-review checklist

* [X] (3) Tests passing

### DIFF
--- a/spec/features/cases/foi/case_closing_spec.rb
+++ b/spec/features/cases/foi/case_closing_spec.rb
@@ -313,6 +313,7 @@ feature 'Closing a case' do
     random_exemption =  Random.new
                           .rand(1..(total_number_exemptions - 1))
 
+    expect(excemption_options.sizes > 1).to eq true
     excemption_options[random_exemption].click
 
     excemption_options[random_exemption].text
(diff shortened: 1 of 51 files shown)


spec/features/cases/foi/case_closing_spec.rb
```diff
@@ -313,6 +313,7 @@ def select_random_exemption
     random_exemption =  Random.new
                           .rand(1..(total_number_exemptions - 1))
 
+    expect(excemption_options.size > 1).to eq true
```
willmcb, line 316: Has this not been merged already?

willmcb (approved):
LGTM nice work 👍 One small comment in that we might want to remove the code to fix the test for the case closing spec as I thought we had already merged this. But this is a small thing.

Hopefully this will dramatically reduce the Sentry errors 👍

PR #1753, opened 2021-09-30
CT-3481 Make count tab values to reflect filters

willmcb:
## Description
 Makes the tabs on the "my open cases page" reflect the totals of the cases returned from the filters.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
[CT-3481](https://dsdmoj.atlassian.net/browse/CT-3481)

- create some complaint cases and assign them to yourself, make sure they are of different types.
- go to cases > my open cases 
- use the filters
- the numbers in the tabs should now reflect the totals of the results.


app/controllers/cases/filters_controller.rb
```diff
@@ -133,6 +135,29 @@ def open
...
+    end
+
+    def cases_for_tab(tab)
+      tab.cases.includes(
+        :message_transitions,
+        :responder,
+        :responding_team,
+        :approver_assignments
+      )
+    end
```
ymao2, line 159: It is good idea to introduce an attribute to store the count for tab. 
If just for getting the number of cases,  there is no need to pre-load all those sub tables like `message_transitions, responders etc. `

you can just use `tab.cases` directly,   the cases_for_tab may not be needed.

app/controllers/cases/filters_controller.rb
```diff
@@ -133,6 +135,29 @@ def open
...
     private
 
+    def set_filtered_tab_counts
+      @global_nav_manager.current_page.tabs.each do |tab| 
+        tab.set_count(
+          filtered_count_for_tab(
+            cases_for_tab(tab)
+          )
+        )
+      end
```
ymao2, line 145: for this part,  could I suggest only to calculate the non-active tab,  as the active tab will be calculated in my_open anyway,  so it may just add the following line to filter them out 
again, this way to reduce the unnecessary query time when loading the page
`    def set_filtered_tab_counts
      @global_nav_manager.current_page.tabs.each do |tab| 
        if tab.name != @global_nav_manager.current_page_or_tab.name
          tab.set_count(filtered_count_for_tab(tab.cases))
        end
      end
    end`

app/controllers/cases/filters_controller.rb
```diff
@@ -82,6 +83,7 @@ def my_open
           :responding_team,
           :approver_assignments
         )
+
       service = call_search_service(unpaginated_cases, cookies[:search_result_order])
       @query = service.query
```
ymao2, line 89: you can set the current active tab's count here like 

`@global_nav_manager.current_page_or_tab.set_count(service.result_set.count)`

ymao2 (approved):
Really Nice work! 👍

PR #1786, opened 2021-11-09
Ct 3934 make cant consistent

digitalali-moj:
## Description
Change "can't" to "cannot" and "doesn't" to "does not" where ever possible and quick

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
e.g.
image
image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-3934

### Deployment
n/a

Most places it should say in validation errors "cannot be" instead of "can't be"

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -171,7 +171,7 @@ RSpec.configure do |config|
   # Replace stubbing out CASE_UPLOADS_S3_BUCKET with a test harness of our own
   # making. The problem with stubbing out is that it doesn't work in 'before
   # :all' blocks, which are sometimes needed e.g. ICO appeals which always
-  # require an uploaded document. so can't be instantiate in a 'before :all'
+  # require an uploaded document. so cannot be instantiate in a 'before :all'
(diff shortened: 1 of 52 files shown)


spec/rails_helper.rb
```diff
@@ -171,7 +171,7 @@
   # Replace stubbing out CASE_UPLOADS_S3_BUCKET with a test harness of our own
   # making. The problem with stubbing out is that it doesn't work in 'before
   # :all' blocks, which are sometimes needed e.g. ICO appeals which always
-  # require an uploaded document. so can't be instantiate in a 'before :all'
+  # require an uploaded document. so cannot be instantiate in a 'before :all'
```
willmcb, line 174: I guess this should be "instantiated" which we might as well fix if we are here... :)

willmcb (approved):
LGTM - one small non-blocking comment 👍

spec/rails_helper.rb
digitalali-moj: 👍

PR #1796, opened 2021-11-15
CT-3951 added partial case flag filter

ymao2:
## Description

This PR is to add filters for partial case flag 

## Self-review checklist

* [X] (3) Tests passing

### Screenshots

1 - complaint natures 
Screenshot 2021-11-15 at 11 50 25
2 - partial case flag
Screenshot 2021-11-15 at 11 50 34

### Related JIRA tickets

https://dsdmoj.atlassian.net/jira/software/c/projects/CT/boards/25?modal=detail&selectedIssue=CT-3951


app/models/case/base.rb
```diff
@@ -171,6 +171,20 @@ def self.searchable_document_tsvector
...
 
+  scope :partial_case, -> {
+    where(
+      "properties->>'is_partial_case'::text = ? ",
+      true.to_s
+    )
+  }
+
+  scope :not_partial_case, -> {
+    where(
```
willmcb, line 182: could this not just be if !true? rather than checking both false and null.

app/services/case_filter/case_partial_case_flag_filter.rb
```diff
@@ -0,0 +1,51 @@
...
+    def is_partial_case?
+      'partial-case'.in? @query.filter_partial_case_flag
+    end
+
+    def is_not_partial_case?
+      'not-partial-case'.in? @query.filter_partial_case_flag
+    end
+
+    def filter_partial_case(records)
+      if is_partial_case? && !is_not_partial_case?
```
willmcb, line 42: Why the double check here isn't it that `is_partial_case?` is equivalent to `!is_not_partial_case?` ?

app/services/case_filter/case_partial_case_flag_filter.rb
```diff
@@ -0,0 +1,51 @@
...
+        records.partial_case
+      elsif !is_partial_case? && is_not_partial_case?
```
willmcb, line 44: Same here why the double check?

willmcb (approved):
Few minor comments otherwise LGTM 👍

app/models/case/base.rb
ymao2: does't work, != 'false' or <>'false' doesn't include the cases which doesn't have this field. This is why I wrote this way

app/services/case_filter/case_partial_case_flag_filter.rb
ymao2: sorry for this confusion,  the "is_partial_case" and "is_not_partial_case" represents the options from the filter, if only put 'is_partial_case', then 2 situation will fall in this categories 
  - user chose 'is_partial_case' only
  - user chose 'is_partial_case' and also 'is_not_partial_case'
then for first one, it will return partial cases only, but it would be wrong for the second situation,  it should return all cases, not only partial case.   So I put two to make sure the second situation won't be in this category. 

Save reason for your next comment

willmcb: 👍 thanks for confirming :)

PR #1806, opened 2021-11-26
Ct 3922 create sar ir case

willmcb:
## Description
Adds page to enter details of the SAR IR with relevant details pulled from the original SAR.

Note I am going to rails another PR with page objects for the feature tests. This PR was getting big so I wanted to get it reviewed before I made further changes.

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
Adds new page where user can enter the SAR IR case details:

Screenshot 2021-11-30 at 15 58 52

### Related JIRA tickets
[CT-3922](https://dsdmoj.atlassian.net/browse/CT-3922)
[CT-3923](https://dsdmoj.atlassian.net/browse/CT-3923)

- copy and existing SAR case number
- click create case
- enter copied SAR number
- Continue
- Add the extra details to the page
- Continue - should be directed to the case show page (not the case details partial for this is a placeholder at the moment)


app/views/cases/sar_internal_review/_new_form_common.html.slim
```diff
@@ -0,0 +1,65 @@
+section
+  header
```
digitalali-moj, line 2: How come we're using header here? and further down?

app/views/cases/sar_internal_review/_new_form_common.html.slim
```diff
@@ -0,0 +1,65 @@
...
+
+  br
+  br
+    = f.text_field :subject_type, class: 'visually-hidden',
```
digitalali-moj, line 6: Does this need to be here? As it will show without styles. And with the other ones below. Can we use hidden inputs instead?

app/views/cases/sar_internal_review/_new_form_common.html.slim
```diff
@@ -0,0 +1,65 @@
...
+    = f.text_field :subject_type, class: 'visually-hidden', 
+      label_options: { class: 'visually-hidden' }
+
+    = f.text_field :flag_for_disclosure_specialists, 
+      class: 'visually-hidden', 
+      label_options: { class: 'visually-hidden' }, 
+      value: 'true'
+
+    = f.radio_button_fieldset :sar_ir_subtype, 
+      choices: ['Timeliness', 'Compliance'],
```
digitalali-moj, line 15: Can this go in the lang file? And other strings below

digitalali-moj (commented):
Nice f/e work 👍 
Minor front end questions / potential changes.

app/models/case/sar/internal_review.rb
```diff
@@ -6,6 +6,19 @@ class Case::SAR::InternalReview < Case::SAR::Standard
 
   attr_accessor :original_case_number
 
+  validate :validate_email
+  validate :validate_postal_address
+  validate :name
+  validate :validate_subject
+  validate :validate_third_party_relationship
+
+  after_initialize :remove_conditional_validators
```
ymao2, line 15: @willmcb, I haven't tried it yet, but this line seems risky to me ,  and in my thought, it shouldn't need to do this. The problem with valid_attributes is that  it didn't really invoke the validator defined on specific attributes  correctly which means the validator will be skipped ( actually I start to doubt this valid_attributes? every works as I dug this issue a little more).   But those skips shouldn't affect your new validators in the SAR IR classes.   we can have discuss about it

app/controllers/concerns/sar_internal_review_form_validators.rb
```diff
@@ -0,0 +1,85 @@
...
+      :email,
+      :postal_address,
+      :third_party_relationship,
+      :subject
+    ]
+    _validators.reject! do |key, _val| 
+      parent_class_validators_to_remove.include?(key)
+    end
+  end
+end
```
ymao2, line 85: Hmm,  I think you don't really need to this file,  you already have all those new validators defined in data model level. any particular reason for such duplication?

app/models/case/sar/internal_review.rb
ymao2: even there is possible conflict ( which I feel it shouldn't,  you should be allowed to have multiple validators against one field, as long as one of them failed, an error should be raised. ),  I more prefer to redefine those validators in parent class level, rather do "removal" actions

ymao2 (commented):
I have some comments and some confuses, happy to discuss them

app/controllers/concerns/sar_internal_review_form_validators.rb
willmcb: Nope this is an oversight - I moved them here and then moved them back so I need to delete them here 👍

app/views/cases/sar_internal_review/_case_details_step.html.slim
```diff
@@ -1,5 +1,20 @@
...
-  = "holding page pending next step"
+- content_for :heading, flush: true
+  = t('cases.sar_internal_review.case_details')
 
-h2
-  = "Case name: #{@case.name}"
+= render partial: 'layouts/header'
+
+= GovukElementsErrorsHelper.error_summary @case.object,
+        "#{pluralize(@case.errors.count, t('common.error'))} #{ t('common.summary_error')}", ""
```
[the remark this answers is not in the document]
willmcb: Thanks will do - so is the idea to use 'cannot' instead?

app/models/case/sar/internal_review.rb
willmcb: I tend to agree, I added it this way as I thought changing them in the base class may have more side effects. Thinking about it though I think this way of doing it actually causes an issue in that you remove all of the validators on a given attribute, so unless you redeclare all of them some may be missing. I think I will just move them to the parent. Happy to chat about it though.

app/views/cases/sar_internal_review/_case_details_step.html.slim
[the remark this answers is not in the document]
willmcb: I have fixed these other than the link to the upload field. The error message shows but the hyperlink to the field doesn't work. However I think this is an existing issue with the upload partial.

app/views/cases/sar_internal_review/_case_details_step.html.slim
```diff
@@ -1,5 +1,20 @@
...
+
+table
```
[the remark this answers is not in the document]
willmcb: 👍

app/views/cases/sar_internal_review/_new_form_common.html.slim
willmcb: Just an oversight no real reason.

app/views/cases/sar_internal_review/_new_form_common.html.slim
willmcb: Do you mean the values of the array?

app/models/case/sar/internal_review.rb
willmcb: Have added these to the parent class and it appears to be working as expected.

ymao2: @willmcb, I have thought about another methods to do the partial validation, we can discuss tomorrow, it only need to add a small function on controller level,  it is safer and no need to changes anything in the parent class, let us catch-up tomorrow

app/views/cases/sar_internal_review/_case_details_step.html.slim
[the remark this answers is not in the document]
digitalali-moj: yes please if possible - some places you just can't :)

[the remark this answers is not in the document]
digitalali-moj: 👍

app/views/cases/sar_internal_review/_new_form_common.html.slim
digitalali-moj: Yes and any others on the page that are easy to put in en

app/views/cases/sar_internal_review/_new_form_common.html.slim
willmcb: Removed 👍

app/views/cases/sar_internal_review/_new_form_common.html.slim
willmcb: It is here as the form needs to submit this value on error else the params are blank and the error page renders Nil in the table above if this is not here. There may be other ways of doing this but it seemed like the most straight forward.

Is there a reason to use the hidden input over this method i.e. better for accessibility etc.?

app/views/cases/sar_internal_review/_new_form_common.html.slim
willmcb: I'll add these into the en.yml 👍

app/views/cases/sar_internal_review/_new_form_common.html.slim
digitalali-moj: Oh OK. Yes hidden input stays hidden if styles are off and won't be read out by the screen reader. But tbh it seems we're hiding and showing input fields in many places anyway. But good to minimise if easy, if not no big deal as we're not sticking to the rule in other places. (e.g. conditional radio reveals)

app/models/case/sar/internal_review.rb
willmcb: I managed to make it work by removing the call to `.attributes_valid?` on case completion and then just assigning the attributes before the in final `@case.valid?` check. With the new methods I was having issues with the linking validation as `.attributes_valid?` calls the link validation logic in the form object via interpolating the current step into a method name. In any case it's fine now as the logic is so simple in the steps that we can rely on the preexisting validations in the `Case::SAR::Standard`. Also allowed me to remove the form validator for the SAR IR which is good as it means less code :)

app/views/cases/sar_internal_review/_new_form_common.html.slim
willmcb: I have added `hidden: true` to the fields and this works but the labels still requite a 'visually-hidden' class else they show up. I think this is a little bit of an improvement still though, so I'll go with it 👍

app/models/case/sar/internal_review.rb
ymao2: I see, yes, for you case there is no need to do the partial validation actually as first step and second step are really simple,  I agree it would be simpler,  although this won't apply to offender SAR flow 

since I thought about the new way of doing partial validation , I am quite tempting to refactor the form and validators for Offender SAR and complaint Flow , as the new way most of them can be removed, the logic would become much easy even for understanding, I may find a chance to do this refactor in the future

ymao2 (approved):
Nice work and also the changes you made for the validation parts ! 👍

app/views/cases/sar_internal_review/_new_form_common.html.slim
```diff
@@ -0,0 +1,63 @@
...
+
+  #third-party
+    = f.radio_button_fieldset :third_party do |fieldset|
+      - fieldset.radio_input(true, text_method: :humanize)
+        = f.text_field :name, label_options: { value: t('helpers.fieldset.sar_internal_review.name') }
+        = f.text_field :third_party_relationship
+      - fieldset.radio_input(false, text_method: :humanize)
+
+section
+  header
```
digitalali-moj, line 25: does this need to go too?

app/views/cases/sar_internal_review/_new_form_common.html.slim
```diff
@@ -0,0 +1,63 @@
...
+    .grid-row
+      .column-two-thirds
+        div class="#{case_uploaded_request_files_class}" id=case_uploaded_request_files_id
+          = render partial: 'shared/dropzone_form',
+                  locals: { s3_direct_post: @s3_direct_post,
+                          file_input_name: 'sar_internal_review[uploaded_request_files][]' }
+
+  = f.hidden_field :flag_for_disclosure_specialists
+
+  h2.heading-medium How and where should the information be sent?
```
digitalali-moj, line 52: One for lang file?

digitalali-moj (approved):
Brilliant work! couple of minor comments but easy fixes.

PR #1807, opened 2021-11-29
Added a script for removing invalid cases on prod env

ymao2:
## Description

This PR is to add a script to clear up invalid cases from non-prod env
## Self-review checklist

* [X] (3) Tests passing

### Screenshots

https://dsdmoj.atlassian.net/jira/software/c/projects/CT/boards/25?modal=detail&selectedIssue=CT-3990


lib/tasks/remove_invalid_cases.rake
```diff
@@ -0,0 +1,157 @@
...
+          end
+        end
+      end
+      puts "Totally #{counter} invalid cases have been removed."
+    end
+
+    private 
+
+    def is_on_production?
+      ENV['ENV'].present? && ENV['ENV'] == 'prod'
```
willmcb, line 54: Doesn't everything on the test environments run in 'prod' mode? Would it not be better to flag based on the host instead like in `Settings.yml` with the feature flags 🤔

willmcb (approved):
LGTM one small commit though 👍

lib/tasks/remove_invalid_cases.rake
ymao2: I think you may get it mixed with another environment variable RAILS_ENV,   which has 2 values 'development' and 'production' and they are for rails's own running mode.   but  ENV['ENV'] is set from our config.yaml which is about which cloud environment this app is in and has values of 'staginig', 'production' and 'development' etc.

willmcb: Yep thats it 🤦

PR #1813, opened 2021-12-03
Turn off autocomplete for sar ir form

willmcb:
Turns autocomplete off for SAR IR create form


digitalali-moj (approved):
I'm assuming this stops Chrome Autocomplete which I guess we have to do here as it's mucking about with hidden fields. Please let me know if it's something else. 👍

PR #1816, opened 2021-12-06
Add ability to assign case

willmcb:
## Description
Adds ability to assign case. 

Will raise another PR to deal with assigning cases as trigger.

## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/app/views/cases/sar_internal_review/_case_details.html.slim
+++ b/app/views/cases/sar_internal_review/_case_details.html.slim
@@ -1,1 +1,32 @@
-// preliminary page pending next PR
+.grid-row
+  div class="column-full #{case_details_for_link_type(link_type)}"
+    h2.request--heading
+      / i18n-tasks-use t('common.case.case-details')
+      / i18n-tasks-use t('common.case.original-case-details')
+      = translate_for_case(case_details, 'common', case_details_for_link_type(link_type))
+
+    table
+      tbody.sar-basic-details
+        = render partial: 'cases/shared/case_type', locals: { case_details: case_details }
+        = render partial: 'cases/sar/data_subject', locals: { case_details: case_details }
+        = render partial: 'cases/sar/subject_type', locals: { case_details: case_details }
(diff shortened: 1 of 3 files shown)


digitalali-moj (commented):
f/e looks good to me, thanks :)

app/views/cases/sar_internal_review/_case_details.html.slim
```diff
@@ -1 +1,32 @@
...
+        = render partial: 'cases/shared/case_deadlines', locals: { case_details: case_details }
+        = render partial: 'cases/sar/response_address', locals: { case_details: case_details }
+        - if case_details.has_date_draft_compliant?
+          = render partial: 'cases/shared/draft_compliance_details', locals: { case_details: case_details }
+        - if case_details.responding_team.present?
+          = render partial: 'cases/shared/responding_team', locals: { case_details: case_details }
+        - if case_details.closed?
+          = render partial: 'cases/shared/closed_case_details', locals: { case_details: case_details }
+
+    - if @case.type_abbreviation != case_details.type_abbreviation
```
ymao2, line 27: Sorry, this line looks very suspicious to me,  are you trying to share this page between another case type? not sure under which scenario the case_detail will have different case type with @case, especially you place this under SAR_IR which is the leaf class already.  

Also if you just want to show the linked original case, there is already a shared template page for it 
shared/components/case_linked_cases,  you can check it from show.html.slim
but I may misunderstand your intention, let me know, if you need to discuss it

ymao2: @willmcb , tbh, I am not fully sure  whether the line of code will cause trouble or not, but at least the linked cases should show up automatically from link-cases-section.  we can always add it back if somehow we need this one in the future

ymao2 (approved):
LGTM :)

app/views/cases/sar_internal_review/_case_details.html.slim
willmcb: 👍

PR #1822, opened 2021-12-16
CT-4002 Handle main exceptions with GA and page

digitalali-moj:
## Description
GA Tracking and pages for main exceptions

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
Old (wth is a work basket?)
image
image

New pages

image
image
etc

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-4002

go to the wrong page

### DIFF
--- a/public/400.html
+++ b/public/400.html
@@ -1,31 +1,19 @@
 <!DOCTYPE html>
-<!--[if lt IE 9]><html class="lte-ie8" lang="en"><![endif]-->
-<!--[if gt IE 8]><!--><html lang="en"><!--<![endif]-->
+<html lang="en">
     <head>
         <meta charset="utf-8" />
-        <title>This page cannot be found - GOV.UK Contact</title>
+        <title>Error - 400 - Bad request - Track-a-query</title>
 
         <link rel="stylesheet" media="all" href="/assets/application.css" />
-        <!--[if IE 6]><link href="/assets/govuk-template-ie6.css" media="screen" rel="stylesheet" /><![endif]-->
-        <!--[if IE 7]><link href="/assets/govuk-template-ie7.css" media="screen" rel="stylesheet" /><![endif]-->
-        <!--[if IE 8]><link href="/assets/govuk-template-ie8.css" media="screen" rel="stylesheet" /><![endif]-->
-        <!--[if gt IE 8]><!--><link href="/assets/govuk-template.css" media="screen" rel="stylesheet" /><!--<![endif]-->
+        <link href="/assets/govuk-template.css" media="screen" rel="stylesheet" />
         <link href="/assets/govuk-template-print.css" media="print" rel="stylesheet" />
-        <!--[if gte IE 9]><!--><link href="/assets/fonts.css" media="all" rel="stylesheet" /><!--<![endif]-->
-
-        <!--[if IE 8]><link href="/assets/fonts-ie8.css" media="all" rel="stylesheet" /><![endif]-->
-        <!--[if gte IE 9]><!--><link href="/assets/fonts.css" media="all" rel="stylesheet" /><!--<![endif]-->
+        <link href="/assets/fonts.css" media="all" rel="stylesheet" />
 
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <meta property="og:image" content="/assets/opengraph-image.png">
-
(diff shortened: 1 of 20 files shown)


public/400.html
```diff
@@ -1,31 +1,19 @@
 <!DOCTYPE html>
-<!--[if lt IE 9]><html class="lte-ie8" lang="en"><![endif]-->
-<!--[if gt IE 8]><!--><html lang="en"><!--<![endif]-->
+<html lang="en">
     <head>
         <meta charset="utf-8" />
-        <title>This page cannot be found - GOV.UK Contact</title>
+        <title>Error - 400 - Bad request - Track-a-query</title>
```
willmcb, line 5: Im interested is there an accessibility reason to change the copy here - is it always a 400 that would cause this pare to load for instance?

public/400.html
```diff
@@ -52,9 +40,9 @@
                 <div id="wrapper" class="group live">
                     <section id="content" role="main">
                         <header class="page-header group">
-                            <h1 class="heading-xlarge">This page cannot be found</h1>
+                            <h1 class="heading-xlarge">Sorry, there is a problem with the service</h1>
                         </header>
-                        <p>Check that you’ve entered the correct web address or go to your Workbasket.</p>
+                        <p>Please contact <a href="mailto:correspondence@digital.justice.gov.uk?subject=400">correspondence@digital.justice.gov.uk</a> with the details on how you got here to report this issue.</p>
```
willmcb, line 45: Could having the subject as just "400" be a little ambiguous for a user? Would they know that that meant perhaps "400 error report" or something might be clearer.

public/400.html
```diff
@@ -81,21 +69,20 @@ <h1 class="heading-xlarge">This page cannot be found</h1>
...
-                                                                                                                            a.async = 1;
-                                                                                                                            a.src = g;
-                                                                                                                            m.parentNode.insertBefore(a, m)
-                                                                                                                        })
-                                                                                                                        (window, document, 'script', '//www.google-analytics.com/analytics.js', 'ga');
-                                                                                                                        ga('create', 'UA-37377084-57', 'auto');
-                                                                                                                        ga('send', 'pageview');</script>
+        <script>
+            (function (i, s, o, g, r, a, m) {
```
willmcb, line 73: This function is really difficult to understand, are their more descriptive names we could give these parameters? 

Perhaps extract the anon function to a named function as well as it would be clearer to read.

public/401.html
```diff
@@ -0,0 +1,88 @@
...
+                        <a href="http://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/copyright-and-re-use/crown-copyright/">© Crown copyright</a>
+                    </div>
+                </div>
+            </div>
+        </footer>
+
+        <div id="global-app-error" class="app-error hidden"></div>
+
+        <script>
+            (function (i, s, o, g, r, a, m) {
```
willmcb, line 73: could we extract this out to a module so it is not declared twice on separate pages, or is there a reason for doing it this way?

willmcb (changes requested):
I have left a few comments here as I'm not too sure about the function as it is has a lot of one char arguments in it which makes it hard to understand. Also the repetition of the function seems like it could be reduced by adding a partial and including it that way or creating a JS module.

Happy to chat about this if you'd like 👍

public/400.html
digitalali-moj: Yes so it should always load the correct page based on the error that Ruby thinks it is. The title has to reflect that for Google Analytics so we know what happened. The second part is just to be consistent with other pages.

public/400.html
digitalali-moj: Good point - I was thinking more from a dev perspective but you're right, I'll add "error report" as that sounds better.

public/400.html
digitalali-moj: This is the standard GA code so has to be applied as is so they say - It has a complicated history. I won't pretend to understand it haha. Happy to explore options of putting it somewhere and tidying it but that would have to be a site-wide change. Not sure if it should block this little piece though.

willmcb (approved):
👍

public/400.html
digitalali-moj: Done.

PR #1840, opened 2022-01-10
Add edit functionality + spec for SAR IR

willmcb:
## Description
Adds ability to edit cases in line with the acceptance criteria on the ticket.

## Self-review checklist

* [ ] (3) Tests passing

Create new case take it through the flow:
- Managers can edit 
- Responders cant 
- Approvers can edit as per ticket.

### DIFF
--- a/app/policies/case/sar/internal_review_policy.rb
+++ b/app/policies/case/sar/internal_review_policy.rb
@@ -12,7 +12,8 @@ class Case::SAR::InternalReviewPolicy < Case::SAR::StandardPolicy
   end
 
   def edit?
-    super || user_is_an_approver_on_case?
+    clear_failed_checks
+    check_can_trigger_event(:edit_case)
   end
 
   def can_close_case?
@@ -23,12 +24,6 @@ class Case::SAR::InternalReviewPolicy < Case::SAR::StandardPolicy
   def respond_and_close?
     clear_failed_checks
     user.managing_teams.include?(self.case.managing_team)
-  end
-
-  private
-
-  def user_is_an_approver_on_case?
-    self.case.approvers.include?(user)
   end
 end
 
--- a/app/policies/case/sar/internal_review_policy.rb
+++ b/app/policies/case/sar/internal_review_policy.rb
@@ -11,6 +11,10 @@ class Case::SAR::InternalReviewPolicy < Case::SAR::StandardPolicy
     check_can_trigger_event(:respond)
   end
 
+  def edit?
+    super || user_is_an_approver_on_case?
+  end
+
   def can_close_case?
     clear_failed_checks
     user.managing_teams.include?(self.case.managing_team)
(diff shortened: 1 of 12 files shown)


app/policies/case/sar/internal_review_policy.rb
```diff
@@ -11,6 +11,10 @@ def can_respond?
     check_can_trigger_event(:respond)
   end
 
+  def edit?
+    super || user_is_an_approver_on_case?
+  end
```
ymao2, line 16: From the consistency between the permission policy and workflow, IMO, it would be better to work with the following. lines
  ```
     def edit?
        clear_failed_checks
        check_can_trigger_event(:edit_case)
     end
```
it should work same as above old code , and should work for manager and approver

File: config/routes.rb [hunk 2]
```diff
@@ -115,6 +115,8 @@
 
     resources :sar_internal_review, only: only, controller: 'sar_internal_review', as: :case_sar_internal_review do
       get '/(:step)', on: :collection, to: 'sar_internal_review#new', as: 'step'
+      get '/edit', on: :member, to: 'offender_sar_complaint#edit' 
+      post '/update', on: :member, to: 'offender_sar_complaint#update'
```
ymao2, line 119: hmm... should it be `sar_internal_review#edit `and `sar_internal_review#update`?

app/policies/case/sar/internal_review_policy.rb
```diff
@@ -20,5 +24,11 @@ def respond_and_close?
     clear_failed_checks
     user.managing_teams.include?(self.case.managing_team)
   end
+
+  private
+
+  def user_is_an_approver_on_case?
+    self.case.approvers.include?(user)
+  end
```
ymao2, line 32: if you change the `edit?` to the one suggested above ( if it works , I didn't try it although I think it should work 😳 ), this function is not needed any more. Actually this one is duplicated with one in the `Workflows::Predicates#user_is_approver_on_case `

app/views/cases/sar_internal_review/_edit_case_details.html.slim
```diff
@@ -0,0 +1,23 @@
+- content_for :page_title, flush: true
+  = t('page_title.create_case_page.sar_internal_review.case_details_step')
```
ymao2, line 2: should  you change the title to some wording for `editing` rather than `Add`?

app/views/cases/sar_internal_review/_edit_case_details.html.slim
```diff
@@ -0,0 +1,23 @@
...
+
+- content_for :heading, flush: true
+  = t('cases.sar_internal_review.case_details')
```
ymao2, line 5: similar as above, you may need to change the wording

ymao2 (commented):
Nice work! 👍  only just a few comments for you to consider and double check

config/routes.rb
willmcb: Yes 🤦

app/views/cases/sar_internal_review/_edit_case_details.html.slim
willmcb: Thanks good spot!

ymao2 (approved):
LGTM :), thank you for the changes

PR #1847, opened 2022-01-17
CT-3998 Added user settings for db-anonymizer

ymao2:
## Description

This PR is to add extra json settings file around users for allowing us to keep certain users from live and add new users on QA env.  Also fix the some issues found for anonymizer -> setting up a pod just for anonlymizer due to the db connection to replication and the fact of using s3 bucket 
## Self-review checklist

* [X] (3) Tests passing

### Related JIRA tickets

https://dsdmoj.atlassian.net/jira/software/c/projects/CT/boards/25?modal=detail&selectedIssue=CT-3998


lib/db/database_loader.rb
```diff
@@ -10,16 +12,13 @@ def initialize(env, folder_for_load)
   def run
     drop_and_recreate_database
     load_database
+    user_settings = UsersSettingsForAnonymizer.new(nil)
```
willmcb, line 15: Could use a default `nil` argument here

willmcb (approved):
👍 LGTM

PR #1874, opened 2022-02-02
CT-3925: Add sar ir subtype to report

willmcb:
## Description
Amends the cases reports so as allow for SAR IRs. Adds new column to reports to show whether a SAR IR is for 'timeliness' or 'compliance'.

New column title is: 'SAR IR - timeliness or compliance'

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
[CT-3925](https://dsdmoj.atlassian.net/browse/CT-3925)

- you should already be able to search of SAR IR cases, so please check that you can.
- download the open cases report - you should be able to see a new column in the report that reports the SAR IR subtypes of 'timeliness' or 'compliance'


app/services/csv_exporter.rb
```diff
@@ -90,7 +91,7 @@ class CSVExporter
     'in_target',
     'number_of_days_taken',
     'number_of_days_late',
-    'number_of_days_taken_after_extension',
+    'number_of_days_taken_after_extension'
```
ymao2, line 94: Not sure about this line, did you mean a new field name?

ymao2 (commented):
I tried the PR, the open cases report looks,  the close report needs a little bit more work, I suggest to change the content of case_type for SAR IR rather than adding new field as it is in line with other case types and is simple. welcome more discussion,

app/services/csv_exporter.rb
willmcb: No I had a new field name here and then realised that it wasn't needed - I removed the comma as it is not needed in any case 👍

willmcb:
Closing as creating a new branch with slightly different approach

PR #1895, opened 2022-02-14
fix validation issues with email and postal address

willmcb:
## Description
The validations for email and postal address are still firing when they shouldn’t as the valid_attribute?

 method doesn’t respect the inline conditionals of :send_by_post? and :send_by_email?

This PR also creates a method for the email format validator as this was impacted in the same way.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
[CT-4087](https://dsdmoj.atlassian.net/browse/CT-4087)

- validations for postal / email address was firing if the message / upload field was left empty if even if the radio for each respectively was not checked.
- Create a new SAR IR leave the message field blank and then select email and add an email address. The validation for postal address should not fire
- do the same for email, select postal address and submit and the validation emails should not fire.

### DIFF
--- a/app/models/case/base.rb
+++ b/app/models/case/base.rb
@@ -205,7 +205,7 @@ class Case::Base < ApplicationRecord
   scope :updated_since, ->(date) { where('updated_at >= ?', date) }
 
   validates :current_state, presence: true, on: :update
-  validates :email, format: { with: /\A.+@.+\z/ }, if: -> { email.present? }
+  validate :validate_email_format
   validates_presence_of :received_date
   validates :type, presence: true, exclusion: { in: %w{Case}, message: "Case type cannot be blank" }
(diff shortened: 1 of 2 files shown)


app/models/case/base.rb
```diff
@@ -931,6 +932,15 @@ def validate_received_date
     errors[:received_date].any?
   end
 
+  def validate_email_format
+    if email =~ /\A.+@.+\z/ && email.present?
```
ymao2, line 936: just being very picky 😳 ,  it maybe better write as `email.present? && email =~ /\A.+@.+\z/`?

ymao2 (approved):
LGTM 👍 , just a tiny tiny comment for your consideration, but it won't prevent this PR

app/models/case/base.rb
willmcb: Yes that would be better I'll make that change! :)

PR #1903, opened 2022-02-17
Default updated / new BUs to have access to SAR IRs

willmcb:
## Description
Default updated / new BUs to have access to SAR IRs.

Fixes issue where BUs couldn't be assigned SAR IRs if they were new event though they could view SAR IR cases.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
[CT-4089](https://dsdmoj.atlassian.net/browse/CT-4089)
- create a new BU give it access to the SAR correspondence type
- Create a new SAR, it should be able to be assigned to the new BU.

### DIFF
--- a/app/controllers/cases_controller.rb
+++ b/app/controllers/cases_controller.rb
@@ -241,10 +241,10 @@ class CasesController < ApplicationController
   end
 
   def add_sar_ir_to_permitted_types_if_sars_allowed(types)
-    types.delete(CorrespondenceType.sar_internal_review)
-    if types.include?(CorrespondenceType.sar)
+    does_not_have_sar_ir = types.exclude?(CorrespondenceType.sar_internal_review)
+    has_sar = types.include?(CorrespondenceType.sar)
+    if has_sar && does_not_have_sar_ir 
       types << CorrespondenceType.sar_internal_review
     end
   end
-
 end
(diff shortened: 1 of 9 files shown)


app/controllers/cases_controller.rb
```diff
@@ -241,10 +241,10 @@ def correspondence_types_for_current_user
   end
 
   def add_sar_ir_to_permitted_types_if_sars_allowed(types)
-    types.delete(CorrespondenceType.sar_internal_review)
```
ymao2, line 244: @willmcb , this function don't need any more, you can remove this part entirely.  The idea is that if you choose to add the SAR IR into team's list when creating a new by default + data migration for existing team,  this part becomes redundant and also potentially cause inconsistent if there is change for case type permission . 

I tried your PR without this func, it works good ,  but it may affect the automated tests -> you may need to add this case type into the SAR team explicitly, give a try,  if too many cases are affected, you may keep this part, sorry for being picky 😳

willmcb: @ymao2 thanks for your comment - I was wondering about this but if you look at the way the types are collected from `current_user` you can see that it calls `menu_visible` which does not include the SAR IR correspondence type so I think that we need to specifically add it in in this case. It disappears from the "Create case" page if not.

ymao2: Hi @willmcb , I see, forgot the menu visible thing.

ymao2 (approved):
LGTM :)

PR #1971, opened 2022-05-03
Ct 4128 implement off sars closure logic

willmcb:
## Description
This PR updates the case closure logic for Offender SAR and Offender SAR Complaint cases so that as they are closed the case in question and its related cases receives a RetentionSchedule object which is used to mark a case for GRPR destruction etc.

For **Offender SARs** the way that cases receive retention schedules is like the following:

offendersarcases

Each receives a separate instance of RetentionSchedule with the same `planned_erasure_date` as the initial case that was closed.

For **Offender SAR Complaints** it is slightly more complicated:

- RS = RetentionSchedule

offsarcomplaints

When the complaint case is closed it receives a RetentionSchedule, and so does it's original case (i.e. the case that the initial complaint is about).

However, the original case can have more than one complaint against it. So each of these "other" complaints against the case also gets a RetentionSchedule with the same date as the initial complaint case, if each case is in a closed state.  

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
There are not many UI changes - only that the planned erasure date is now visible on the the show pages for each case:

Screenshot 2022-05-05 at 10 37 06

### Related JIRA tickets
[CT-4128](https://dsdmoj.atlassian.net/browse/CT-4128)
[CT-4129](https://dsdmoj.atlassian.net/browse/CT-4129)

For Offender SARs:
- Create some Offender SARs and link them, take one to closure, after one is closed both should have a planned erasure date on their separate show pages.

For Offender SAR complaints:
- Create a complaint case against an existing offender SAR case 
- Log into the console and find amend its created_at date to some time a few days ago (this is needed so that we can spot the difference in updates between different complaint cases later on. 
- then take it through to closure. 
- You should see a planned erasure date on both the Complaint Case, and Offender SAR cases' show page.
- Create another Complaint against the same Offender SAR case, and keep it open
- Create another Complaint against the same Offender SAR case and take it through to closure, once closed the planned_erasure_date on the first complaint case we created, and the Offender SAR case the complaint was about should all be the same on their respective show pages.
- However the open Complaint should not have a plan erasure date on its show page at all.


app/services/retention_schedules/add_schedule_service.rb
```diff
@@ -0,0 +1,88 @@
...
+        @kase.linked_cases.to_a
+      end
+    end
+
+    def planned_erasure_date
+      years = Settings.retention_timings.off_sars.erasure.years
+      months = Settings.retention_timings.off_sars.erasure.months
+      retention_period = years.years + months.months
+
+      closure_date + retention_period + 1.day
```
zheileman, line 80: Not sure if this is a real-life possibility, but as `closure_date` method can return `nil` if the case is not closed, when running this line of code will blow up with `NoMethodError: undefined method `+' for nil:NilClass`.

Because `planned_erasure_date` is executed on initialising the class, it means it will raise the exception to the `CaseClosureService` class and perhaps have an unexpected behaviour.

I think is just theoretical because by the time this code gets run the case _should_ have been marked as `closed` regardless. Then I suppose the method `closure_date` can just return the `@kase.last_transitioned_at.to_date` without any further checks and that way will never return `nil`?

Disclaimer: still getting my head around so apologies in advance if I'm getting this wrong! 😄

app/views/cases/shared/_closed_case_details.html.slim
```diff
@@ -67,3 +67,8 @@ tbody.response-details
       th = t('common.case.outcome')
       td = case_details.pretty_ico_decision
 
+  - if case_details.retention_schedule.present? && FeatureSet.branston_retention_scheduling.enabled?
+    tr.planned-erasure-date
+      th = "Planned erasure date"
```
zheileman, line 72: Super minor thing! String is not localised (was looking at the other strings in this file!)

willmcb: Thanks this is a good point - there is another ticket to update the wording away from "erasure" to destroy, so I just left this as is and will localise it when I do the update to the wording across the app. Was thinking that the key would probably include the word "erasure" if I did it now - probably just overthinking a minor thing though 🙂

app/services/retention_schedules/add_schedule_service.rb
willmcb: I think this is a really good spot - I was thinking that it would be as mentioned in your third paragraph that this _should_ only be called on closed cases - but I actually think it is possible as it gets reused across cases we cant be sure this won't be the case. I think we should think about what to do in this case 👍

zheileman (approved):
LGTM! 👍

PR #1989, opened 2022-05-12
Archive old migrations and fix data migration

zheileman:
## Description

This is a pre-requirement in order to work in another ticket.

In essence, the application code base is quite old, with the first migrations created back in 2016. As a result of this, back at the time a gem named `schema_plus_enums` was used to create the Postgres Enums. These old migrations called helper methods like `create_enum` and `drop_enum`.

More recently the application was upgraded to Rails 6.x and some other dependencies also bumped, which made the gem incompatible and newer migrations no longer use the helper methods, and instead just execute SQL to create/drop the enums.

The problem is, if a new developer joins the team and has to run the migrations, these will fail at the point where some of the old ones use `create_enum`.

A solution was to adapt those migrations and convert them to execute SQL code instead (like the newer ones), but we decided against this as it really is not necessary anymore to have migrations that old hanging around as these will never be run in reality and a rollback that old is not realistic.
So we've moved all migrations except the current year 2022 ones to a folder `db/archived_migrations` just for reference but not intended to be run.

**Migrations will only be run when there are new ones, this continues to work as expected.**

The README has been slightly updated to note this. The current setup of the application locally doesn't involve running migrations, but instead is done through the following commands, which uses the `structure.sql` and then runs data migrations to load seed data:

```
$ rails db:create
$ rails db:reseed
```

Additionally, as part of this reorganisation, another related issue was found, where a data migration can't run as it expects a specific DB migration to had run in the past for the DB to contain specific data/structure.
This has been worked around by merging together the DB migration into the Data migration so both runs sequentially, as they complement each other.

## Self-review checklist

* [ ] (3) Tests passing

In local environments, it is possible to ensure this runs as expected by dropping your local database, and recreating it and the data seed by running these commands. **Please be aware you will loose everything in your local database**:

```sh
RAILS_ENV=development rails db:drop
RAILS_ENV=development rails db:create
RAILS_ENV=development rails db:reseed
```


willmcb (approved):
LGTM 👍

PR #2009, opened 2022-05-31
Ct 4208 anonymise cases

willmcb:
## Description
PR is first of two to do with anonymisation for Branston cases. This PR includes a service that anonymises cases and its associated spec.

As such there is no user testing to be done just a code review.

A second PR will be raised to orchestrate individually feeding cases into this service to process their anonymisation.:w
 

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
n/a

### Related JIRA tickets
[CT-4208](https://dsdmoj.atlassian.net/browse/CT-4208)

- run the spec, and a code review please :)

### DIFF
--- a/app/services/retention_schedules/anonymise_case_service.rb
+++ b/app/services/retention_schedules/anonymise_case_service.rb
@@ -28,32 +28,32 @@ module RetentionSchedules
 
     ANON_VALUES = {
       case_reference_number: 'XXXX XXXX',
-      # date_of_birth: Date.new(01, 01, 0001),
-      # email: 'anon.email@cms-gdpr.justice.gov.uk',
-      # message: 'Information has been anonymised',
-      # name: 'XXXX XXXX',
-      # other_subject_ids: '',
-      # postal_address: 'Anon address',
-      # previous_case_numbers: '',
-      # prison_number: '',
-      # requester_reference: '',
-      # subject: 'XXXX XXXX',
-      # subject_address: 'Anon location',
-      # subject_aliases: '',
-      # subject_full_name: 'XXXX XXXX',
-      # third_party_company_name: '',
-      # third_party_name: 'Anon requester',
+      date_of_birth: Date.new(01, 01, 0001),
+      email: 'anon.email@cms-gdpr.justice.gov.uk',
+      message: 'Information has been anonymised',
+      name: 'XXXX XXXX',
+      other_subject_ids: '',
+      postal_address: 'Anon address',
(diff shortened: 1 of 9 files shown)


app/services/retention_schedules/anonymise_case_service.rb
```diff
@@ -0,0 +1,136 @@
...
+      requester_reference: 'XXXX XXXX',
+      subject: 'XXXX XXXX',
+      subject_address: 'XXXX XXXX',
+      subject_aliases: 'XXXX XXXX',
+      subject_full_name: 'XXXX XXXX',
+      third_party_company_name: 'XXXX XXXX',
+      third_party_name: 'XXXX XXXX',
+
+      # Offender SAR Complaint specific fields
+      gld_contact_email: 'XXXX XXXX',
```
zheileman, line 47: Does it need to comply to an email format, like the other attribute above `email: 'anon@email.com'`?

app/services/retention_schedules/anonymise_case_service.rb
```diff
@@ -0,0 +1,136 @@
...
+        @kase.save
+      end
+    end
+
+    private
+
+    def anonymise_core_case_fields
+      ANON_VALUES.each do |key, value|
+        if @kase.respond_to?(key)
+          @kase.update(key => value)
```
zheileman, line 91: Wondering if we should disable paper trail in all these updates, as we are going to delete all versions later on, seems like unnecessary.

```ruby
PaperTrail.request(enabled: false) { case update here }
```

app/services/retention_schedules/anonymise_case_service.rb
```diff
@@ -0,0 +1,136 @@
...
+        end
+      end
+    end
+
+    def anonymise_case_notes
+      notes = @kase.transitions.where(event: 'add_note_to_case')
+
+      if notes
```
zheileman, line 99: If you want, you can get rid of this condition, as `notes` is a relationship so can't be nil, and if empty, the `each` will not iterate regardless.

Same for `data_requests`.

app/services/retention_schedules/anonymise_case_service.rb
zheileman: Also, perhaps is safe to bypass validations and some other overhead here, by using `#update_attribute` instead of `#update`?

zheileman (approved):
👍 LGTM
Just a few suggestions/comments!

PR #2013, opened 2022-06-06
CT-2206 Enable Style/MutableConstant rubocop rule

zheileman:
## Description

This was an old ticket, so we either do it or remove the ticket and forget about it.
I think it is not a bad idea to do it. There are a lot of modified files but the changes are very specific.

This includes converting a mutating constant `TYPED_FILTER_FIELDS` into a class variable which is ugly but better at least keeps the functionality intact as the constant could not be frozen after applying this rubocop rule.

## Self-review checklist

* [ ] (3) Tests passing


willmcb (approved):
I agree this is worth doing, pretty quick change and might avoid future bugs and things 👍

PR #2014, opened 2022-06-06
Ct 4102 implement anon orchestration logic

willmcb:
## Description
Amends the `RetentionScheduleUpdateService` so that it calls the retention schedules `AnonymiseCaseService` to allow the user to anonymise multiple cases.

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
UI is essentially the same, should see a message "x cases have been destroyed" once case has been anonymised 

### Related JIRA tickets
[CT-4102](https://dsdmoj.atlassian.net/browse/CT-4102)

- go to the RRD page
- Click Ready for removal tab
- Select cases for anonymisation
- Click Destroy cases
- You should see a success message
- Those cases should no longer be visible
- If you then go to the show page for those cases (search case number in closed case tab) You should see that the details of the case have been anonymised.

### DIFF
--- a/app/services/retention_schedules/anonymise_case_service.rb
+++ b/app/services/retention_schedules/anonymise_case_service.rb
@@ -1,7 +1,7 @@
 module RetentionSchedules
   class CaseNotClosedError < StandardError
     def message
-      "Case cannoted be destroyed as it is still open"
+      "Case cannot be destroyed as it is still open"
     end
   end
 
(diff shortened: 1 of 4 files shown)


zheileman (approved):
👍 

Probably part of another piece of work, but it seems like anonymised cases can be edited though the UI, I can start complaint, etc. which doesn't make that much sense as I believe this cases are final and can't be touched anymore, kind of frozen.

But other than that, it does what it says in the tin 😄

PR #2016, opened 2022-06-08
Small bug fixes

willmcb:
## Description
Uses `date_responded` to calculate draft retention dates as this is more accurate. There are a few migrated cases that don't have `date_responded` so for these we use `received_date` instead.

Also fixes the date format on the show page.

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
New date format:
Screenshot 2022-06-08 at 15 25 42

### Related JIRA tickets
See comments on:
[CT-4127](https://dsdmoj.atlassian.net/browse/CT-4127)

And:
[CT-4215](https://dsdmoj.atlassian.net/browse/CT-4215)

- Date formats should be fixed on show page.
- On closing a case `planned_destruction_date` should be 8y6m + 1 day from the `date_responded` for a case

### DIFF
--- a/app/services/retention_schedules/add_schedule_service.rb
+++ b/app/services/retention_schedules/add_schedule_service.rb
@@ -90,11 +90,9 @@ module RetentionSchedules
     end
 
     def closure_date
-      if @kase.last_transitioned_at.nil?
-        @kase.transitions.most_recent.created_at.to_date
-      else
-        @kase.last_transitioned_at.to_date
-      end
+      # all cases recieve a date 
+      # responded on closure
+      @kase.date_responded
     end
   end
 end
--- a/app/services/retention_schedules/add_schedule_service.rb
+++ b/app/services/retention_schedules/add_schedule_service.rb
@@ -90,8 +90,6 @@ module RetentionSchedules
     end
 
     def closure_date
-      # all cases recieve a date 
-      # responded on closure
       if @kase.date_responded.present?
         @kase.date_responded
       else
--- a/app/services/retention_schedules/add_schedule_service.rb
+++ b/app/services/retention_schedules/add_schedule_service.rb
@@ -92,7 +92,11 @@ module RetentionSchedules
     def closure_date
       # all cases recieve a date 
       # responded on closure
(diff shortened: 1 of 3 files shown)


app/services/retention_schedules/add_schedule_service.rb
```diff
@@ -90,10 +90,12 @@ def planned_destruction_date
     end
 
     def closure_date
-      if @kase.last_transitioned_at.nil?
-        @kase.transitions.most_recent.created_at.to_date
+      # all cases recieve a date
```
zheileman, line 93: Super minor thing, but the comment feels misleading as it says all cases receive a date responded on closure, but there is a conditional in case it doesn't (which you mention in the PR can happen with some old cases) 😄

app/views/cases/filters/retention.html.slim
```diff
@@ -89,7 +89,7 @@ section#retention-cases.govuk-tabs__panel
                       td 
                         = kase.subject_name
                       td 
-                        = kase.retention_schedule.planned_destruction_date 
+                        = kase.retention_schedule.planned_destruction_date.strftime(Settings.default_date_format)
```
zheileman, line 92: As there is a default date format [defined in the locales](https://github.com/ministryofjustice/correspondence_tool_staff/blob/c6f51ee87e2413ec94964b8f26a93f7fc54704f0/config/locales/en.yml#L449) you can use `#l()` helper to localise the date:

```ruby
l(kase.retention_schedule.planned_destruction_date)
```

zheileman (approved):
👍 Super minor suggestions

app/services/retention_schedules/add_schedule_service.rb
willmcb: haha yes this doesn't make sense now, I wrote it when i thought it was the case and then forgot to do update it 🤦 

A micro lesson in why comments are not to be relied on!

app/views/cases/filters/retention.html.slim
willmcb: Thanks I'll do that 👍

PR #2021, opened 2022-06-13
CT-4134 Branston GDPR - Edit RRD details

zheileman:
## Description

Note: this PR was too big so I'm going to implement this in more than one PR, this is the first one that includes the editing of the destruction date. This does not include yet the editing of the retention status (review, retain...)

**Individual commits with more detail.**

A new edit page has been implemented, to edit the retention details of a case. To access this page, select a closed case that has already some retention schedule assigned and click the "Change" link in the case details.

As this is editing a relationship table (`has_one :retention_schedule`) the existing "framework" to edit case details wasn't ideal and I think it is best to just do this in a clean way with a form object. Also this setup will work fine with other pages in the future, or we can migrate some of the existing ones if we consider.

Most of this code is a streamlined version, with just the absolute minimum, of the form objects / steps mechanism I've implemented and iterated in many of my previous apps like C100, Disclosure Checker or Tax Tribunals, and it has proved robust and easy to work with, but happy to have a chat/demo/walkthrough if necessary.

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
Screenshot 2022-06-13 at 15 08 35

Screenshot 2022-06-13 at 15 09 33

Screenshot 2022-06-13 at 15 10 02

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-4134

Go to the details of a closed case with retention schedule, and click the Change link, it will open the retention details page where you can edit the date (only the date for now) and there is validation as well.
Note at the moment permissions are not being checked (it is enough you are signed in).

### DIFF
--- a/app/forms/base_form_object.rb
+++ b/app/forms/base_form_object.rb
@@ -1,0 +1,67 @@
+class BaseFormObject
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attr_accessor :record
+
+  # This will allow subclasses to define after_initialize callbacks
+  # and is needed for some functionality to work, i.e. acts_as_gov_uk_date
+  define_model_callbacks :initialize
+
+  def initialize(*)
+    run_callbacks(:initialize) { super }
+  end
+
+  # Initialize a new form object given an AR model, setting its attributes
+  def self.build(record)
+    attributes = attributes_map(record)
+
+    attributes.merge!(
+      record: record
+    )
+
+    new(attributes)
+  end
+
+  def save
+    valid? && persist!
+  end
+
+  def to_key
+    # Intentionally returns nil so the form builder picks up _only_
+    # the class name to generate the HTML attributes.
+    nil
+  end
+
+  def new_record?
+    true
+  end
+
+  # Add the ability to read/write attributes without calling their accessor methods.
+  # Needed to behave more like an ActiveRecord model, where you can manipulate the
+  # database attributes making use of `self[:attribute]`
+  def [](attr_name)
+    instance_variable_get("@#{attr_name}".to_sym)
+  end
+
+  def []=(attr_name, value)
+    instance_variable_set("@#{attr_name}".to_sym, value)
+  end
+
+  def attributes_map
+    self.class.attributes_map(self)
+  end
+
+  # Iterates through all declared attributes in the form object, mapping its values
+  def self.attributes_map(origin)
+    attribute_names.to_h { |attr| [attr, origin[attr]] }
+  end
+
+  private
+
+  # :nocov:
+  def persist!
+    raise 'Subclasses of BaseForm need to implement #persist!'
+  end
+  # :nocov:
+end
--- a/app/views/cases/filters/retention.html.slim
+++ b/app/views/cases/filters/retention.html.slim
@@ -88,7 +88,7 @@ section#retention-cases.govuk-tabs__panel
                 th 
                   = t('retention_schedules.table_headings.status')
               tbody
-                = form_with url: "/retention_schedules", method: :patch, id: 'retention_schedules_form' do |f|
+                = form_with url: bulk_update_retention_schedules_path, method: :patch, id: 'retention_schedules_form' do |f|
                   - @cases.each do |kase|
                     tr.case_row
                       td
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -169,7 +169,9 @@ Rails.application.routes.draw do
     end
   end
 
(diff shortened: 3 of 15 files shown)


app/forms/base_form_object.rb
```diff
@@ -0,0 +1,67 @@
+class BaseFormObject
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  attr_accessor :record
+
+  # This will allow subclasses to define after_initialize callbacks
+  # and is needed for some functionality to work, i.e. acts_as_gov_uk_date
+  define_model_callbacks :initialize
```
willmcb, line 9: 👍

app/forms/base_form_object.rb
```diff
@@ -0,0 +1,67 @@
...
+  # database attributes making use of `self[:attribute]`
+  def [](attr_name)
+    instance_variable_get("@#{attr_name}".to_sym)
+  end
+
+  def []=(attr_name, value)
+    instance_variable_set("@#{attr_name}".to_sym, value)
+  end
+
+  def attributes_map
```
willmcb, line 51: It would be good to have a chat about these attribute map methods - not sure I understand why there are a wrappers on the class methods is this just for ease of use? :)

app/views/cases/filters/retention.html.slim
```diff
@@ -87,7 +87,7 @@ section#retention-cases.govuk-tabs__panel
                 th 
                   = t('retention_schedules.table_headings.status')
               tbody
-                = form_with url: "/retention_schedules", method: :patch, id: 'retention_schedules_form' do |f|
+                = form_with url: bulk_update_retention_schedules_path, method: :patch, id: 'retention_schedules_form' do |f|
```
willmcb, line 90: 👍

config/routes.rb
```diff
@@ -169,7 +169,9 @@
     end
   end
 
-  patch '/retention_schedules', to: "retention_schedules#bulk_update"
+  resources :retention_schedules, only: [:edit, :update] do
```
willmcb, line 172: Much nicer 👍

spec/controllers/retention_schedules_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  let(:case_with_rrd) {
+    create(
+      :offender_sar_case, :closed, :with_retention_schedule,
+      planned_destruction_date: Date.tomorrow
+    )
+  }
+
+  let(:retention_schedule) { case_with_rrd.retention_schedule }
+
+  # TODO: implement proper access control and add tests here when ready.
```
willmcb, line 15: 👍 - probably easiest to follow the @user.team_admin? approach that I did in my most recent PR, but you've probably already thought about this :-)

spec/controllers/retention_schedules_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+        expect(assigns(:case)).to eq(case_with_rrd)
+        expect(assigns(:form_object).record).to eq(retention_schedule)
+
+        expect(response).to render_template(:edit)
+      end
+
+      context 'when the retention schedule does not exist' do
+        it 'raises an exception' do
+          expect {
+            get :edit, params: { id: 12345 }
```
willmcb, line 35: Are you sure the tests won't get this big 😂

spec/controllers/retention_schedules_controller_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  describe '#update' do
+    context 'when user has access' do
+      before do
+        sign_in team_admin
+      end
+
+      context 'when the update was successful' do
+        it 'redirects back to the case details page, showing a flash notice' do
```
willmcb, line 49: Nice

spec/requests/retention_schedules_spec.rb
```diff
@@ -1,7 +0,0 @@
-require 'rails_helper'
-
```
willmcb, line 2: Thanks - think this was the generator and then I forgot about it 🤦

willmcb (approved):
Really nice, I like the way you have built this out with separate form objects to hold logic related this specific action. It should also be easily extensible if need be 👍

app/forms/base_form_object.rb
zheileman: Oh I see what you mean. The instance method is not being used in this incarnation of the code, it is used un some more advanced code but I streamlined and removed anything that wouldn't apply immediately to CT and this edit functionality and this is a left over.

Will get rid of the instance method in the next PR. The only one needed right now is the class method 👍

PR #2024, opened 2022-06-14
Change scope of RRD pages

willmcb:
- makes the Pending tab also have 'destroy' cases that are in the future. Those that have passed their destruction dates and have a destroy status move to the 'Ready for removal' tab. 
- makes the Ready for removal page only have cases where the destruction
  date is >= Date.today (i.e. no future dates)
- cases that are marked as 'Destroy' will remain on Pending tab until their planned destruction date is passed at which point they pass to ready for removal tab.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
[CT-4220](https://dsdmoj.atlassian.net/browse/CT-4220)
[CT-4221](https://dsdmoj.atlassian.net/browse/CT-4221)

- change some cases destruction dates to fit inside and outside the rules specified above, they should be in the right places at the right times 👍

### DIFF
--- a/app/services/case_finder_service.rb
+++ b/app/services/case_finder_service.rb
@@ -111,29 +111,29 @@ class CaseFinderService
   def erasable_cases_scope
     retention_cases_scope.where(
       retention_schedule: { 
-        state: [:to_be_anonymised],
-        planned_destruction_date: ..Date.today
+        state: RetentionSchedule::STATE_TO_BE_ANONYMISED,
+        planned_destruction_date: RetentionSchedule.erasable_cases_viewable_range 
       })
   end
 
   def triagable_cases_scope
     triagable_statuses =  [
-      :review, 
-      :retain, 
-      :not_set
+      RetentionSchedule::STATE_REVIEW, 
+      RetentionSchedule::STATE_RETAIN, 
+      RetentionSchedule::STATE_NOT_SET
     ]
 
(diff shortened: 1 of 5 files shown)


app/services/case_finder_service.rb
```diff
@@ -111,15 +111,29 @@ def open_cases_scope
   def erasable_cases_scope
     retention_cases_scope.where(
       retention_schedule: { 
-        state: [:to_be_anonymised]
+        state: [:to_be_anonymised],
```
zheileman, line 114: I suggest using the (automatically generated) state constants everywhere. In this case this `:to_be_anonymised` would become `RetentionSchedule::STATE_TO_BE_ANONYMISED`.

It makes the code a lot less fragile, because if for any reason in the state machine we change the name of a state, the constants will break, whereas the symbols or strings depending on the code, might not break and continue working (just not as expected).

I think the tests are fine to use symbols or strings tho.

zheileman (approved):
👍 
I will update the filters in one of my next PRs, as I removed the "deleted" state from the pending to delete but now we can make use of it.

app/services/case_finder_service.rb
willmcb: Yep this makes perfect sense will amend to do this - this is the first time I've used aasm so I wasn't aware it made these constants available until I read your most recent PR 👍

PR #2029, opened 2022-06-16
Refactor the retention schedule case notes

zheileman:
## Description
As part of some previous PRs, we now have 2 places where we add retention schedule notes to the case history, and these messages relate to the same so we can unify and make the code nicer with a little presenter.

This presenter also handles the logic for which message (if any) to add to the history, as it can be only the `state` was changed but not the `planned_destruction_date` or the opposite, or even no changes at all.

This is done out of the box with ActiveRecord Dirty (`model.saved_changes` returns a map of the changes if any).

The presenter will also localise the date and translate the state. All quite easy to tweak if we want any different message in the future or even more attributes.

Note: this same mechanism can be applied to similar use-cases in the future. For now as we only use this for one note type (retention schedules) I've put everything in a class, but it has already almost the structure to use inheritance or some concern to be used for other kind of annotations based on attribute changes.

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
Screenshot 2022-06-16 at 15 34 54

Change some case retention schedules via the UI, in the RRD tabs or edit a retention schedule via their "Change" link, and observe the case message(s) being added depending on what you change.

### DIFF
--- a/app/controllers/retention_schedules_controller.rb
+++ b/app/controllers/retention_schedules_controller.rb
@@ -31,8 +31,8 @@ class RetentionSchedulesController < ApplicationController
 
   def update
     update_and_advance(RetentionScheduleForm, record: current_retention_schedule) do |form_object|
-      add_note_to_case!(
-        form_object.previous_values
+      annotate_case!(
+        form_object.record.saved_changes
       )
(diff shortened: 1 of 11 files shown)


willmcb (approved):
Really nice to have this on one place 👍

PR #2031, opened 2022-06-17
Update README.md

Davidslv:
## Description
This commit also changes the order, as it talks about testing the app before setting it up.
This was confusing so I changed it to read it down the line.


rohanssalunkhe (approved):
This sequence works and added missing env.sh file

PR #2039, opened 2022-06-21
CT-4219 Hide change links/buttons if case is readonly

zheileman:
## Description
This is a follow-up to PR #2032 where the backend side was introduced.

In this PR we hide from the UI functionality like change links, buttons, etc. that is no longer usable once a case has been anonymised and thus is in read only mode.

I might have missed some place but it easy to fix later on. And in any case even if the user tries to update something the backend side will stop them raising an exception.

Note: I've tried to piggy-back as much as possible on the already existing `allow_editing` partial propagation. However this affects mainly the Change links, not the buttons and some other actions, so I had to add some conditionals in some other places.

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
Screenshot 2022-06-20 at 15 45 54

Go to the details page for an anonymised case, and you should not be able to execute any update action (buttons, change links, etc. are hidden from the UI).

### DIFF
--- a/app/models/case/base.rb
+++ b/app/models/case/base.rb
@@ -463,6 +463,10 @@ class Case::Base < ApplicationRecord
     closed? && !!retention_schedule.try(:anonymised?)
   end
 
+  def editable?
+    !readonly?
+  end
+
   def to_csv
     CSVExporter.new(self).to_csv
   end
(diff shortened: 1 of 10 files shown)


willmcb (approved):
Nice one LGTM 👍

PR #2040, opened 2022-06-21
Make sort text for sort order reflect sort actions

willmcb:
## Description
The sort text for retention schedules was the opposite to what it was meant to be, clicking:

- "Sort by oldest destruction date" - sorted by newest
- "Sort by newest destruction date - sorted by oldest

This PR makes the text reflect the action

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
[CT-4232](https://dsdmoj.atlassian.net/browse/CT-4232)

Click the links on the RRD page you should see the page sort correctly based on the link you just clicked.

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -3,8 +3,8 @@ module SearchHelper
   SEARCH_SCOPE_SET = {
     :search_result_order_by_oldest_first => { "order" => "cases.id ASC"},
     :search_result_order_by_newest_first => {"order" => "cases.received_date DESC, cases.id DESC"},
-    :search_result_order_by_oldest_destruction_date => {"order" => 'retention_schedule.planned_destruction_date DESC'},
-    :search_result_order_by_newest_destruction_date => {"order" => 'retention_schedule.planned_destruction_date ASC'},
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -1112,8 +1112,8 @@ en:
     new_case_button: Create case
     show_search_result_order_by_newest_first: Show newest cases first
     show_search_result_order_by_oldest_first: Show oldest cases first
-    show_search_result_order_by_newest_destruction_date: Show newest destruction date first
-    show_search_result_order_by_oldest_destruction_date: Show oldest destruction date first
+    show_search_result_order_by_newest_destruction_date: Show oldest destruction date first
```
zheileman, line 1115: Wouldn't be better to leave the translation as before and just change where the translations are being used instead? As now it is confusing as the key says `order_by_newest` and the translation says `oldest first` 🤣 

Also not consistent with existing translations like `show_search_result_order_by_newest_first` and `show_search_result_order_by_oldest_first`

zheileman (approved):
👍

config/locales/en.yml
willmcb: Yes that would be better 😂 I'll do that

PR #2044, opened 2022-06-27
Rubocop: enable Style/WordArray

Davidslv:
I will be doing some rubocop fixes in smaller batches.
There are a lot of turned off cops so I'll do my best to keep PR's small and may do 1 or 2 cop fixes per PR.


zheileman:
@Davidslv I think we can exclude some paths like `spec`, `tasks`, `config`, `db`, etc? Enforcing rubocop on those is really annoying and pointless. Let's focus on pure application code.
In the old PR #1960 there are some exclusions applied, I think we can start with those ones and even add some more.

zheileman:
Also, not sure if it exists, but would be nice to find a ranking of the "top most dangerous offences", to focus on those ones (maybe top 5 or top 10), as not all offences are equally dangerous and some are completely innocuous and just a matter of personal preference (which I think is fine, leaving some scope to personal choices).

Davidslv:
> Also, not sure if it exists, but would be nice to find a ranking of the "top most dangerous offences", to focus on those ones (maybe top 5 or top 10), as not all offences are equally dangerous and some are completely innocuous and just a matter of personal preference (which I think is fine, leaving some scope to personal choices).

I will have a look into that, I have updated the PR to exclude those folders

zheileman (approved):
🥇

PR #2045, opened 2022-06-27
CT-4234 Add retention case notes on SAR closure

zheileman:
## Description
When a case is closed, the system calculates and updates the case's retention schedule. It also looks for any linked cases and updates the destruction date on all those related cases.

Upon any of these updates, a case note will be added to any affected case, stating the retention schedule changes, but noting this was a "System update" as it was not a direct action of the user (the user just closed a case).

If this is for new retention schedules, state will always be `Not set` (the default), so we don't add this to the case note, we only add the destruction date.

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
Screenshot 2022-06-27 at 11 08 40

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CT-4234
https://dsdmoj.atlassian.net/browse/CT-4235

On SAR case closure, a case history entry will be added to the case, and any linked case, stating it was a "System update".

### DIFF
--- a/app/presenters/retention_schedule_case_note.rb
+++ b/app/presenters/retention_schedule_case_note.rb
@@ -13,13 +13,25 @@ class RetentionScheduleCaseNote
     end
 
     def to_s
-      I18n.t(
-        self.class::ATTR_NAME, scope: scope, from: from, to: to
-      ) unless blank?
+      return if blank?
+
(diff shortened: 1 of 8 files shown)


Davidslv (approved):
This looks good to me 👍

PR #2046, opened 2022-06-27
Enabling linting capabilities of Rubocop

Davidslv:
This PR continues to enable more Rubocop linting features.
I have made a few manual changes but nothing that would break tests or change functionality.

I also tried to find the most problematic offences but that doesn't seem to be a thing people thought of.
Will continue to see where we can improve - I have skipped a few cops based on what I think is personal styling.


Davidslv:
Will need #2047 merged into this PR, before merging this one.

zheileman (approved):
💯

PR #2047, opened 2022-06-27
Continue rubocop enabling

Davidslv:
I'm issuing this PR against #2046 so that it's easier to see changes.

Here I have turned a few cops which did not had anything to correct - which is good, enabling them keeps future people from actually creating new offences.

Most of this PR is about whitespacing and changing to the ruby 1.9 hash syntax (instead of `=>`).


zheileman (approved):
Good lord. How many extraneous spaces, blank lines, etc. It would be difficult to do this even on purpose 🤣 

LGTM. I just found a file where the fixing didn't work well, but it is something minor. I added a comment for that.

PR #2049, opened 2022-06-28
[CT-2395] Enable Rubocop Rails

Davidslv:
Closes #1960

This will be the last mass enabling/disabling of Rubocop, this enables the rubocop-rails gem to run with its own rules.

I'd suggest that at a later stage some of the disabled cops for Rails are enabled again.

Apologies for the long PR

### DIFF
--- a/app/models/correspondence_type.rb
+++ b/app/models/correspondence_type.rb
@@ -99,45 +99,47 @@ class CorrespondenceType < ApplicationRecord
     CorrespondenceType.properties_where_not(report_category_name: '').order(Arel.sql("(correspondence_types.properties -> 'report_category_name') asc"))
   end
 
+  # rubocop:disable Rails/DynamicFindBy
   def self.foi
-    find_by! abbreviation: 'FOI'
+    find_by_abbreviation! 'FOI'
   end
 
   def self.gq
-    find_by! abbreviation: 'GQ'
+    find_by_abbreviation! 'GQ'
   end
 
   def self.sar
-    find_by! abbreviation: 'SAR'
+    find_by_abbreviation! 'SAR'
   end
 
   def self.offender_sar
-    find_by! abbreviation: 'OFFENDER_SAR'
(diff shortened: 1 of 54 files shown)


zheileman:
I feel like this is too much all at once.

I suggest disabling the `Rails/TimeZone` cop as a first step, as this is introducing a lot of changes for a legacy codebase like this, adding noise, so we can focus on the important ones. After all, the app has been working fine for 7 years even without the `zone`.

Also I believe rubocop might not be smart enough to do some auto-fixes, particularly around relationships, as there are lots of failures for this on CircleCI.

Let's try to relax those cops to get down the number of changes hopefully to something that can be reviewed 😬

zheileman (approved):
💯 I've checked all the changes and I think these are all safe.
Maybe good to squash or rebase the commits, as there are many, to not create a lot of noise in the history? Up to you! 🥇

app/models/correspondence_type.rb
```diff
@@ -100,39 +100,39 @@ def self.by_report_category
   end
 
   def self.foi
-    find_by_abbreviation! 'FOI'
+    find_by! abbreviation: 'FOI'
```
zheileman, line 103: Oh I think this is a false positive from rubocop. This is suppose to be calling a custom class method defined above `.find_by_abbreviation!` so probably we want to keep it as it is.

app/services/case_require_further_action_service.rb
```diff
@@ -15,7 +15,7 @@ def initialize(user, kase, update_parameters)
   def call
     begin
       ActiveRecord::Base.transaction do
-        update_attributes
+        update
```
zheileman, line 18: Same here, false positive, this is calling a custom method defined below, `#update_attributes`

zheileman: Probably better to rename the method to `do_update` or anything like that so this can't happen in the future 😬

PR #2050, opened 2022-06-28
[CT-2395] Enable Rubocop Rails

Davidslv:
Should close #1960
Superseeds #2049 (we used what we learned from there and applied it to this PR)

I've decided to turn off most rails cops, this will make it easier to turn them on and tackle any issues one by one.
This PR is more readable this way and future PR's that turn cops on will be too.


zheileman (approved):
🥳

PR #2066, opened 2022-08-08
Tag latest tag with latest and make anonymizer use it

digitalali-moj:
## Description
Make anonymizer cronjob run by adding latest tag back into build process and making the cronjob use that tag

## Self-review checklist

* [ ] (3) Tests passing


build.sh
```diff
@@ -75,10 +75,20 @@ function _build() {
...
+      latest_tag=${docker_registry}:${component}-${current_branch}-latest
+      ;;
+  esac
+  docker tag $docker_registry_tag $latest_tag
+
+  p "Latest tag name (used for Cron jobs that don't know tag hashes on spin up..."
+  p $latest_tag
 
   p "Beginning push to ECR..."
+  docker push $latest_tag
```
Polsonby, line 91: This looks fine, just what we talked about, but looking at it, I wonder if we want or need the non-main branches tagged as latest - we'll never use them and it may take up more ECR space.  🤔

Instead of the case with two branches, could we potentially put the latest bit in a conditional, something like this?

```
p "Push branch to ECR..."
docker push ${docker_registry_tag}

if [[ $current_branch == "main" ]]
    p "...and tagging main as latest for Cron jobs that don't know tag hashes on spin up"
    latest_tag=${docker_registry}:${component}-latest
    docker tag $docker_registry_tag $latest_tag
    docker push $latest_tag
fi
p "...push to ECR complete"
```

config/kubernetes/production/cronjob-anonymizer.yaml
```diff
@@ -13,7 +13,7 @@ spec:
         spec:
           containers:
           - name: jobs
-            image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/correspondence/track-a-query-ecr:latest
+            image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/correspondence/track-a-query-ecr:track-a-query-latest
```
Polsonby, line 16: 👍

Polsonby (approved):
Looks good! 👍

PR #2078, opened 2022-09-28
Output validation errors on the new Overturned ICO case page

Polsonby:
## Description
When validation errors occur on the Overturned ICO new case form, they don't seem to be output. This PR should allow them to display 
## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/app/views/cases/overturned_shared/_new_form.html.slim
+++ b/app/views/cases/overturned_shared/_new_form.html.slim
@@ -1,6 +1,10 @@
 - content_for :page_title do
   = t('page_title.create_overturned_ico_page')
 
+= GovukElementsErrorsHelper.error_summary @case,
+    "#{pluralize(@case.errors.count, t('common.error'))} #{ t('common.summary_error')}",
+    ""
+
 section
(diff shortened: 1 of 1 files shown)


rohanssalunkhe (approved):

app/views/cases/overturned_shared/_new_form.html.slim
```diff
@@ -1,6 +1,10 @@
 - content_for :page_title do
   = t('page_title.create_overturned_ico_page')
 
+= GovukElementsErrorsHelper.error_summary @case,
+    "#{pluralize(@case.errors.count, t('common.error'))} #{ t('common.summary_error')}",
+    ""
```
wilson1000-MoJ, line 6: Hiya, are the empty quotes expected here?

Polsonby: Yes, the carriage return before them is a bit unexpected, isn't it? It's the same exact code as other instances in our view partials, though. I expect it's laid out like this as a code formatting thing to meet line length limits.

PR #2079, opened 2022-09-29
CDPT-90 Add request method to warehouse table

Polsonby:
## Description
For efficient reporting on the origin of subject access requests, we need to add a new column to the "data warehouse" table and populate it to store the request_method property from existing SAR cases. We also need to ensure that when the warehouse sync task runs on new or updated cases in the future, the warehouse value is also updated

This PR:

* Adds a database migration to add the column to warehouse table and the new resulting structure.sql
* Adds a data migration that executes a SQL query to populate the new column
* Updates the warehouse/case_report class which does the sync to update the request_method warehouse field for a case when it is modified 

## Self-review checklist

* [x] (3) Tests passing

### Screenshots
image

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-90

### Deployment
This PR includes both a database and data migration which are run against a database when we deploy to a namespace for the first time. See lines 10 and 11 of config/docker/entrypoint-webapp.sh

After checking out your branch, ensure that you have recent SAR cases created in your development database, either or both of SAR and OFFENDER-SAR

Run `bundle exec rails db:migrate` and then `bundle exec rails data:migrate`

You should see a new column on the end of the `warehouse_case_reports` table with values matching relevant SAR-style cases. The value should be NULL for any other case types.

### DIFF
--- a/app/models/warehouse/case_report.rb
+++ b/app/models/warehouse/case_report.rb
@@ -99,6 +99,7 @@ module Warehouse
         case_report.original_internal_deadline = kase.respond_to?(:original_internal_deadline) ? kase.original_internal_deadline : nil
         case_report.original_external_deadline = kase.respond_to?(:original_external_deadline) ? kase.original_external_deadline : nil
         case_report.num_days_late_against_original_deadline = kase.respond_to?(:original_external_deadline) ? kase.num_days_late_against_original_deadline : nil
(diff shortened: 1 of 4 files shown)


wilson1000-MoJ (approved):
LGTM :)

PR #2086, opened 2022-10-14
CDPT-276 Run migrations in Kubernetes Job

vertism:
## Description
Currently long running migrations can fail because Kubernetes will kill the container due to the healthcheck failing after the time allowed for the readinessProbe has expired.

The following changes have been made:
1. Migration tasks have been removed from the deployment script and added to a dedicated Kubernetes Job which will run `config/docker/entrypoint-migrations.sh`. This job is run as part of the deploy script.
2. A new [Init Container](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) has been added to each pod. They runs a `rake db:abort_if_pending_migrations` task so that the pod will not start up until the schema migrations are completed [as described here](https://thoughtbot.com/blog/zero-downtime-rails-deployments-with-kubernetes#init-containers).
3. We've bumped the memory requested by the migrations job container because we were hitting OOMKill errors when running heavy data migrations.

## Self-review checklist

* [ ] (3) Tests passing


config/kubernetes/demo/deployment.yaml
```diff
@@ -174,3 +174,50 @@ spec:
...
+            - name: AWS_ACCESS_KEY_ID
+              valueFrom:
+                secretKeyRef:
+                  name: track-a-query-s3-output
+                  key: access_key_id
+            - name: AWS_SECRET_ACCESS_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: track-a-query-s3-output
+                  key: secret_access_key
```
Polsonby, line 218: I don't think the migrations container requires these env vars. In the interest of reducing boilerplate, could we try without these and see if it's happy?

config/kubernetes/production/deployment.yaml
```diff
@@ -21,7 +21,7 @@ spec:
         - name: webapp
           image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/correspondence/track-a-query-ecr:latest
           ports:
-          - containerPort: 3000
+            - containerPort: 3000
```
Polsonby, line 24: Just checking, was this additional whitespace intended?

vertism: It was, just to make the yaml consistent between envs.

config/kubernetes/demo/deployment.yaml
vertism: OK, I'll test without. Thanks.

Polsonby:
Just a thought - It might be good to add a little bit more context to the description on the PR for colleagues visiting from the future to clarify that the two "legs" of this are 

1. A new Kubernetes job that is kicked off explicitly from the deployment.yml to start running migrations, instead of within the webapp container entrypoint script
2. A new initcontainer which is added to each pod that runs a `rake db:abort_if_pending_migrations` task so that the pod will not start up until the migrations are completed [as described here](https://thoughtbot.com/blog/zero-downtime-rails-deployments-with-kubernetes#init-containers)
3. And we've bumped the memory requested by the container because we were hitting OOMKill errors when running heavy data migrations. 

Final thought, I see there is a `rake data:abort_if_pending_migrations` task for the [data_migrations](https://www.rubydoc.info/gems/data_migrate) gem. What do we think about adding that to the initcontainer?

vertism:
> Final thought, I see there is a `rake data:abort_if_pending_migrations` task for the [data_migrations](https://www.rubydoc.info/gems/data_migrate) gem. What do we think about adding that to the initcontainer?

I did see that during dev and tried it, which led me to discover that we are actually using a different gem which doesn't have that method - https://github.com/OffgridElectric/rails-data-migrations

vertism:
> Just a thought - It might be good to add a little bit more context to the description on the PR for colleagues visiting from the future to clarify that the two "legs" of this are

I've the title and description to hopefully be more useful. (I copied some of your words so thanks!)

Polsonby (approved):
Great work, thanks for the hard work on this

PR #2103, opened 2022-11-04
CDPT-334 Remove timestamp from filename

vertism:
## Description
The users no longer wish to see the date and time the file was generated at as it can be confusing for some staff as it looks very similar to the case ref number and they would like to avoid errors.

## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/app/models/commissioning_document.rb
+++ b/app/models/commissioning_document.rb
@@ -32,7 +32,7 @@ class CommissioningDocument
   def filename
     return unless valid?
 
-    "Day1_#{request_type}_#{case_number}_#{subject_name}_#{timestamp}.#{mime_type}"
+    "Day1_#{request_type}_#{case_number}_#{subject_name}.#{mime_type}"
   end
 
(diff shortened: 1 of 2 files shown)


app/models/commissioning_document.rb
```diff
@@ -32,7 +32,7 @@ def document
   def filename
     return unless valid?
 
-    "Day1_#{request_type}_#{case_number}_#{subject_name}_#{timestamp}.#{mime_type}"
```
wilson1000-MoJ, line 35: Hi Andrew (@vertism), are you happy the removal of the timestamp won't cause a problem such as duplicate filename issues?

vertism: It certainly could in theory. The request to change this came from the users, but I will check that they understand this risk. Thanks.

Polsonby (approved):
LGTM :)

vertism:
This change is no longer required

PR #2109, opened 2022-11-16
CDPT-365 Handle RecordNotUnique error

vertism:
## Description
When a case is created, a WarehouseCaseReport is created and associated to the case. This happens asynchronously via a Redis job.
The WarehouseCaseReport is kept in sync with the case via Jobs whenever the case is updated.

If a new case is immediately updated after creation a race condition can occur because the update case object is not aware of the associated WarehouseCaseReport. So the sync code tries to create a new WarehouseCaseReport which fails because the Primary Key for the table is the case ID which already exists.

The solution implemented here is to rescue `ActiveRecord::RecordNotUnique` exceptions and retry. This works because the Case is reloaded which will load the associated WarehouseCaseReport which has just been created, rather than trying to create a new one.

## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/app/models/warehouse/case_report.rb
+++ b/app/models/warehouse/case_report.rb
@@ -106,8 +106,9 @@ module Warehouse
           process_class_related_process(kase, case_report)
           case_report.save!
           case_report
-        rescue ActiveRecord::RecordNotUnique => err
+        rescue ActiveRecord::RecordNotUnique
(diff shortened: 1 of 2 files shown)


Polsonby (approved):
Really nice solution, great work 👏

PR #2110, opened 2022-11-17
CDPT-368 Allow assigning to non-branston users

vertism:
## Description
When an offender SAR complaint is created, the case is auto assigned to the creator. This fails if the creator does not belong to the team that is assigned to that correspondence type. There is a team from SSCL who also create these cases and the assigning fails for them and causes an exception.

This change removes the restriction and means anyone that can create an offender SAR complaint will can be assigned to it. As well as stopping the exception, this will also mean these users can see their assigned cases in the "My Open Cases" tab.

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
Error seen before change
Screenshot 2022-11-16 at 17 05 48

A user with the email darkwin@digital.justice.uk can be used in development to test this by creating an offender SAR complaint. With the changes applied, the case will be created and no error will be seen.

### DIFF
--- a/config/state_machine/configs/00_moj/50_offender_sar_complaint/20_offender_sar_complaint_standard_workflow.yml
+++ b/config/state_machine/configs/00_moj/50_offender_sar_complaint/20_offender_sar_complaint_standard_workflow.yml
@@ -17,9 +17,8 @@
                 preview_cover_page:
                 edit_case:
                 assign_to_team_member:
-                  if: Workflows::Predicates#responder_is_member_of_assigned_team_and_not_assigned?
+                  if: Workflows::Predicates#responder_is_not_assigned?
                 reassign_user:
-                  if: Workflows::Predicates#responder_is_member_of_assigned_team_and_assigned?
                 reset_to_initial_state:
                   transition_to: to_be_assessed
               data_review_required:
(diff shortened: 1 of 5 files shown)


config/state_machine/configs/00_moj/50_offender_sar_complaint/20_offender_sar_complaint_standard_workflow.yml
```diff
@@ -17,9 +17,8 @@
                 preview_cover_page:
                 edit_case:
                 assign_to_team_member:
-                  if: Workflows::Predicates#responder_is_member_of_assigned_team_and_not_assigned?
+                  if: Workflows::Predicates#responder_is_not_assigned?
                 reassign_user:
-                  if: Workflows::Predicates#responder_is_member_of_assigned_team_and_assigned?
```
Polsonby, line 22: Probably being dim, but I can't figure out why there were two different predicate methods used for the conditional on reassign_user - there's this one and `responder_is_member_of_assigned_team?` - do you understand why?

We've taken out all the predicate methods for `reassign_user`in this workflow anyway, so I guess it's moot!

Polsonby (approved):
Looks good to me! One question in the comments there for my own understanding but happy to approve 👍

config/state_machine/configs/00_moj/50_offender_sar_complaint/20_offender_sar_complaint_standard_workflow.yml
vertism: It's a fair question. There was a check on whether the was already a user assigned for 2 of the states (`legal_proceedings_ongoing` and `to_be_assessed`), but not for the others. 

I couldn't find a good reason for it, and as there shouldn't be a way for a case to be created without someone assigned, I think it is safe to not check that now anyway.

PR #2114, opened 2022-11-18
CDPT-370 Avoid updating old cases when reindexing

vertism:
## Description
When a case is updated a reindex is triggered and the case is marked as "dirty"
After a case is reindexed the case is set as “clean” to stop unnecessary indexing.  
Some old cases fail validation, so updating causes an exception.

Reindexing can also be triggered when a case has not been updated, and in that instance there is no need to update the object and set it as "clean"

## Self-review checklist

* [ ] (3) Tests passing


Polsonby (approved):
Fab, nice one 👍

PR #2117, opened 2022-11-22
Refactor Business Days Deadline Calculator

vertism:
## Description
The aim of this refactor is to remove duplication of logic by moving it into the `calculate` method, and simplifying methods where possible.

## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/app/models/case/ico/base.rb
+++ b/app/models/case/ico/base.rb
@@ -46,7 +46,7 @@ class Case::ICO::Base < Case::Base
                  ico_decision: :string,
                  ico_decision_comment: :string,
                  late_team_id: :integer,
-                 date_draft_compliant: :date, 
+                 date_draft_compliant: :date,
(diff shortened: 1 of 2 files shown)


app/models/case/ico/base.rb
```diff
@@ -258,7 +258,7 @@ def set_deadlines
     # For ICOs deadlines are manually set and don't need to be automatically
     # calculated. So this method called by a before_update hook in Case::Base
     # becomes a nop.
-    self.internal_deadline ||= @deadline_calculator.internal_deadline
+    nil
```
vertism, line 261: This is set manually as the comment says, so this is not needed.
Looking at the history it appears to have been kept in temporarily during some changes, but never got removed.

Polsonby (approved):
Changes look really good, thanks for this 👍

PR #2133, opened 2022-12-20
Speed up tests

vertism:
#### Using build_stubbed
Use [FactoryBot.build_stubbed](https://thoughtbot.com/blog/use-factory-bots-build-stubbed-for-a-faster-test) to instantiate models. This reduces DB calls and therefore speeds up the test suite.

#### Group tests more efficiently
A log of test runtimes has been added, so that tests can be grouped more evenly when run in parallel. This log was generated in CI.

#### Results 
Previously the test suite took 16-18 minutes on CI, now it takes 9-10 minutes. So around 45% faster.

## Self-review checklist

* [ ] (3) Tests passing


Polsonby (approved):
Nice work, thanks for this!

PR #2135, opened 2022-12-22
CDPT-337 Allow managers to remove responses

vertism:
## Description
Allows users who are managers to remove responses for FOI and ICO cases when the case is in some states. 
Sensitive data is sometimes uploaded by other users by mistake, and currently this has to be manually removed via the console.

The logic for how this is decided has also been refactored so that it is all done within the state machine, and not partly via policy classes.

## Self-review checklist

* [ ] (3) Tests passing


wilson1000-MoJ (approved):
Looks good :)

PR #2136, opened 2022-12-23
CDPT-297 Add cronjob for resetting sidekiq queues

digitalali-moj:
## Description
I couldn't find a way to automatically delete Redis jobs so hoping maybe this cronjob will do it. Quite crude atm but wanted to see if the approach is OK. Won't merge into main until I've tested this a bit more but thought to see if it's the way to go or not first. It's supposed to delete completed jobs but it's scheduled at a time that there shouldn't be jobs in progress anyway just to be cautious. 

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-297

### Deployment
Will need to be tested manually from local before releasing into pipeline

Run cronjob locally by changing time to immediate and keep an eye on production.

### DIFF
--- a/lib/tasks/sidekiq.rake
+++ b/lib/tasks/sidekiq.rake
@@ -21,5 +21,15 @@ namespace :sidekiq do
     end
 
   end
+
+  desc 'Clear sidekiq processed and failed queues to save memory'
+  task :clearQueues => :environment do
+
+    require 'sidekiq/api'
+
+    Sidekiq::Stats.new.reset
+    Sidekiq::DeadSet.new.clear
+
+  end
 end
 
--- a/config/kubernetes/production/cronjob-sidekiq-queues.yaml
+++ b/config/kubernetes/production/cronjob-sidekiq-queues.yaml
@@ -1,0 +1,32 @@
+apiVersion: batch/v1beta1
+kind: CronJob
+metadata:
+  name: clear-sidekiq-queues
+spec:
+  schedule: "0 5 * * 1"
+  concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 3
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: jobs
+            image: 754256621582.dkr.ecr.eu-west-2.amazonaws.com/correspondence/track-a-query-ecr:track-a-query-latest
+            imagePullPolicy: IfNotPresent
+            command:
+              - sh
+              - "-c"
+              - "bundle exec rake sidekiq:clearQueues"
+            env:
+              - name: REDIS_URL
+                valueFrom:
+                  secretKeyRef:
+                    name: track-a-query-elasticache-redis-output
+                    key: url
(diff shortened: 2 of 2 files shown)


lib/tasks/sidekiq.rake
```diff
@@ -21,5 +21,15 @@ namespace :sidekiq do
...
 
   end
+
+  desc 'Clear sidekiq processed and failed queues to save memory'
+  task :clearQueues => :environment do
+
+    require 'sidekiq/api'
+
+    Sidekiq::Stats.new.reset
+    Sidekiq::DeadSet.new.clear
```
vertism, line 31: Looking at the docs, by default failed jobs are stored in the dead set for 6 months, and a maximum of 10,000 of them. I can understand that if this filled up it could cause an OOM issue, although looking at prod there are only 34 jobs in there at the moment.

I think those numbers can be changed in config too, although I'm not exactly sure how. Maybe a simpler way to handle this if we think that's the issue?

lib/tasks/sidekiq.rake
```diff
@@ -21,5 +21,15 @@ namespace :sidekiq do
     end
 
   end
+
+  desc 'Clear sidekiq processed and failed queues to save memory'
+  task :clearQueues => :environment do
+
+    require 'sidekiq/api'
+
+    Sidekiq::Stats.new.reset
```
vertism, line 30: Does this do more than reset the counters? I'm not clear on how this would reduce memory usage, or does it actually remove the jobs too?

digitalali-moj: I'm under the impression that this is supposed to clear the jobs from Redis that are completed, based on what I've read but not too sure so experimenting. One of the places was here: https://gist.github.com/wbotelhos/fb865fba2b4f3518c8e533c7487d5354
I need to monitor the memory usage and see if this makes any difference. Just checking if this is an option though or if there's a better approach.

lib/tasks/sidekiq.rake
digitalali-moj: OK thanks, will see if I can find a way to do this in the config. We may need to have a discussion around this https://docs.google.com/document/d/1lKfSbCZehTQALoHCvdGZcNwtK65_a_Az-c54-T20YHw/edit# when Ian is back since 34 jobs doesn't seem a lot, yet there were lots on development before I cleared it like thousands in the processed.

lib/tasks/sidekiq.rake
vertism: Looking at the code - https://github.com/mperham/sidekiq/blob/main/lib/sidekiq/api.rb#L158-L170 - I _think_ it's just changing some keys in Redis which are used to hold the stat information.

lib/tasks/sidekiq.rake
vertism: There is a screenshot of the stats count, but most values are 0, so I presume that was run after the flushdb command was run.
Looking at the memory usage graph in that link, it shows fairly consistent rises for a year. If dead jobs are removed after 6 months I think it would look different.

PR #2155, opened 2023-03-07
CDPT-581 Use method instead of proc

vertism:
## Description
An exception was being raised when a user submitted a message without any text. This was caused by a strange bug whereby some internal rails code was trying to serialise the case_transition object but was failing because it contained a proc that was being used to perform validation on the message attribute.

The issue is fixed by changing the validation to use a method instead of the proc. 

## Self-review checklist

* [ ] (3) Tests passing


Polsonby (approved):
Great, good spot 👍 :) LGTM

PR #2158, opened 2023-03-13
Fix request_type migration for local development

vertism:
## Description
When `rake db:reseed` is run locally, the local DB is cleared, then structure.sql is loaded. After that data migrations are run.
This was causing an issue as there was a data migration for the `request_types` type which was out of date and was replacing the correct data.

This migration has been removed.

## Self-review checklist

* [ ] (3) Tests passing


wilson1000-MoJ (approved):
Looks good @vertism - echos in code what is described :) well done.

PR #2160, opened 2023-03-15
CDPT-417 CT (London Disclosure): Validation when creating ICO overturned case

MeenaModhvadia:
## Description
Amended time limit of creating an ico overturned case from 29 to 40 days.

## Self-review checklist

* [ ] (3) Tests passing


spec/models/case/overturned_ico/foi_spec.rb
```diff
@@ -106,22 +106,22 @@
       context 'too far in the past' do
         context 'new record' do
           it 'errors' do
-            new_case.received_date = 29.days.ago
+            new_case.received_date = 41.days.ago
             expect(new_case).not_to be_valid
-            expect(new_case.errors[:received_date]).to eq ['is too far in the past']
+            expect(new_case.errors[:received_date]).to eq ['decision received from ICO must be within 40 days of today']
           end
```
wilson1000-MoJ, line 112: SORRY! I just looked at the language file and realised the FOI wasn't affected. Your tests might break now; if so, simply roll back the update to FOI... apologies @MeenaModhvadia

spec/models/case/overturned_ico/foi_spec.rb
```diff
@@ -106,22 +106,22 @@
...
-            expect(record.errors[:received_date]).to eq ['is too far in the past']
+            record = described_class.create(received_date: 41.days.ago)
+            expect(record.errors[:received_date]).to eq ['decision received from ICO must be within 40 days of today']
           end
         end
 
         context 'existing record' do
           it 'does not error' do
-            record = described_class.create(received_date: 29.days.ago)
+            record = described_class.create(received_date: 40.days.ago)
```
vertism, line 124: This is testing that the received_date validation does not apply to existing records, so I think it should still be using 41.days.ago. 
Same in the instance below.

MeenaModhvadia: I have amended the code as per above.

vertism (approved):

PR #2168, opened 2023-04-04
CDPT-571 Restriction dates on the MAPPA template

MeenaModhvadia:
Include restriction dates on the MAPPA template


app/models/commissioning_document_template/mappa.rb
```diff
@@ -8,6 +8,7 @@ def context
       super.merge(
         aliases: kase.subject_aliases,
         pnc: kase.other_subject_ids,
+        date_range: data_request.request_dates.capitalize,
```
vertism, line 11: The screenshot on the ticket doesn't have the first character capitalised
```suggestion
        date_range: data_request.request_dates,
```

vertism (approved):
👍

PR #2169, opened 2023-04-04
CDPT-569 Restriction date content on cross boarder transfer template

MeenaModhvadia:
Move date range (pre-populated) to be incorporated in first paragraph.


vertism (approved):
LGTM

PR #2207, opened 2023-05-16
Restrict commissioning document filetype to docx

vertism:
## Description
Commissioning Documents are created as .docx, so when replaced the uploaded file should also be a .docx.
A warning will appear if a different type of document is uploaded.

### Screenshots
Screenshot 2023-05-17 at 11 08 23


njpreddy (approved):
LGTM 👍

PR #2230, opened 2023-07-03
cdpt 784 Prevent user sending Probation records email when selected Organisation has no email

lucas-shaw:
## Description
If recipient emails are empty (regardless of whether the branston registry email is present), ensures that the moj info banner and email cannot be sent

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?modal=detail&selectedIssue=CDPT-784

### DIFF
--- a/app/helpers/data_request_helper.rb
+++ b/app/helpers/data_request_helper.rb
@@ -1,0 +1,6 @@
+module DataRequestHelper
+  def only_branston_registry_email
+    @recipient_emails.empty? || (@recipient_emails.count == 1 && @recipient_emails.include?("BranstonRegistryRequests2@justice.gov.uk"))
+
+  end
+end
--- a/app/helpers/data_request_helper.rb
+++ b/app/helpers/data_request_helper.rb
@@ -1,6 +1,5 @@
 module DataRequestHelper
   def only_branston_registry_email
-    recipient_emails = instance_variable_get(:@recipient_emails)
-    recipient_emails.empty? || (recipient_emails.count == 1 && recipient_emails.include?(CommissioningDocumentTemplate::Probation::BRANSTON_ARCHIVES_EMAIL))
+    @recipient_emails.empty? || (@recipient_emails.count == 1 && @recipient_emails.include?(CommissioningDocumentTemplate::Probation::BRANSTON_ARCHIVES_EMAIL)) # rubocop:disable Rails::HelperInstanceVariable
   end
 end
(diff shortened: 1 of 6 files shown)


app/helpers/data_request_helper.rb
```diff
@@ -0,0 +1,6 @@
+module DataRequestHelper
+  def only_branston_registry_email
+    recipient_emails = instance_variable_get(:@recipient_emails)
```
vertism, line 3: You don't need to use `instance_variable_get` here as the instance variable will be available here, although rubocop would complain about it's use.

The reasons for Rubocop's complaint can be seen here - https://www.rubydoc.info/gems/rubocop-rails/RuboCop/Cop/Rails/HelperInstanceVariable.

There is usually a good reason for the rubocop warning but sometimes it might be right to ignore it and you can disable a rule by adding a comment on the offending line e.g:
```
some line of code with an @instance_var # rubocop:disable Rails::HelperInstanceVariable
```
or you can disable it for multiple lines:
```
# rubocop:disable Rails::HelperInstanceVariable
line with @instance_var
line with another @var
and another @var
# rubocop:enable Rails:HelperInstanceVariable
```

app/helpers/data_request_helper.rb
```diff
@@ -0,0 +1,6 @@
+module DataRequestHelper
```
vertism, line 1: You could avoid the use of a helper method by creating a new instance variable in the controller method to show whether @recipient_emails is empty before the Branston email gets added in.

You can use helper methods for many things that need to happen in a view, but the best use case is where it is only concerned with how something is displayed and where it needs to be used in several places. The less business logic contained in a helper the better.

spec/views/cases/data_requests/send_email_html_slim_spec.rb
```diff
@@ -38,7 +38,26 @@
...
       it "has required content" do
         expect(page.page_heading.heading.text).to eq "Are you sure you want to send the commissioning email?"
-        expect(page.page_banner.text).to include "The selected location does not have an email address."
+        expect(page.page_banner.text).to include "The selected location does not have an email address. Please update or select another."
+        expect(page.button_send_email.disabled?).to eq true
+        expect(page.link_cancel.text).to eq "Cancel"
+      end
+    end
+
+    context "with data request contact with only the branston probation records email given" do
```
vertism, line 47: Good to see a test! As we've already tested for other parts of the page being present, I think it would be find for this test to just test that the error banner is present and that the button is disabled.

app/helpers/data_request_helper.rb
vertism: Either way there should probably be a test for this method, or for the change to the controller if you do that.

vertism (dismissed):

app/helpers/data_request_helper.rb
lucas-shaw: Thanks for the tip, I did get the rubocop error so I used that variable. I will amend

vertism (commented):
LGTM

vertism (approved):

PR #2236, opened 2023-07-12
CDPT-814 Pass env vars to docker image and fix build tag

vertism:
## Description
Adds in ENV vars to be shown on /ping page.

## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -151,11 +151,13 @@ references:
     run:
       name: Create build tag
       command: |
-        current_branch=$(git branch | grep \* | cut -d ' ' -f2)
-        short_version=$(git rev-parse --short $current_branch)
-        build_tag=cmoj-${current_branch}-${short_version}
-        echo "export BUILD_TAG=${build_tag}" >> $BASH_ENV
+        prefix="ct"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -151,11 +151,13 @@ references:
     run:
       name: Create build tag
       command: |
-        current_branch=$(git branch | grep \* | cut -d ' ' -f2)
-        short_version=$(git rev-parse --short $current_branch)
-        build_tag=cmoj-${current_branch}-${short_version}
-        echo "export BUILD_TAG=${build_tag}" >> $BASH_ENV
+        prefix="ct"
```
wilson1000-MoJ, line 154: Small point; could `ct` be inserted into the string on line 156 instead of being assigned to a variable?

wilson1000-MoJ (commented):
Nice work @vertism

lucas-shaw (approved):
LGTM

.circleci/config.yml
vertism: It could, the reason I did it this way is that we will have similar code across a few applications and I wanted it to be obvious that the prefix would need to change if this code is re-used elsewhere.

PR #2239, opened 2023-07-21
Replace webdrivers

vertism:
## Description
A change to chrome (https://developer.chrome.com/blog/chrome-for-testing/) means that the latest version is not available via the webdrivers gem. 

It's use is no longer required with more recent versions of selenium-webdriver.

## Self-review checklist

* [ ] (3) Tests passing


njpreddy (approved):
Nice!

PR #2250, opened 2023-08-08
Add telephone number to dispatch template

lucas-shaw:
## Description
Changed the IMS001 word doc. 
Used a mergefield to display dynamic content for the telephone numbers (currently used for the letter_address, letter_date and body content). As the dispatch and acknowledgement letters use the same IMS001 template. This prevents the need to have a template doc for each of the two types.

This is affecting both the dispatch and acknowledgement letters as they share the same template.

Looking at the below checklist (8), no migrations were needed for this change to work. Although the template has technically changed.

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
(https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?modal=detail&selectedIssue=CDPT-860&search=860)


app/models/letter.rb
lucas-shaw, line 1: Not sure whether it's best to store the tel numbers as variables rather than hard-coding them in. Wasn't sure where would be the best place to store them. Happy to amend if needed.

vertism: You could certainly make an argument to extract the phone numbers, but as this is the only place they are being used I'm OK with them living here. 

One suggestion I would make is that you could move this method into the LetterTemplate model as the logic is based information about the template.

vertism: You could make the telephone numbers constants. If you also use the constants in the tests that would make them more robust.

lucas-shaw: Thanks! I get you however I'm not sure how to then implement the dynamic text. If I move the telephone_number method to letter_template.rb it throws an error of `undefined method `telephone_number'`

From what I've gathered the letters_controller.rb `show` method is where the message is sent to add the text to the word doc.

vertism: The letter template gets loaded in the show method so you need to call the telephone_number method on that object.

vertism: Also, I'm not sure it's worthwhile having a test to check the phone number value. 
You want to be testing the logic which I think you were doing before, you just need to move them to LetterTemplate and use the constant instead of the phone number.

app/models/letter_template.rb
```diff
@@ -15,4 +18,13 @@ def render(values, letter, field)
     template = ERB.new(send(field))
     template.result(binding)
   end
+
+  def telephone_number
+    case @letter_template.template_type
```
vertism, line 23:
```suggestion
    case template_type
```

vertism (approved):
Nice one, all looks good!

PR #2252, opened 2023-08-08
Fix InitialSeed migration

lucas-shaw:
## Description

When attempting to use 'rails db:reseed' on local db, an error occurred as below. Using join on the Rails root fixes the issue. 

== 20180101000000 InitialSeed: migrating ======================================
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:
wrong number of arguments (given 1, expected 0)

## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/db/data_migrations/20180101000000_initial_seed.rb
+++ b/db/data_migrations/20180101000000_initial_seed.rb
@@ -1,8 +1,8 @@
 class InitialSeed < ActiveRecord::DataMigration
   def up
-    require Rails.root.join("db/seeders/correspondence_type_seeder").to_s
-    require Rails.root.join("db/seeders/case_closure_metadata_seeder").to_s
-    require Rails.root.join("db/seeders/report_type_seeder").to_s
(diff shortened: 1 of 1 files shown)


db/data_migrations/20180101000000_initial_seed.rb
```diff
@@ -1,8 +1,8 @@
 class InitialSeed < ActiveRecord::DataMigration
   def up
-    require Rails.root("db/seeders/correspondence_type_seeder")
-    require Rails.root("db/seeders/case_closure_metadata_seeder")
-    require Rails.root("db/seeders/report_type_seeder")
+    require Rails.root.join("db/seeders/correspondence_type_seeder").to_s
```
vertism, line 3: Looks good, although I don't think the `to_s` is required.

wilson1000-MoJ: Yes, `.to_s` attempts to convert the response to a string. It won't be needed here.

wilson1000-MoJ (approved):
Nice :o)

PR #2261, opened 2023-08-18
London Disclosure response time inconsistent with case history time

lucas-shaw:
## Description
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?modal=detail&selectedIssue=CDPT-867

Case history time is correct with current localtime, however case response upload time shows +1 hour.

## Self-review checklist

* [ ] (3) Tests passing


lucas-shaw:
Initial thoughts: 
The method `in_time_zone` was adding +1 hour onto the current time. Seems as though rails automatically adjusts for the local time zone without needing this method, but not certain.

The tests `expect` (for the upload_group strings) were anticipating the +1 hour therefore needed to be changed to check the current time and not including the +1 hour for BST. 

Seems to affect only the response section within the cases view, the request section didn't have this issue.

app/models/case_attachment_upload_group.rb
```diff
@@ -38,6 +38,6 @@ def <=>(other)
 private
 
   def get_time(upload_group)
-    Time.find_zone("Etc/UTC").parse(upload_group).in_time_zone
```
vertism, line 41: Is `Time.find_zone` required, maybe just `Time.zone.parse` will work?

I think it would be good to add some tests for the `date_time` method which test times in in GMT and BST.

spec/models/case_attachment_upload_group_spec.rb
```diff
@@ -1,11 +1,9 @@
...
 
-# rubocop:disable RSpec/InstanceVariable, RSpec/BeforeAfterAll
+# rubocop:disable RSpec/InstanceVariable
 describe CaseAttachmentUploadGroup do
-  before(:all) do
-    DbHousekeeping.clean
-
-    @upload_group = "20170608101112"
+  before do
+    @upload_group = "20170608111112"
```
vertism, line 6: You should define these variables using let instead outside of the before block.
You can then also remove the rubocop rule disabling on lines 3 and 80.

spec/models/case_attachment_upload_group_spec.rb
```diff
@@ -34,8 +28,22 @@
   end
 
   describe "date_time" do
-    it "returns the formatted upload group time converted to BST" do
-      expect(upload_group.date_time).to eq "08 Jun 2017 11:11"
+    context "when timezone offset is BST" do
```
vertism, line 31: Looking at this now has made me realise there is no point testing GMT and BST because nothing is being done with the time zone anymore. 
So a test that checks the correct format is returned is all that is required.

spec/models/case_attachment_upload_group_spec.rb
```diff
@@ -34,8 +28,22 @@
...
+        expect(upload_group.date_time).to eq "08 Jun 2017 11:11"
+      end
+    end
+
+    context "when timezone offset is GMT" do
+      let(:timestamp) { "20170608101112" }
+      let(:upload_group) { described_class.new([timestamp, @responder.id], :responder, @kase, @kase.attachments) }
+
+      it "returns the formatted upload group time in GMT" do
+        expect(upload_group.date_time).to eq "08 Jun 2017 10:11"
```
vertism, line 45: This doesn't matter anymore because of my previous comment, but if the date was 8th June, you would expect a BST time to be returned.

app/models/case_attachment_upload_group.rb
```diff
@@ -38,6 +38,6 @@ def <=>(other)
 private
 
   def get_time(upload_group)
```
vertism, line 40: One final thought is that you could change this param name to `timestamp` or similar so it's clearer what is expected.

vertism (approved):
All looks good, thanks!

PR #2289, opened 2023-10-18
CDPT-727: CT: Display the contact addresses in alphabetical order

MeenaModhvadia:
## Description
Display the contact addresses in alphabetical order

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots
(https://github.com/ministryofjustice/correspondence_tool_staff/assets/92523605/de5b2ed4-d502-4ae3-a404-a8665d79c259)

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-727?atlOrigin=eyJpIjoiYWQ5ZTNhNWNlMzFmNDc4OGJiYjYwYWM0MDVkNjgxM2EiLCJwIjoiaiJ9


app/controllers/contacts_controller.rb
```diff
@@ -5,7 +5,7 @@ class ContactsController < ApplicationController
   before_action :set_contact_type, only: %i[update create]
 
   def index
-    @contacts = Contact.includes([:contact_type]).all.decorate
+    @contacts = Contact.order(name: :asc).includes([:contact_type]).all.decorate
```
vertism, line 8: Looks good! I think the `all` here is redundant (and already was) so you could remove that.
Also, while there is no rule around this, people tend to put order clauses at the end of a query like this. It doesn't make any difference to what it does, it's just a style thing. 
i.e. `@contacts = Contact.includes([:contact_type]).order(name: :asc).decorate`

I can see that there isn't a spec for this controller for some reason, but it would be good to at least add one with a test for this change.

MeenaModhvadia: Thank you - I will amend the change I have made and also add a spec and test

spec/controllers/contacts_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    before do
+      sign_in user
+    end
+
+    it "returns an alphabetically sorted list of contacts" do
+      # Assuming the controller action assigns @contacts
+      get :index
+
+      # Extract the contacts from the controller's instance variable
+      contacts = assigns(:contacts.name)
```
vertism, line 20: I'm not sure what `.name` is here, or indeed how this is working with it!

vertism: I've looked into this and it is a new method that can be called on symbols in Ruby 3 - https://www.bigbinary.com/blog/ruby-3-adds-symbol-name - which basically returns the symbol as a string.

It's not needed here anyway.

MeenaModhvadia: Thank you for the above.  I have removed it from the code base.

vertism (approved):
Nice work, LGTM.

PR #2307, opened 2023-11-27
Create 'rejected' case type

lucas-shaw:
## Description
Create a new Offender SAR type of 'rejected', which is defined in `current_state`

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?selectedIssue=CDPT-1033


app/models/case/sar/offender.rb
```diff
@@ -72,7 +72,8 @@ def searchable_fields_and_ranks
                  number_exempt_pages: :integer,
                  is_partial_case: :boolean,
                  partial_case_letter_sent_dated: :date,
-                 further_actions_required: :string
+                 further_actions_required: :string,
+                 is_rejected: :boolean
```
lucas-shaw, line 76: Thoughts are to add this attribute as a boolean. Defines whether a case is rejected.

app/controllers/cases/offender_sar_controller.rb
```diff
@@ -29,6 +29,7 @@ def initialize
 
     def new
       permitted_correspondence_types
+      @is_rejected = !!params.fetch("is_rejected", false) # rubocop:disable Style/DoubleNegation
```
vertism, line 32: To avoid the rubocop complaint, it might be better to use `key?` instead:
```suggestion
      @is_rejected = params.key?("is_rejected")
```

vertism: A very minor one, but as it's user facing, maybe the URL param should be `rejected` instead of `is_rejected`?

vertism (approved):
Looks good!

PR #2309, opened 2023-12-05
CDPT-1034 CT Rejections: Add Rejected to title

MeenaModhvadia:
## Self-review checklist

* [ ] (3) Tests passing


app/views/cases/_new.html.slim
```diff
@@ -2,7 +2,7 @@
   = t('.heading')
 
 - content_for :sub_heading
-  = t4c(kase, '', 'sub_heading', case_type: kase.type_abbreviation)
+  = get_offender_sar_heading(@offender_sar_type, @case)
```
vertism, line 5: This needs to be more generic so it works for any case type. At the moment it's going to break the new case page for FOIs and SARs.

To start you should create a method that takes `kase` as a param and then have it output the exact same thing it does now.

Then you need to work out how to determine when the new case is a rejected offender sar based on information in that kase variable.

app/views/cases/new.html.slim
```diff
@@ -2,7 +2,7 @@
   = t('.heading')
 
 - content_for :sub_heading
-  = t4c(@case, '', 'sub_heading', case_type: @case.pretty_type)
+  - get_offender_sar_heading(@offender_sar_type, @case)
```
vertism, line 5: Does this need to change?

app/views/cases/_new.html.slim
vertism: It is also possible that you can made changes to the `TranslateForCase` class (which is what the t4c method eventually calls to account for a rejected offender sar there.

config/locales/filters.en.yml
```diff
@@ -89,14 +89,15 @@ en:
...
-      not-high-profile: Not high profile 
+      high-profile: High profile
+      not-high-profile: Not high profile
     filter_partial_case_flag:
       partial-case: Is partial case
       not-partial-case: Is not partial case
     filter_status:
       open: 'Open'
       closed: 'Closed'
+      rejected: 'Rejected'
```
vertism, line 100: I'm not sure this change is required for this ticket.

config/locales/page_titles.en.yml
```diff
@@ -19,6 +19,7 @@ en:
       ico: Creating an ICO case - Track-a-query
       offender_sar:
         subject_details_step: Subject details - Creating an Offender SAR case - Track-a-query
+        rejected_details_step:  Subject details - Creating a Rejected Offender SAR case - Track-a-query
```
vertism, line 22: I'm not sure this addition makes sense.

config/locales/state.en.yml
```diff
@@ -15,6 +15,7 @@ en:
       ready_to_dispatch: Ready to dispatch
       vetting_in_progress: Vetting in progress
       waiting_for_data: Waiting for data
+      rejected: Rejected
```
vertism, line 18: This may relate to a different ticket

config/routes.rb
```diff
@@ -163,6 +163,7 @@
       get "deleted" => "filters#deleted"
       get "incoming" => "filters#incoming"
       get "/" => "filters#show"
+      get "rejected" => "filters#rejected"
```
vertism, line 166: I don't think this is necessary for this ticket

config/state_machine/configs/00_moj/20_sar/20_sar_standard_workflow.yml
```diff
@@ -109,6 +109,10 @@
                   if: Workflows::Predicates#responder_is_member_of_assigned_team?
                   after_transition: Workflows::Hooks#reassign_user_email
 
+              rejected:
```
vertism, line 112: This doesn't seem relevant for this ticket.
This also goes for the other changes to the state machine config.

app/views/cases/_new.html.slim
MeenaModhvadia: I am thinking of adding a flag to set on the case if it's a rejected offender sar then the t4c can check and output the rejected heading if its an offender sar & rejected.
What do you think?

vertism: That seems like a reasonable idea.

MeenaModhvadia:
This pr is no longer needed as created new branch off cdpt-1033 to allow for those changes as my ticket follows on from there

PR #2310, opened 2023-12-21
CDPT 1034 ct rejections subject page headings

MeenaModhvadia:
## Self-review checklist

* [ ] (3) Tests passing


app/helpers/views_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ViewsHelper
+  def get_sar_heading(kase)
+    if @case.current_state == "rejected" # rubocop:disable Rails/HelperInstanceVariable
```
vertism, line 3: You are passing `@case` into the method as `kase` so you can use that variable. That will take care of the warning.

vertism: You should probably check that the case type is an Offender SAR as well, just in case any of the other case types ever have a "rejected" state in the future.

app/views/cases/new.html.slim
```diff
@@ -2,7 +2,7 @@
   = t('.heading')
 
 - content_for :sub_heading
-  = t4c(@case, '', 'sub_heading', case_type: @case.pretty_type)
+  = get_sar_heading(@case)
```
vertism, line 5: You probably want to name this something more generic, as it is getting the heading for all case types.

MeenaModhvadia:
Closed  need to merge into branston not main

PR #2311, opened 2023-12-22
CDPT-1034 CT Rejections subject page headings

MeenaModhvadia:
## Description
As a Branston user
I want to  capture the correct information needed to create a rejected record
So that I can have all the relevant identifiers and correct information required to make it a valid SAR

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-1034?


lucas-shaw (approved):
LGTM

app/helpers/views_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ViewsHelper
+  def get_headings(kase, correspondence_type)
+    if kase.current_state == "rejected" && correspondence_type.abbreviation == "OFFENDER_SAR"
+      content_tag(:span, t("cases.new.offender_sar.rejected"), class: "page-heading--secondary")
```
vertism, line 4: Is this the same as doing `t4c(kase, "", "rejected", case_type: kase.pretty_type)`?

app/helpers/views_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ViewsHelper
+  def get_headings(kase, correspondence_type)
+    if kase.current_state == "rejected" && correspondence_type.abbreviation == "OFFENDER_SAR"
```
vertism, line 3:
```suggestion
    if kase.current_state == "rejected" && correspondence_type.abbreviation == CorrespondenceType.offender_sar.abbreviation
```

app/views/cases/offender_sar/_subject_details_step.html.slim
```diff
@@ -38,5 +38,5 @@
 
   = f.submit 'Continue', class: 'button'
 
-
+  = link_to "Cancel", open_filter_path, class: 'acts-like-button button-right-spacing'
```
vertism, line 41: This needs removing

app/helpers/views_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ViewsHelper
+  def get_headings(kase, correspondence_type)
+    if kase.current_state == "rejected" && correspondence_type.abbreviation == CorrespondenceType.offender_sar.abbreviation
+      t4c(kase, "", t("cases.new.offender_sar.rejected"), case_type: kase.pretty_type)
```
vertism, line 4: I think this will work:
```suggestion
      t4c(kase, "", t("rejected"), case_type: kase.pretty_type)
```

app/helpers/views_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ViewsHelper
+  def get_headings(kase, correspondence_type)
```
vertism, line 2: This method needs a unit test

spec/helpers/views_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rails_helper"
+  describe ViewsHelper, type: :helper do
+    include ViewsHelper
```
vertism, line 3: This shouldn't be needed. RSpec knows which file to test.

spec/helpers/views_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+
+    describe "Creating an offender sar" do
+      context "adding a rejected offender sar case" do
+        describe Case, type: :model do
+
+          it "creates an offender object" do
+            # Use the factory to create an instance of Case::Sar::Offender
+            kase = FactoryBot.build(Case::Sar::Offender)
```
vertism, line 12: Using `FactoryBot.build` is fine, but you can just use `build` instead as a shortcut.

There are many different factories defined in various files under spec/factories. These will set the attributes to use. It would probably be worth reading how FactoryBot works starting from https://github.com/thoughtbot/factory_bot/blob/main/GETTING_STARTED.md#defining-factories

spec/helpers/views_helper_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+            headings = get_headings(kase, "rejections", OFFENDER_SAR")
```
vertism, line 14: There is an additional `"` in this line

app/helpers/views_helper.rb
```diff
@@ -0,0 +1,9 @@
...
+    if kase.current_state == "rejected" && correspondence_type.abbreviation == CorrespondenceType.offender_sar.abbreviation
+      t4c(kase, "", "cases.new.offender_sar.rejected", case_type: kase.type)
+    else
+      t4c(kase, "", "cases.new.offender_sar.sub_heading", case_type: kase.type)
```
vertism, line 6: If you look at the original method that was being called:
```
t4c(@case, '', 'sub_heading', case_type: @case.pretty_type)
```
The argument to case_type is `@case.pretty_type`, so you need to use the same thing here in both case. `pretty_type` is a decorator method so you need to call `kase.decorate.pretty_type` to access it.

When t4c is called from within a view, the `t` method will look for the translation in the part of the translation file that relates to that page. The 2nd argument of the t4c method allows you to specify where that is e.g. "cases.new".

spec/helpers/views_helper_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+RSpec.configure do |config|
+  config.include AbstractController::Translation
+end
+
+describe ViewsHelper, type: :helper do
+  include CasesHelper
+
+  describe "#get_headings" do
+    it "text for rejected case type heading" do
```
vertism, line 11: The test name should read like a sentence. eg. it "returns text for rejected case type heading"

spec/helpers/views_helper_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      view_title = get_headings(test_kase, test_correspondence_type)
+
+      expect(view_title).to eq("Create Rejected Offender SAR case")
+    end
+
+    it "#text for a sub_heading" do
+      test_kase = build(:offender_sar_case)
+
+      test_correspondence_type = build(:offender_sar_correspondence_type)
```
vertism, line 24: This is the same in both tests, so would be better to extract to a let.

app/views/cases/offender_sar_select_type.html.slim
```diff
@@ -12,19 +12,17 @@
...
       = p link_to t('.valid_offender_sar'), new_case_sar_offender_path
   <br/>
 
   label.form-label
-    = t('.helpers.label.correspondence_types.status.rejected')
+    = t('helpers.label.correspondence_types.status.rejected')
   li
-      = p link_to t('.rejected_offender_sar'),
-      new_case_sar_offender_path(params: {rejected: true})
-  <br/>
```
vertism, line 25: I don't think this <br> should be removed.

app/views/cases/new.html.slim
```diff
@@ -2,7 +2,7 @@
   = t('.heading')
 
 - content_for :sub_heading
-  = t4c(@case, '', 'sub_heading', case_type: @case.pretty_type)
+  = get_headings(@case, @correspondence_type, '')
```
vertism, line 5: You don't need to pass in the last argument now that you've added a default value to the method.

app/helpers/views_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ViewsHelper
+  def get_headings(kase, correspondence_type, key_path = "")
+    if kase.current_state == "rejected" && correspondence_type.abbreviation == CorrespondenceType.offender_sar.abbreviation
+      t4c(kase, key_path, "offender_sar.rejected", case_type: kase.decorate.pretty_type)
```
vertism, line 4: I don't think you need to pass in case_type as an argument here.

app/helpers/views_helper.rb
```diff
@@ -0,0 +1,9 @@
+module ViewsHelper
+  def get_headings(kase, correspondence_type, key_path = "")
```
vertism, line 2: Should this be singular? You're only getting 1 heading.

vertism (approved):
LGTM

PR #2312, opened 2023-12-27
Add 'rejected' filter option

lucas-shaw:
## Description
Add the new 'rejected' case state filter option

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-1084

### DIFF
--- a/app/services/case_filter/open_case_status_filter.rb
+++ b/app/services/case_filter/open_case_status_filter.rb
@@ -27,14 +27,22 @@ module CaseFilter
         end
       end
 
-      # To ensure 'rejected' is at the end of the filter list, .delete returns the
-      # value of the given key. It is then immediately assigned back.
-      state_choices["rejected"] = state_choices.delete("rejected")
+      offender_sar_move_state_filter_option_to_end(state_choices, "rejected")
 
       { filter_open_case_status: state_choices }
     end
 
   private
+
+    # For users dealing with Offender SARs
+    # To ensure 'rejected' is at the end of the filter list, .delete returns the
+    # value of the given key. It is then immediately assigned back.
+    def offender_sar_move_state_filter_option_to_end(filters, filter_key)
+      if @user.permitted_correspondence_types.any? { |h| h[:abbreviation] == "OFFENDER_SAR" }
+        filters[filter_key] = filters.delete(filter_key)
+        filters
+      end
+    end
 
     def get_tranlation_of_state_for_filter(sub_class, state)
       I18n.t("filters.filter_open_case_status.#{state}", default: nil) || TranslateForCase.translate(sub_class, "state", state)
--- a/app/services/case_filter/open_case_status_filter.rb
+++ b/app/services/case_filter/open_case_status_filter.rb
@@ -26,6 +26,11 @@ module CaseFilter
           end
         end
       end
+
+      # To ensure 'rejected' is at the end of the hash, .delete returns the
+      # value of the given key. It is then immediately assigned back.
+      state_choices["rejected"] = state_choices.delete("rejected")
+
       { filter_open_case_status: state_choices }
     end
 
--- a/app/services/case_filter/open_case_status_filter.rb
+++ b/app/services/case_filter/open_case_status_filter.rb
@@ -27,19 +27,14 @@ module CaseFilter
         end
       end
 
-      move_state_filter_option_to_end(state_choices, "rejected")
+      # To ensure 'rejected' is at the end of the filter list, .delete returns the
+      # value of the given key. It is then immediately assigned back.
+      state_choices["rejected"] = state_choices.delete("rejected")
 
       { filter_open_case_status: state_choices }
     end
 
   private
-
-    # To ensure 'rejected' is at the end of the filter list, .delete returns the
-    # value of the given key. It is then immediately assigned back.
-    def move_state_filter_option_to_end(filters, filter_key)
-      filters[filter_key] = filters.delete(filter_key)
-      filters
-    end
 
     def get_tranlation_of_state_for_filter(sub_class, state)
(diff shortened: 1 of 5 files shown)


spec/services/case_filter/open_case_status_filter_spec.rb
```diff
@@ -22,6 +22,7 @@
       std_responded_foi
       trig_responded_foi
       full_responded_foi
+      std_rejected_sar
```
lucas-shaw, line 25: Not sure whether this is the correct naming convention. Standard workflow, rejected offender sar case. Followed the comments in `spec/support/standard_setup.rb`

app/services/case_filter/open_case_status_filter.rb
```diff
@@ -26,11 +26,21 @@ def available_choices
           end
         end
       end
+
+      # To ensure 'rejected' is at the end of the filter list, .delete returns the
```
vertism, line 30: The explanation for what is being done should probably be with the change_case_filter_order method. Its name could maybe be changed to make it clear that the element is being moved to the end of the hash?

app/services/case_filter/open_case_status_filter.rb
```diff
@@ -26,11 +26,21 @@ def available_choices
           end
         end
       end
+
+      move_state_filter_option_to_end(state_choices, "rejected")
```
vertism, line 30: Just thinking about this again. Do the options need to be in a particular order? If so, would it be the order where they are originally defined, in which case can you not change that order instead of doing it here?

lucas-shaw: Hi the ticket only asks for 'rejected' to be at the end. The order is defined within the workflow file 'config/state_machine/configs/00_moj/40_offender_sar/10_offender_sar_preamble.yml', with `rejected` near the bottom of `permitted states` (`closed` is not shown). 

I did consider your suggestion however it seems that `{ filter_open_case_status: state_choices }` returns most of the status options. Then when `available_choices` is looped through within the view `app/views/cases/search_filters/_filter_multiple_choices.slim`, more options are added which seem to relate to SAR complaints only, so `rejected' is no longer at the bottom. `filter.available_choices[filter_attribute]` within the view adds on the extra filter options. I'm not sure how to resolve this in a different way.

lucas-shaw: I think `describe ".process_params!" do` within `open_case_status_filter_spec.rb` tests for list order, however I'm not sure how to do a test for the full view to ensure that 'rejected' appears after the SAR complaint status filters as well.

app/services/case_filter/open_case_status_filter.rb
```diff
@@ -26,11 +26,24 @@ def available_choices
...
       { filter_open_case_status: state_choices }
     end
 
   private
 
+    # For users dealing with Offender SARs
+    # To ensure 'rejected' is at the end of the filter list, .delete returns the
+    # value of the given key. It is then immediately assigned back.
+    def offender_sar_move_state_filter_option_to_end(filters, filter_key)
+      if @user.permitted_correspondence_types.any? { |h| h[:abbreviation] == "OFFENDER_SAR" }
```
lucas-shaw, line 41: Reused the method as it now includes a conditional. Without the conditional non branston users had an unwanted filter option named `filter open case status rejected`

spec/services/case_filter/open_case_status_filter_spec.rb
```diff
@@ -198,6 +213,40 @@
...
+
+    context "when logged in as a London Disclosure user" do
+      let(:user) { find_or_create :disclosure_specialist_bmt }
+      let(:open_case_status_filter) { described_class.new search_query, user, Case::Base }
+      let(:search_query) do
+        create :search_query
+      end
+
+      it "has the rejected key" do
+        expect(open_case_status_filter.available_choices[:filter_open_case_status]).not_to have_key("rejected")
```
lucas-shaw, line 241: Do we need to check that all non Offender users do not have the `rejected` key?

spec/services/case_filter/open_case_status_filter_spec.rb
```diff
@@ -198,6 +213,40 @@
...
+      end
+
+      it "returns the filter list with rejected as nil" do
```
lucas-shaw, line 244: Worth removing this now?

app/services/case_filter/open_case_status_filter.rb
lucas-shaw: I did that just in case it needed to be a new or different key in the future, as is easier to change. Happy to remove if needs be?

vertism: That would be easy enough to change in the future anyway. Probably better to design this for how you need it to work now.

spec/services/case_filter/open_case_status_filter_spec.rb
```diff
@@ -198,6 +213,40 @@
...
+    end
+
+    context "when logged in as a London Disclosure user" do
+      let(:user) { find_or_create :disclosure_specialist_bmt }
+      let(:open_case_status_filter) { described_class.new search_query, user, Case::Base }
+      let(:search_query) do
+        create :search_query
+      end
+
+      it "has the rejected key" do
```
vertism, line 240: This test name needs changing

spec/services/case_filter/open_case_status_filter_spec.rb
vertism: I don't really understand the purpose of it, so probably :)

vertism: I think you just need a test to check that there is no rejected key present

vertism (approved):
LGTM

PR #2318, opened 2024-01-11
CDPT-1075 CT Rejections Who is making the request page

MeenaModhvadia:
## Description
As a Branston user I want to capture the correct information needed to create a rejected record.  So that I can have all the relevant identifiers and correct information required to make it a valid SAR

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-1075?


db/structure.sql
```diff
@@ -9,148 +9,6 @@ SET xmloption = content;
 SET client_min_messages = warning;
 SET row_security = off;
 
---
```
vertism, line 12: None of this should be deleted

app/models/case/sar/offender.rb
```diff
@@ -239,6 +241,18 @@ def validate_third_party_address
     errors[:third_party_relationship].any?
   end
 
+  def validate_third_party_email
+    if (third_party || recipient == "third_party_recipient")
+      return if third_party_email.blank?
+
+      third_party_email.split =~ /\A([^@,]+)@([^@,]+)\z/ # regex disallows commas and additional @s
```
vertism, line 248: I don't think you need to use split here.

app/models/case/sar/offender.rb
```diff
@@ -239,6 +241,18 @@ def validate_third_party_address
...
+        errors.add(
```
vertism, line 249: This will always add an error if an email value exists

app/models/case/sar/offender.rb
MeenaModhvadia: Removed split and refactored code

app/models/case/sar/offender.rb
```diff
@@ -239,6 +241,18 @@ def validate_third_party_address
     errors[:third_party_relationship].any?
   end
 
+  def validate_third_party_email
+    if third_party
+      return if third_party_email.blank?
+      unless third_party_email =~ /\A([^@,]+)@([^@,]+)\z/ # regex disallows commas and additional @s
```
vertism, line 247: There's a email validation no case which might be better to copy - https://github.com/ministryofjustice/correspondence_tool_staff/blob/main/app/models/case/base.rb#L967

config/locales/en.yml
```diff
@@ -281,6 +281,8 @@ en:
               blank: cannot be blank
             third_party_address:
               blank: cannot be blank
+            third_party_email:
+              invalid: enter an email address in the correct format, like name@example .com
```
vertism, line 285: There's an extra space that can be removed
```suggestion
              invalid: enter an email address in the correct format, like name@example.com
```

db/structure.sql
```diff
@@ -160,10 +160,10 @@ SET default_table_access_method = heap;
 --
 
 CREATE TABLE public.ar_internal_metadata (
```
vertism, line 162: All these extra spaces should be removed. You should be able to get the original file from main by running
```
git checkout origin/main db/structure.sql
```

spec/models/case/sar/offender_spec.rb
```diff
@@ -328,9 +328,36 @@
...
+
+      it "does not validate email address when there is no email address entered" do
+        kase = build_stubbed :offender_sar_case, third_party: true, third_party_relationship: "Solicitor",
+                                                 third_party_email: "", third_party_company_name: "ABC LLP",
+                                                 recipient: "third_party_recipient"
+
+        expect(kase).to be_valid
+      end
+
+      it "validates email address when there is an email address entered" do
```
vertism, line 356: This test has the same name as a previous one. Also the entered email address is invalid, but is passing.

app/models/case/sar/offender.rb
MeenaModhvadia: This regex breaks the third_party_email validation and browser hangs

vertism: In this commit - https://github.com/ministryofjustice/correspondence_tool_staff/pull/2318/commits/246099428db673eac0996a55b4eca28671fd0155 - you change the attribute to `email` instead of `third_party_email`

app/views/cases/offender_sar/_requester_details_step.html.slim
```diff
@@ -18,8 +18,10 @@
         = f.text_field :third_party_company_name
         = f.text_area  :postal_address, {rows: 4, class: 'address_input'}
         = render partial: 'contacts/open_search_dialogue_button', locals: { search_filters: "solicitor", bespoke_lookup_text: 'find_solicitor_address' }
+        = f.text_field :third_party_email, label_options: { value: 'Email address (optional)' }
```
vertism, line 21: This label can be set in en.yml

spec/controllers/cases/offender_sar_controller_spec.rb
```diff
@@ -415,6 +415,7 @@ def third_party_validations_found(errors)
     expect(errors[:third_party_company_name]).to eq ["cannot be blank if representative name not given"]
     expect(errors[:third_party_relationship]).to eq ["cannot be blank"]
     expect(errors[:postal_address]).to eq ["cannot be blank"]
+    expect(errors[:third_party_email.format).to eq []
```
vertism, line 418: This is causing the rubocop error, but you can probably remove it completely.

config/locales/en.yml
```diff
@@ -54,6 +54,7 @@ en:
         third_party_company_name : "Company name"
         third_party_name : "Full name of the representative"
         third_party_relationship: "Relationship to the subject"
+        third_party_email: "Email address (optional)"
```
vertism, line 57: The value here will also be used in places like the error message (where you don't want the "optional" part to appear. There is also a translation for labels where this should be added.

vertism (approved):

PR #2320, opened 2024-01-31
CDPT-1036: CT Rejection status update case details to include rejection reason

MeenaModhvadia:
## Description
As a Business Support Team member
I want to capture the correct information needed to create a rejected record
So that I can have all the relevant identifiers and correct information required to make it a valid SAR

As a Business Support Team member
I want the rejection case to have a unique DPA number when created
So that I can  provide requestor with a reference, and it doesn't interfere with valid SAR start date

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-1036


app/models/case/sar/offender.rb
```diff
@@ -408,6 +418,17 @@ def user_dealing_with_vetting
     user_for_vetting
   end
 
+  def next_seq_for_rejected_case_number(received_date)
```
vertism, line 421: This method is not getting called anywhere.

I'm not sure what is being done below is necessary. It should be possible to use the existing method that creates a number in the base class and prefix a letter.

app/models/case/base.rb
```diff
@@ -1030,7 +1030,11 @@ def set_managing_team
   end
 
   def set_number
-    self.number = next_number
+    self.number = if current_state == "rejected"
```
vertism, line 1033: The rejected state only affects Offender SARs, so it's probably best to only make this change in that class. You can override this method as it's a subclass.

app/models/case/sar/offender.rb
```diff
@@ -181,6 +181,7 @@ def searchable_fields_and_ranks
 
   before_validation :ensure_third_party_states_consistent
   before_validation :reassign_gov_uk_dates
+  before_create :set_number
```
vertism, line 184: This callback already happens on the base class

spec/models/case/sar/offender_spec.rb
```diff
@@ -72,6 +72,15 @@
     end
   end
 
+  describe "#set_number" do
+    it "assigns a case number to rejected offender sar" do
```
vertism, line 76: This test does not test the set_number method. You can verify this by making a change to the method - e.g. change the prefix letter from "R" to something else - and seeing that this test still passes.

app/models/case/sar/offender.rb
```diff
@@ -475,5 +475,14 @@ def ensure_third_party_states_consistent
...
+
+  def set_number
+    self.number = if current_state == "rejected"
+                    "R#{next_number}"
+                  else
+                    next_number
+                  end
+  end
+
+  # rubocop:enable Metrics/ClassLength
```
vertism, line 487: I think you can safely delete this rubocop comment

app/controllers/concerns/offender_sar_cases_params.rb
```diff
@@ -38,6 +38,7 @@ def create_offender_sar_params
       :request_method,
       :requester_reference,
       :current_state,
+      :rejected,
```
vertism, line 41: Where is this param used?

MeenaModhvadia: This rejected param is being used in _case_details.html.slim

vertism: The params in this file are used to allow updating of attributes via a form. It's a security measure called Strong Parameters - https://guides.rubyonrails.org/action_controller_overview.html#strong-parameters.

In _case_details.html.slim you are checking the value of an attribute which is an unrelated action.

spec/models/case/sar/offender_spec.rb
```diff
@@ -72,6 +72,17 @@
     end
   end
 
+  describe "#set_number" do
+    let(:case_rejected)  { create(:offender_sar_case, :rejected, received_date: Date.parse("08/02/2024")) }
+
+    it "is composed of prefix R, received date and an incremented suffix" do
```
vertism, line 78: You're only testing the prefix, so the test name needs changing and you also don't need to set the received date or use timecop.

Also, there should be a test for when the case is not rejected.

app/views/cases/offender_sar/_reason_rejected_details.html.slim
```diff
@@ -0,0 +1,4 @@
+tr.rejected_reasons
+  th = t('common.case.rejected_reason') if case_details.rejected_reasons.present?
```
vertism, line 2: Is this check necessary? Can a case be rejected without a reason?

spec/models/case/sar/offender_spec.rb
```diff
@@ -72,6 +72,21 @@
     end
   end
 
+  describe "#set_rejected_prefix" do
+    let(:case_rejected) { create(:offender_sar_case, :rejected) }
+    let(:case) { create(:offender_sar_case) }
+
+    it "is composed of prefix R, received date and an incremented suffix" do
```
vertism, line 79: The test name needs updating

spec/models/case/sar/offender_spec.rb
```diff
@@ -72,6 +72,21 @@
...
+      kase = create :offender_sar_case, :rejected, rejected_reasons: %w[cctv_bwcv]
```
vertism, line 80: You should add the rejected_reasons to the factory instead of here.

spec/views/cases/offender_sar/_case_details_html_slim_spec.rb
```diff
@@ -60,6 +60,23 @@ def login_as(user)
       expect(partial.third_party_name.data.text).to eq "Rick Westor"
       expect(partial.requester_reference.data.text).to eq "FOOG1234"
       expect(partial.third_party_company_name.data.text).to eq "Foogle and Sons Solicitors at Law"
+      expect(partial.third_party_email.data.text).to eq "foogle@solicitors.com"
+    end
+
+    it "displays rejected details if present" do
+      rejected_case = (create :offender_sar_case, :rejected, rejected_reasons: "ID required, Court data request, further identification, police data").decorate
```
vertism, line 67: This is not a valid rejected_reasons which needs to be an array.

spec/views/cases/offender_sar/_case_details_html_slim_spec.rb
```diff
@@ -60,6 +60,23 @@ def login_as(user)
...
+        case_details: rejected_case,
+        link_type: nil,
+        allow_editing: true,
+      }
+
+      partial = offender_sar_case_details_section(rendered)
+
+      # expect(partial.has_selector?("rejected")).to eq true
+      expect(partial.has_selector?(".rejected_reasons")).to eq(true)
+      expect(partial.rejected_reasons.data.text).to eq "ID required, Court data request, further identification, police data"
```
vertism, line 79: You've called it `rejected_reason` in the site_prism file.

MeenaModhvadia: site_prism file?

spec/models/case/sar/offender_spec.rb
```diff
@@ -72,6 +72,19 @@
     end
   end
 
+  describe "#set_rejected_prefix" do
```
vertism, line 75: This describe needs updating

spec/views/cases/offender_sar/_case_details_html_slim_spec.rb
```diff
@@ -60,6 +60,20 @@ def login_as(user)
...
+      rejected_case = (create :offender_sar_case, :rejected)
+      assign(:case, rejected_case)
+      render partial: "cases/offender_sar/case_details", locals: {
+        case_details: rejected_case,
+        link_type: nil,
+        allow_editing: true,
+      }
+
+      partial = offender_sar_case_details_section(rendered)
+      expect(partial.rejected_reason.data.text).to eq "[further identification]"
```
vertism, line 76: This will need changing. An array is not what you are expecting

app/helpers/cases_helper.rb
```diff
@@ -236,6 +236,11 @@ def action_button_for(event)
               case_cover_page_path(@case),
               id: "action--preview-cover-page",
               class: "button-secondary"
+    when :information_received
+      link_to "Information received",
+              "",
```
lucas-shaw, line 241: Not added the link yet as its to be done in [cdpt-1125-information-received-page](url)

app/decorators/case/sar/offender_decorator.rb
```diff
@@ -16,4 +16,10 @@ def request_methods_sorted
   def request_methods_for_display
     request_methods_sorted - %w[unknown]
   end
+
+  def rejected_reasons_descriptions
+    rejected_reasons.map { |rejected_reasons|
+      Case::SAR::Offender::REJECTED_REASONS[rejected_reasons]
+    }.compact.join("<br>")
```
vertism, line 23: a multiline map should use "do end". I've also updated the variable to be singular in the suggestion
```suggestion
    rejected_reasons.map do |reason|
      Case::SAR::Offender::REJECTED_REASONS[reason]
    end.compact.join("<br>")
```

app/decorators/case_transition_decorator.rb
```diff
@@ -35,6 +35,14 @@ def event_name
     I18n.t(specific_key, default: I18n.t(default_key))
   end
 
+  def rejected_case_creation_event
+    # This method prevents the case history changing from "rejected case created" to "case created"
+    # when a Case state is changed from "rejected" to another state
+    if object.case.has_attribute?(:case_originally_rejected) && (object.event == "create" && object.case.case_originally_rejected == true)
```
vertism, line 41: You don't need to check a boolean value like that, it will evaluate as true or false directly.
Parentheses can be useful for comprehension, but probably not needed in this case.
```suggestion
    if object.case.has_attribute?(:case_originally_rejected) && object.event == "create" && object.case.case_originally_rejected
```

spec/decorators/case/sar/offender_decorator_spec.rb
```diff
@@ -75,6 +75,18 @@
...
   end
 
+  describe "#rejected_reasons_descriptions" do
+    let(:offender_sar_case) do
+      build_stubbed(
+        :offender_sar_case, :rejected
+      ).decorate
+    end
+
+    it "returns the REJECTED_REASONS hash value" do
```
vertism, line 85: It would be good to test for when there are multiple values.

app/views/cases/offender_sar/_reason_rejected_details.html.slim
```diff
@@ -0,0 +1,4 @@
+tr.rejected-reasons
+  th = t('common.case.rejected_reason')
+  td = raw case_details.rejected_reasons_descriptions
```
vertism, line 3: Should we be showing the other reason, if there is one?

app/decorators/case_transition_decorator.rb
```diff
@@ -35,6 +35,14 @@ def event_name
     I18n.t(specific_key, default: I18n.t(default_key))
   end
 
+  def rejected_case_creation_event
```
vertism, line 38: Is there a test for this?

app/decorators/case/sar/offender_decorator.rb
```diff
@@ -16,4 +16,10 @@ def request_methods_sorted
   def request_methods_for_display
     request_methods_sorted - %w[unknown]
   end
+
+  def rejected_reasons_descriptions
+    rejected_reasons.map do |reason| # rubocop:disable Style/BlockDelimiters
```
vertism, line 21: If rubocop is complaining about this, then go with what it wants you to do.

vertism (approved):
LGTM

PR #2321, opened 2024-02-01
Add rejected flag to case database

lucas-shaw:
## Description
Add a flag to the case model which confirms whether a case was ever rejected. To be used for reporting purposes. 

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?selectedIssue=CDPT-1207

### DIFF
--- a/app/models/case/sar/offender.rb
+++ b/app/models/case/sar/offender.rb
@@ -95,6 +95,7 @@ class Case::SAR::Offender < Case::Base
                  is_partial_case: :boolean,
                  partial_case_letter_sent_dated: :date,
                  further_actions_required: :string,
+                 case_originally_rejected: :boolean,
                  other_rejected_reason: :string,
                  rejected_reasons: [:string, { array: true, default: [] }]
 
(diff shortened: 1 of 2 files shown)


app/models/case/sar/offender.rb
```diff
@@ -95,6 +95,7 @@ def searchable_fields_and_ranks
                  is_partial_case: :boolean,
                  partial_case_letter_sent_dated: :date,
                  further_actions_required: :string,
+                 case_originally_rejected: :boolean,
```
lucas-shaw, line 98: I didn't add a ? to the attr name as the other booleans don't seem to.

app/models/case/sar/offender.rb
```diff
@@ -459,6 +461,10 @@ def use_subject_as_requester
     self.name = subject_full_name
   end
 
+  def set_case_originally_rejected
+    self.case_originally_rejected = true
```
lucas-shaw, line 465: I assumed it would be ok to not give a value as false if its not a rejected case. So either nil or true.

spec/models/case/sar/offender_spec.rb
```diff
@@ -290,6 +290,13 @@
     end
   end
 
+  describe "when creating a rejected case" do
```
lucas-shaw, line 293: Thought about adding a test to check if it returns nil for a non rejected case, but not sure whether thats necessary.

vertism (approved):
LGTM

PR #2324, opened 2024-02-05
Fix offender sar creation page flow 

lucas-shaw:
## Description
For valid and rejected offender sar creation. When swapping midway through the creation steps from valid to rejected (e.g if the user changed their mind on the case type to create), the state was not changing to reflect this change. 

Therefore, starting a valid case after a partially made rejected case would incorrectly render the reason-rejected page and give the case a rejected state.

## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/app/controllers/cases/offender_sar_controller.rb
+++ b/app/controllers/cases/offender_sar_controller.rb
@@ -29,7 +29,7 @@ module Cases
 
     def new
       permitted_correspondence_types
-      @is_rejected = params["rejected"]
+      @rejected = params["rejected"]
       authorize case_type, :can_add_case?
       @case = build_case_from_session(case_type)
       @case.current_step = params[:step]
--- a/app/controllers/cases/offender_sar_controller.rb
+++ b/app/controllers/cases/offender_sar_controller.rb
@@ -331,12 +331,12 @@ module Cases
       request_dated_exists = values.fetch("request_dated", false)
       values["request_dated"] = nil unless request_dated_exists
 
-      rejected_set_current_state(params, values)
+      rejected_set_current_state(values)
 
       correspondence_type.new(values).decorate
     end
 
-    def rejected_set_current_state(params, values)
+    def rejected_set_current_state(values)
       case params["rejected"]
       when "true"
         values["current_state"] = "rejected"
--- a/app/controllers/cases/offender_sar_controller.rb
+++ b/app/controllers/cases/offender_sar_controller.rb
@@ -29,7 +29,7 @@ module Cases
 
     def new
       permitted_correspondence_types
-      @is_rejected = params.key?("rejected")
+      @is_rejected = params["rejected"]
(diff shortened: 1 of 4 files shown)


app/controllers/cases/offender_sar_controller.rb
```diff
@@ -330,11 +330,21 @@ def build_case_from_session(correspondence_type)
...
 
       correspondence_type.new(values).decorate
     end
 
+    def rejected_set_current_state(params, values)
+      case params["rejected"]
+      when "true"
+        values["current_state"] = "rejected"
+      when "false"
+        values["current_state"] = "data_to_be_requested"
```
lucas-shaw, line 344: Not sure whether its good to hard code the 'data_to_be_requested" state, however not sure how else to do it

app/views/cases/offender_sar/_subject_details_step.html.slim
```diff
@@ -32,7 +32,7 @@
     - fieldset.radio_input(true, text_method: :humanize)
     - fieldset.radio_input(false, text_method: :humanize)
 
-  = f.hidden_field :current_state, value: "rejected" if @is_rejected
+  = f.hidden_field :current_state, value: "#{@is_rejected == "true" ? 'rejected' : 'data_to_be_requested'}"
```
lucas-shaw, line 35: as above

app/controllers/cases/offender_sar_controller.rb
vertism: What would happen if you removed the value?
```
values.delete("current_state")
```

app/views/cases/offender_sar/_subject_details_step.html.slim
vertism: Why is this field needed?

app/controllers/cases/offender_sar_controller.rb
```diff
@@ -330,11 +330,21 @@ def build_case_from_session(correspondence_type)
...
       request_dated_exists = values.fetch("request_dated", false)
       values["request_dated"] = nil unless request_dated_exists
-      values["current_state"] = "rejected" if params.key?("rejected")
+
+      rejected_set_current_state(params, values)
 
       correspondence_type.new(values).decorate
     end
 
+    def rejected_set_current_state(params, values)
```
vertism, line 339: params is available to the class. It doesn't need to be passed as an argument.

app/controllers/cases/offender_sar_controller.rb
```diff
@@ -29,7 +29,7 @@ def initialize
 
     def new
       permitted_correspondence_types
-      @is_rejected = params.key?("rejected")
+      @is_rejected = params["rejected"]
```
vertism, line 32: If this isn't a boolean then maybe best to change from being called `is_`

vertism (approved):
LGTM

PR #2327, opened 2024-02-09
Cdpt 1125 information received page

lucas-shaw:
## Description
Creation of the information received page when updating the rejected reasons for an offender SAR.

## Self-review checklist

* [x] (3) Tests passing


app/controllers/cases/offender_sar_controller.rb
```diff
@@ -210,6 +212,27 @@ def confirm_sent_to_sscl
       redirect_to case_path(@case) and return
     end
 
+    def information_received
```
vertism, line 215: This method isn't explicitly needed if this is all it is doing. It should render without the method.

app/controllers/cases/offender_sar_controller.rb
```diff
@@ -210,6 +212,27 @@ def confirm_sent_to_sscl
...
+      render :information_received
+    end
+
+    def confirm_information_received
+      # NoMethodError is raised if calling :information_received
+      # as its default value is nil
+      information_received = params[:offender_sar].try(:[], :information_received)
```
vertism, line 222: My preference would be to have a guard clause here like
```
return unless params[:offender_sar].fetch(:information_received, nil)
```

app/views/cases/offender_sar/information_received.html.slim
```diff
@@ -0,0 +1,20 @@
...
+
+= render partial: 'layouts/header'
+
+= GovukElementsErrorsHelper.error_summary @case.object,
+  "#{pluralize(@case.errors.count, t('common.error'))} #{ t('common.summary_error')}", ""
+
+= form_for @case, url: confirm_information_received_case_sar_offender_path(@case), as: :offender_sar do |f|
+  = f.radio_button_fieldset :information_received, choices: [:yes, :no]
+
+  input name="current_step" type="hidden" value=@case.current_step
```
vertism, line 18: Is this needed?

app/views/cases/offender_sar/information_received.html.slim
```diff
@@ -0,0 +1,20 @@
+- content_for :page_title do
+  = t('page_title.information_received', case_number: @case.number)
+
+- content_for :heading, flush: true
+  = t('cases.offender_sar.information_received_heading')
+
+- content_for :sub_heading
+  = t('cases.edit.offender_sar.rejected_sub_heading')
```
vertism, line 8: This should use the `get_heading` method to be consistent.

config/locales/en.yml
```diff
@@ -1519,12 +1523,17 @@ en:
...
       edit_partial_case_flags_heading: Edit partial case information
       partial_case_flags_heading: SSCL COVID-19 partial case
+      information_received_heading: Can the Subject Access Request be accepted?
       is_partial_case_flag_heading: Confirmed SSCL COVID-19 partial case
       partial_case_letter_sent_dated_heading: Date letter sent from SSCL
       further_actions_required_heading: SSCL managing case
       flag_choice_yes: "Yes"
       flag_choice_no: "No"
       sent_to_sscl_heading: "Data sent to SSCL"
+      information_received:
```
lucas-shaw, line 1533: This works however not sure if its the best way to implement due to extra nesting

app/controllers/cases/offender_sar_controller.rb
lucas-shaw: Adding this at the top raises undefined method `fetch' for nil:NilClass

lucas-shaw: Removed the "yes" "no" conditional for now as its not required in this ticket.

vertism (approved):

PR #2337, opened 2024-02-27
CDPT: 1075 CT Rejections Back Link

MeenaModhvadia:
## Description
During the creation of a rejected offender sar if the back link is used then the progress of the flow changes to an accepted offender sar.

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-1075


spec/features/cases/offender_sar/case_creating_spec.rb
```diff
@@ -249,7 +265,7 @@ def expect_to_have_correct_sending_address
     expect(cases_show_page).to have_content "Address\n22 High Street"
   end
 
-  def and_back_previous_step_to_change_to_non_third_party
```
vertism, line 252: This method is used elsewhere, so you'll need to make the name change there as well.

spec/features/cases/offender_sar/case_creating_spec.rb
```diff
@@ -24,6 +24,22 @@
     then_expect_open_cases_page_to_be_correct
   end
 
+  scenario "0 Rejected offender SAR created after using back link" do
```
vertism, line 27: You've got two "0" scenarios now. I'm not sure why they are numbered so you could update or remove the numbering.

app/controllers/cases/offender_sar_controller.rb
```diff
@@ -391,7 +391,9 @@ def has_optional_flags?
     end
 
     def back_link_url
-      if @case.get_previous_step
+      if @case.get_previous_step && @case.get_previous_step == "subject-details"
```
vertism, line 394: I think this logic should be done in `build_url_params_from_flags`, as that is responsible for adding query params

vertism (approved):
LGTM

PR #2339, opened 2024-02-29
Cdpt 1037 record date info received and valid date creation

lucas-shaw:
## Description
When a user confirms that all the required information has been received, they click the "create valid case" link on the case details page. They are navigated to the "accepted_date_received" page to input the date they received the new information. Once submitted, the case state changes from "rejected" to "data to be requested" and the date received is updated.  

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?selectedIssue=CDPT-1037
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?selectedIssue=CDPT-1038


app/controllers/cases/offender_sar_controller.rb
```diff
@@ -212,15 +212,37 @@ def confirm_sent_to_sscl
       redirect_to case_path(@case) and return
     end
 
-    def confirm_information_received
-      information_received = params[:offender_sar].try(:[], :information_received)
-      return if information_received.present?
+    def accepted_date_received
+      @case = Case::Base.find_by(id: params[:id])
```
vertism, line 216: This should be unnecessary as it's being done in the set_decorated_case before_action. Same with `confirm_accepted_date_received`

app/controllers/cases/offender_sar_controller.rb
```diff
@@ -212,15 +212,37 @@ def confirm_sent_to_sscl
...
+      @case.received_date_dd = ""
```
vertism, line 217: Does `@case.received_date = nil` not work?

app/controllers/cases/offender_sar_controller.rb
```diff
@@ -212,15 +212,37 @@ def confirm_sent_to_sscl
...
-      )
-      render :information_received and return
+    def confirm_accepted_date_received
+      @case = Case::Base.find_by(id: params[:id])
+      authorize @case, :can_edit_case?
+
+      received_date = params[:offender_sar].permit(:received_date_dd, :received_date_mm, :received_date_yyyy)
+      service = case_updater_service.new(current_user, @case, received_date)
+      service.call
+      @case.number.tr!("R", "")
```
vertism, line 229: This shouldn't be done here. It would be better to create a service to:
- Update the case
- Change the case number
- Transition the state
- Add any history messages

app/controllers/concerns/offender_sar_cases_params.rb
```diff
@@ -20,6 +20,7 @@ def create_offender_sar_params
       :prison_number,
       :previous_case_numbers,
       :received_date_dd, :received_date_mm, :received_date_yyyy,
+      :accepted_date_received_dd, :accepted_date_received_mm, :accepted_date_received_yyyy,
```
vertism, line 23: I don't think this is needed

app/models/case/sar/offender.rb
```diff
@@ -102,7 +102,7 @@ def searchable_fields_and_ranks
   attribute :number_final_pages, :integer, default: 0
   attribute :number_exempt_pages, :integer, default: 0
 
-  attr_accessor :remove_sent_to_sscl_reason, :information_received
+  attr_accessor :remove_sent_to_sscl_reason, :accepted_date_received_dd, :accepted_date_received_mm, :accepted_date_received_yyyy
```
vertism, line 105: I don't think this is needed

app/views/cases/offender_sar/_accepted_date_received_error.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="error-summary" role="alert" aria-labelledby="error-summary-heading" tabindex="-1">
+  <h2 id="error-summary-heading" class="heading-medium error-summary-heading"><%= "#{pluralize(@case.errors.count, t('common.error'))} #{ t('common.summary_error')}" %></h2>
+  <p></p>
+  <ul class="error-summary-list">
+    <li>
+      <a href="#error_offender_sar_received_date">When was the outstanding information received? <%= "#{@case.errors.first.message}" %></a>
```
vertism, line 6: It might be better to access the error by key to ensure it is the correct one.

app/views/cases/offender_sar/_case_details.html.slim
```diff
@@ -22,7 +22,7 @@
         - if case_details.third_party_recipient?
           = render partial: 'cases/offender_sar/third_party_details', locals: { case_details: case_details, allow_editing: allow_editing }
 
-        - if case_details.rejected?
+        - if case_details.has_attribute?(:case_originally_rejected) && case_details.case_originally_rejected
```
vertism, line 25: Can you do `case_details&.case_originally_rejected?`?

app/views/cases/offender_sar/_reason_rejected_details.html.slim
```diff
@@ -1,4 +1,3 @@
 tr.rejected-reasons.section
   th = t('common.case.rejected_reason')
   td = raw case_details.rejected_reasons_descriptions
-  td = link_to t('common.links.change'), edit_step_case_sar_offender_path(case_details, "reason_rejected")
```
vertism, line 4: Is this page still accessible?

app/controllers/cases/offender_sar_controller.rb
```diff
@@ -212,15 +212,38 @@ def confirm_sent_to_sscl
       redirect_to case_path(@case) and return
     end
 
-    def confirm_information_received
-      information_received = params[:offender_sar].try(:[], :information_received)
-      return if information_received.present?
+    def accepted_date_received
+      authorize @case, :can_edit_case?
+      authorize @case, :can_validate_rejected_case?
```
vertism, line 217: Rather than calling making 2 authorisation checks, could `can_validate_rejected_case?` also check the user can edit the case?

app/services/case_validate_rejected_offender_sar_service.rb
```diff
@@ -0,0 +1,31 @@
...
+    @params = params
+    @result = :incomplete
+    @error_message = nil
+  end
+
+  def call(message = nil)
+    ActiveRecord::Base.transaction do
+      @case.assign_attributes(@params)
+      @case.save # rubocop:disable Rails/SaveBang
+      @case.number.tr!("R", "")
```
vertism, line 16: I think it would be useful to extract the number change into its own method for clarity.

spec/services/case_validate_rejected_offender_sar_service_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        service.call
+        expect(service.result).to eq :ok
+      end
+    end
+
+    context "when anything fails in the transaction" do
+      let(:params) { { received_date: "" } }
+
+      it "raises an error when it saves" do
+        allow(kase).to receive(:save).and_raise(RuntimeError)
```
vertism, line 50: Calling `save` won't raise an error. 
It would be better to use invalid params rather than mocking the method so you are fully testing what the method does.

app/policies/case/sar/offender_policy.rb
```diff
@@ -55,9 +55,10 @@ def can_send_dispatch_letter?
     check_user_can_manage_offender_sar
   end
 
-  def handle_information_received?
+  def can_validate_rejected_case?
     clear_failed_checks
-    check_user_can_manage_offender
+    edit_case?
```
vertism, line 60: Do you need to check this together with the next line?
If `edit_case?` returns false and `check_can_trigger_event(:validate_rejected_case)` returns true then `can_validate_rejected_case?` will still return true.

app/models/case/sar/offender.rb
```diff
@@ -451,6 +451,14 @@ def user_dealing_with_vetting
     user_for_vetting
   end
 
+  # Overwrites base method to allow case number to remove "R" when
+  # transitioning from 'rejected' to 'valid' offender SAR
+  def prevent_number_change
+    return unless current_state == "rejected" && number_changed?
```
vertism, line 457: You should test this to make sure it's doing what you expect.

app/models/case/sar/offender.rb
```diff
@@ -454,9 +454,7 @@ def user_dealing_with_vetting
   # Overwrites base method to allow case number to remove "R" when
   # transitioning from 'rejected' to 'valid' offender SAR
   def prevent_number_change
-    return unless current_state != "rejected" && number_changed?
-
-    raise StandardError, "number is immutable"
+    raise StandardError, "number is immutable" if !case_originally_rejected && number_changed?
```
lucas-shaw, line 457: Used the case_originally_rejected flag as save is also called from machine.rb write_transition method. The second time prevent_number_change is called the current_state is not rejected which causes the error to be raised which is not wanted.

vertism: The issue is that there is nothing to stop a previously rejected case having its number changed.

So when `write_transition` is called, `number_changed?` returns true?

spec/services/case_validate_rejected_offender_sar_service_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    let(:kase)          { create :offender_sar_case, :rejected, received_date: 1.day.ago.to_date }
+    let(:state_machine) { double ConfigurableStateMachine::Machine,  validate_rejected_case!: true } # rubocop:disable RSpec/VerifiedDoubles
+    let(:service)       do
+      described_class.new(user:,
+                          kase:,
+                          params:)
+    end
+
+    before do
+      allow(kase).to receive(:state_machine).and_return(state_machine)
```
vertism, line 16: I'm not sure this is necessary. Can't you check the actual state_machine received the expected method call:
```
expect(kase.state_machine).to receive(:validate_rejected_case!)
```

vertism (approved):
Nice work

PR #2340, opened 2024-03-04
Remove print

vertism:
## Description
The `p` used here is not a paragraph tag but surprisingly the alias of `print`! It's causing the a tag to be output in the logs which is making the test logs quite noisy.


MeenaModhvadia (approved):
LGTM

PR #2347, opened 2024-03-13
Monthly report to exclude rejected cases

lucas-shaw:
## Description
Rejected offender SAR cases to be excluded from the monthly reports and custom reports.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?selectedIssue=CDPT-1366


app/services/stats/r205_offender_sar_monthly_performance_report.rb
```diff
@@ -23,7 +23,7 @@ def case_analyzer
     end
 
     def case_scope
-      Case::Base.offender_sar.where(received_date: @period_start..@period_end)
+      Case::Base.offender_sar.where(received_date: @period_start..@period_end).where.not(current_state: "rejected")
```
vertism, line 26: Should this also exclude rejected cases that have been closed?

app/services/stats/r205_offender_sar_monthly_performance_report.rb
```diff
@@ -23,7 +23,8 @@ def case_analyzer
     end
 
     def case_scope
-      Case::Base.offender_sar.where(received_date: @period_start..@period_end)
+      number = Case::Base.arel_table[:number]
```
vertism, line 26: I think it might be worth creating a scope on the model for "not_rejected" as we might want to re-use it and naming it will provide some self documentation.

I'm sure what you've done works, but I'm not really familiar with the code you've used.
An alternative would be to use something like
```
where.not("LEFT(cases.number, 1) = 'R'")
```

If we are using the number, then potentially we don't need the state check as well.

app/services/stats/r205_offender_sar_monthly_performance_report.rb
```diff
@@ -23,7 +23,7 @@ def case_analyzer
     end
 
     def case_scope
-      Case::Base.offender_sar.where(received_date: @period_start..@period_end)
+      Case::SAR::Offender.offender_sar.not_rejected.where(received_date: @period_start..@period_end)
```
vertism, line 26: You can leave this as `Case::Base.offender_sar`

vertism (approved):
LGTM

PR #2355, opened 2024-03-18
Rename rejected state

vertism:
## Description
Changing the initial state of a rejected case to "invalid_submission" to remove confusion when between a case in that state and one that is still not accepted that has been closed.

## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/spec/validators/closed_case_validator_spec.rb
+++ b/spec/validators/closed_case_validator_spec.rb
@@ -74,7 +74,7 @@ describe "ClosedCaseValidator" do
       end
 
       it "does not error with blank date_responded when case is rejected" do
-        kase.current_state = "rejected"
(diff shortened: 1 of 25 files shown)


spec/validators/closed_case_validator_spec.rb
```diff
@@ -74,7 +74,7 @@
       end
 
       it "does not error with blank date_responded when case is rejected" do
```
lucas-shaw, line 76: Should this say "when case is invalid_submission"?

vertism: I think this is OK, because it's saying the case has been rejected and the initial state of a rejected case is "invalid_submission".

lucas-shaw (approved):
LGTM

PR #2360, opened 2024-03-20
Cdpt 1464 correct styling for when other is selected for rejected reason

lucas-shaw:
## Description
1. Ensure that the other rejected reason format is inline, rather than on seperate lines (see image)
2. Fix bug - if the "other" checkbox is deselected but an other reason was given, remove the other reason (see image)

## Self-review checklist

* [ ] (3) Tests passing

### Screenshots

Before this PR the other option and its reason aren't on the same line
Screenshot 2024-03-20 at 09 55 12

Before this PR the other rejected reason still has a value
Screenshot 2024-03-20 at 09 56 05

### DIFF
--- a/app/decorators/case/sar/offender_decorator.rb
+++ b/app/decorators/case/sar/offender_decorator.rb
@@ -22,8 +22,14 @@ class Case::SAR::OffenderDecorator < Case::SAR::OffenderBaseDecorator
   end
 
   def rejected_reasons_descriptions
+    other_string = if other_rejected_reason.present?
+      ": #{other_rejected_reason}"
+    else
+      ""
+                   end
+
     rejected_reasons.map { |reason|
       Case::SAR::Offender::REJECTED_REASONS[reason]
-    }.append(other_rejected_reason).compact.join("<br>")
+    }.compact.join("<br>") + other_string
   end
 end
--- a/app/decorators/case/sar/offender_decorator.rb
+++ b/app/decorators/case/sar/offender_decorator.rb
@@ -22,7 +22,7 @@ class Case::SAR::OffenderDecorator < Case::SAR::OffenderBaseDecorator
   end
 
   def rejected_reasons_descriptions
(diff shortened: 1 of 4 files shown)


app/decorators/case/sar/offender_decorator.rb
```diff
@@ -22,8 +22,14 @@ def request_methods_for_display
...
+    other_reason_text = if other_rejected_reason.present?
+                          ": #{other_rejected_reason}"
+                        else
+                          ""
+                        end
+
     rejected_reasons.map { |reason|
       Case::SAR::Offender::REJECTED_REASONS[reason]
-    }.append(other_rejected_reason).compact.join("<br>")
+    }.compact.join("<br>") + other_reason_text
```
vertism, line 33: This relies on OTHER being the last option in the REJECTED_REASONS hash. This is true now, but could change in theory. 
It might then be better to append the other reason within the map when it is the relevant object.

app/decorators/case/sar/offender_decorator.rb
```diff
@@ -23,7 +23,11 @@ def request_methods_for_display
 
   def rejected_reasons_descriptions
     rejected_reasons.map { |reason|
-      Case::SAR::Offender::REJECTED_REASONS[reason]
-    }.append(other_rejected_reason).compact.join("<br>")
+      if reason != "other"
+        Case::SAR::Offender::REJECTED_REASONS[reason]
+      else
+        "Other: #{other_rejected_reason}"
+      end
```
vertism, line 30: Just thinking about this a little more and my personal preference would be to try to use a positive IF statement where possible. Also it does makes sense to still use the value from the REJECTED_REASONS hash as you were doing originally so maybe something like:
```suggestion
      suffix = ": #{other_rejected_reason}" if reason == "other"
      "#{Case::SAR::Offender::REJECTED_REASONS[reason]}#{suffix}"
```

vertism (approved):
LGTM

PR #2363, opened 2024-03-21
Fix tab appearance

vertism:
## Description
Additional CSS brought in with the crown change caused tabs to partially break. This fixes the issue.

Before:
Screenshot 2024-03-21 at 17 17 19

After:
Screenshot 2024-03-21 at 17 17 05


lucas-shaw (approved):
LGTM

PR #2364, opened 2024-03-22
Cdpt 1249 new page to select contact type

lucas-shaw:
## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?selectedIssue=CDPT-1249


app/controllers/contacts_controller.rb
```diff
@@ -61,16 +72,18 @@ def set_contact
   end
 
   def set_new_contact_from_params
-    @contact = Contact.new(contact_params)
+    @contact = Contact.last.update!(contact_params)
```
vertism, line 75: You can't rely on `last` giving you the expected Contact because 2 users could be creating an address at the same time.

app/controllers/contacts_controller.rb
```diff
@@ -1,15 +1,34 @@
...
+
+    if request.get?
+      @contact_type = ContactType.new
+      return false
+    end
+
+    @contact_type = ContactType.new(contact_type_params)
+
+    if @contact_type.valid?
+      redirect_to new_new_details_contacts_path
```
vertism, line 17: When this redirect happens the params are being lost. You either need to add them to the URL, or submit this form to the new_details page.

vertism: My preference would be to submit to the next page and if the form object is invalid, re-render the new page.

spec/models/contact_spec.rb
```diff
@@ -64,6 +64,20 @@
...
+    it "prevents contact_type_id being updated" do
+      contact = described_class.new(name: "HMP halifax",
+                                    address_line_1: "123 test road",
+                                    address_line_2: "little heath",
+                                    town: "bakersville",
+                                    county: "Mercia",
+                                    postcode: "FE2 9JK",
+                                    data_request_emails: "test@test.com\ntest1@test.com",
+                                    contact_type_id: 1)
+      expect { contact.update_attribute(:contact_type_id, 2) }.to raise_error(ActiveRecord::ActiveRecordError, "contact_type_id is marked as readonly") # rubocop:disable Rails/SkipsModelValidations
```
lucas-shaw, line 77: Wasn't quite sure how to test this or whether using "attr_readonly" was the best approach. 

Using 'update!' raised an error saying "contact_type cant be blank". Tried to implement the CategoryReference model but was unclear to me whether that was the best way to do it

app/controllers/contacts_controller.rb
```diff
@@ -1,15 +1,26 @@
 class ContactsController < ApplicationController
   before_action :set_contact, only: %i[edit update destroy]
-  before_action :set_contact_type_options, only: %i[create edit new update]
```
vertism, line 3: You could bring back the `set_contact_type_options` before_action for :new and :new_details

app/models/contact.rb
```diff
@@ -5,6 +5,8 @@ class Contact < ApplicationRecord
   validates :contact_type, presence: true
   validate :validate_emails
 
+  attr_readonly :contact_type_id
```
vertism, line 8: Is this needed?

app/views/contacts/_form.html.slim
```diff
@@ -1,20 +1,15 @@
...
+  = f.text_field :postcode
+  = f.text_area :data_request_name
+  = f.text_area :data_request_emails
 
-    .button-holder
-      = f.submit 'Submit', class: 'button'
+  = f.hidden_field :contact_type_id
 
-    .button-holder
-      = link_to('Cancel', contacts_url)
```
vertism, line 20: Should there still be cancel link?

config/routes.rb
```diff
@@ -1,7 +1,11 @@
 require "sidekiq/web"
 
 Rails.application.routes.draw do
-  resources :contacts, except: :show
+  resources :contacts, except: :show do
+    get "/new_details", on: :collection, to: "contacts#new_details"
+    post "/new_details", on: :collection, to: "contacts#new_details"
+    post "/new", on: :collection, to: "contacts#new"
```
vertism, line 7: Is there a post to /new?

app/models/contact.rb
vertism: I understand why you've done this from looking at the test. The better way to prevent updating an attribute is via StrongParameters in the controller. You can have a different set of allowed params for create and update.

app/controllers/contacts_controller.rb
```diff
@@ -1,15 +1,26 @@
...
   before_action :set_new_contact_from_params, only: :create
```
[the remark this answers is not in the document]
lucas-shaw: Moving this seems to cause a "undefined method `contact_type=' for nil:NilClass" when submitting the new_details page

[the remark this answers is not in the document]
vertism: OK, so the issue is that the `set_contact_type` method relies on the `@contact` variable having been set. However, I'm not sure `set_contact_type` needs to exist. 
If validation of the contact_type_id is required, then it should be done in the form model instead.

spec/controllers/contacts_controller_spec.rb
```diff
@@ -23,4 +23,124 @@
...
+        expect(response.body).to redirect_to new_contact_path
+      end
+    end
+  end
+
+  describe "#create" do
+    let(:contact_type) { create(:category_reference) }
+    let(:params) do
+      {
+        id: stafford.id,
```
vertism, line 85: A create request shouldn't contain an id.

spec/controllers/contacts_controller_spec.rb
```diff
@@ -23,4 +23,124 @@
...
+    before do
+      sign_in user
+    end
+
+    it "redirects to the contacts index page" do
+      patch(:update, params:)
+      expect(response).to redirect_to(contacts_path)
+    end
+
+    it "does not update contact_type_id value" do
```
vertism, line 137: I'm finding this method a bit confusing.
I was thinking you could do something like:
```
expect {
  patch(:update, params:)
}.not_to(
  change { stafford.reload.contact_type },
)
```

Also, be careful with which contact you are updating, as the params are using `stafford`, but you are checking `contact`

vertism (approved):
LGTM

PR #2366, opened 2024-03-25
Remove blank reasons

vertism:
## Description
Remove blank rejected reason to avoid a blank line showing on the cases page.


njpreddy (approved):
LGTM

PR #2367, opened 2024-03-25
Add rejected title to closed case

vertism:
## Description
Update header on closed rejected Offender SAR cases to add "Rejected".

The decorated pretty_type method is updated to use the `rejected?` method which returns true for cases that are initially rejected and also for cases that were rejected, and have since been closed.

The `rejected?` method has been updated to handle new cases that have not yet been saved, so have not been assigned a number.


njpreddy (approved):
LGTM :-)

PR #2372, opened 2024-03-28
Bug fix - Rejected case editing subject details 

lucas-shaw:
## Description
Bug fix - when editing the subject details of a rejected case or rejected-closed case, the status changed to "data to be requested"

## Self-review checklist

* [ ] (3) Tests passing


app/controllers/cases/offender_sar_controller.rb
```diff
@@ -141,7 +141,7 @@ def confirm_update_partial_flags
     def edit
       permitted_correspondence_types
       authorize case_type, :can_add_case?
-
+      get_rejected
```
vertism, line 144: I don't think this is needed if the current_state can't be updated

vertism (approved):
LGTM

PR #2374, opened 2024-04-02
Add date of birth validation

vertism:
## Description
Currently the immediate validation passes when an invalid date (e.g. 0000-00-00) is entered when creating a case because the date is converted to nil. The validation has changed to check for nil.

Also removes boolean returns from some validation methods, as this is no longer required since Rails 5


MeenaModhvadia (approved):
LGTM

PR #2375, opened 2024-04-02
CDPT-1480 Update sub-heading when editing rejected offender SAR

vertism:
## Description
Refactor get_heading method and make work with more sub_headings


MeenaModhvadia (approved):
Were these changes not made before?

vertism:
> Were these changes not made before?

@MeenaModhvadia Not as far as I'm aware. The changes were made for the creation pages, but Alan spotted that the heading needed to be changed when editing cases.

PR #2377, opened 2024-04-08
Upgrade to Rails 7.1.3.2

vertism:
## Description
Upgrading Rails to latest version

### Additional work:
All Rails 6.1 defaults have been enabled including using the Zeitwerk autoloader that is required in Rails 7.

Zeitwerk changes the naming rules, so any class name with an acronym (e.g. SAR) has been added to the inflector settings so it can be all capitalised (and must now be capitalised).

Other changes include:
- Updating ReportType class names in database to match changes in the app.
- Removing a deprecated command to setup the test database.
- Removing Spring


lucas-shaw (approved):
LGTM

PR #2379, opened 2024-04-11
Cdpt 1516 update case dpa number when rejected becomes valid

lucas-shaw:
## Description
After validating a rejected case, ensure that the case number is updated. Following the usual process of defining the number from the received date. 

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
(https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?selectedIssue=CDPT-1516)

### DIFF
--- a/app/models/case/sar/offender.rb
+++ b/app/models/case/sar/offender.rb
@@ -468,6 +468,10 @@ class Case::SAR::Offender < Case::Base
     raise StandardError, "number is immutable" if current_state != "invalid_submission" && number_changed?
   end
 
+  def set_valid_case_number
+    self.number = next_number
+  end
+
 private
 
   def set_subject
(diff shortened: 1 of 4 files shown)


app/models/case/sar/offender.rb
```diff
@@ -468,6 +468,10 @@ def prevent_number_change
     raise StandardError, "number is immutable" if current_state != "invalid_submission" && number_changed?
   end
 
+  def set_valid_case_number
```
lucas-shaw, line 471: Created this new method rather than using the set_number private method. As it would need to moved to a public method and is used only for the case creation process. Thought it best to separate the two.

app/services/case_validate_rejected_offender_sar_service.rb
```diff
@@ -32,6 +32,6 @@ def call(message = nil)
 private
 
   def update_case_number
-    @case.number.tr!("R", "")
+    @case.number.tr!(@case.number, @case.set_valid_case_number)
```
vertism, line 35: You should be able to just call `@case.set_valid_case_number`. 

Also, as the method name is clear, it can probably be called directly from the call method.

vertism (approved):
LGTM

PR #2380, opened 2024-04-22
Fix rails 7.1 issues

vertism:
## Description
- Updated Gems: sidekiq, config, rails-data-migrations
- Added tmp/pids directory to build
- Update DB healthcheck to simple query
- Use "bundle exec" with migration task
- Move Enumerable addition to initializer


njpreddy (approved):
LGTM

PR #2383, opened 2024-04-24
Include the rejected flag and rejected reason to warehouse 

lucas-shaw:
## Description
Allow a Branston Metabase user to select the rejected cases in Metabase.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1152?selectedIssue=CDPT-1465


app/models/warehouse/case_report.rb
```diff
@@ -125,7 +125,6 @@ def generate(kase)
         case_report.number = kase.number
         case_report.case_type = kase.decorate.pretty_type
         case_report.current_state = kase.decorate.status
-        case_report.rejected = kase.rejected? ? "Yes" : "No"
```
lucas-shaw, line 128: Removed this with the thought we can rely on the case_originally_rejected flag and the current state for filtering rejected cases

spec/models/warehouse/case_report_spec.rb
```diff
@@ -87,6 +87,7 @@
       create(:closed_case, :fully_refused_exempt_s40, :extended_for_pit),
       create(:sar_case, :extended_deadline_sar),
       create(:offender_sar_case, :third_party),
+      create(:offender_sar_case, :rejected),
```
lucas-shaw, line 90: Spent a lot of time trying to add in tests for the rejected_reason however when testing the case_report output the tests seem very fiddly when dealing with the string formatting. E.g it changes the [] into "[]" and also also adds in escape chars ("/" and "\") when dealing with the rejected_reason string values. So left them out, the third party stuff doesnt seem to have tests either

app/models/warehouse/case_report.rb
vertism: This attribute is being used in reports in the app so it's needed for that. (which is the reason it's been added)

The easiest way to filter out rejected cases will be to use this attribute, so whilst there might be another way to do it, it would be more complicated and require implementation logic and I'm not sure what the benefit would be.

In theory, regular users should be able to make any report they want in metabase, so having a boolean value for this is also easiest for them.

vertism (approved):
LGTM

PR #2384, opened 2024-04-25
CDPT-1535 Change default mailer queue

vertism:
## Description
Rails 7 has changed the name of the mailer queue from "mailers" to "default"
Sidekiq is monitoring the "mailers" queue, so this config will update the queue name to be "mailers" again.


njpreddy (approved):
LGTM

PR #2385, opened 2024-04-26
Fix updating email status

vertism:
## Description
The rails 7 upgrade has caused the mail observer to stop working as expected. This has meant the Notify ID was not getting set for a data request email, and therefore the status of the email could not be updated.

The underlying cause appears to be that actionmailer no longer calls the `deliver!` method when a non-bang method is used, which means that the relevant method in the govuk_notify_rails gem was not being called.

The fix is to use the `deliver_later!` method instead of `deliver_later` and I have also implemented the new mailer callback feature that is new in Rails 7.1, and replaces the old observer pattern.


njpreddy (approved):
LGTM

PR #2392, opened 2024-05-08
Fix class names in migration files for zeitwerk

vertism:
## Description
Data migration filenames need changing to match what zeitwerk autoloader expectes.


lucas-shaw (approved):
LGTM

PR #2407, opened 2024-06-21
CDPT-1565: Add High profile badge

MeenaModhvadia:
## Description
As a Branston User, I want to see High Profile Offender SAR case.
So that I can I am able to priorities those cases

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-1565


spec/decorators/case/sar/offender_decorator_spec.rb
```diff
@@ -134,4 +134,17 @@
       expect(offender_sar_case.request_methods_for_display).to eq %w[email post web_portal]
     end
   end
+
+  describe "#is_high_profile?" do
```
vertism, line 138: This is the wrong method name

app/models/case/sar/offender.rb
```diff
@@ -505,6 +505,10 @@ def set_valid_case_number
     self.number = next_number
   end
 
+  def is_high_profile?
```
vertism, line 508: Is this method needed?

vertism: If it were needed, `flag_as_high_profile?` is a boolean, so it will already return true or false. You don't need to check it's value.

app/decorators/case/sar/offender_decorator.rb
```diff
@@ -27,4 +27,13 @@ def rejected_reasons_descriptions
       "#{Case::SAR::Offender::REJECTED_REASONS[reason]}#{suffix}"
     }.compact.join("<br>")
   end
+
+  def highlight_flag
+    if object.flag_as_high_profile?
+      '<div class="offender_sar-profile_flag"><span class="visually-hidden">' \
+        'This is a </span>High priority<span class="visually-hidden"> case</span></div>'
```
vertism, line 34: This is the wrong text value. According to the ticket it should be "High Profile"

spec/decorators/case/sar/offender_decorator_spec.rb
```diff
@@ -134,4 +134,17 @@
       expect(offender_sar_case.request_methods_for_display).to eq %w[email post web_portal]
     end
   end
+
+  describe "#highlight_flag" do
+    it 'returns string of "High profile" in a badge' do
```
vertism, line 139:
```suggestion
    it "returns string of 'High profile' in a badge" do
```

vertism (approved):

PR #2409, opened 2024-06-26
CDPT-1698: Display additional details in the data request view

MeenaModhvadia:
## Description
Recording a data request where the information being requested is Nomis other or Other the user provides additional details which are not displayed in the view.

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-1698

### DIFF
--- a/app/views/cases/data_requests/show.html.slim
+++ b/app/views/cases/data_requests/show.html.slim
@@ -29,7 +29,7 @@ span.visually-hidden
       p.data_request_request_type
         = t("helpers.label.data_request.request_type.#{@data_request.request_type}")
         - if @data_request.request_type_note?
-          = ":\n #{@data_request.request_type_note}"
+          = ":<br>#{@data_request.request_type_note}".html_safe
     .form-group
       label.form-label
         = t('helpers.fieldset.data_request.date_requested')
--- a/app/views/cases/data_requests/show.html.slim
+++ b/app/views/cases/data_requests/show.html.slim
@@ -30,7 +30,7 @@ span.visually-hidden
         = t("helpers.label.data_request.request_type.#{@data_request.request_type}")
         - if @data_request.request_type_note?
           =":"
-          <div id="request_type_note_other">"#{@data_request.request_type_note}"
+          div id="request_type_note_other" "#{@data_request.request_type_note}"
     .form-group
       label.form-label
         = t('helpers.fieldset.data_request.date_requested')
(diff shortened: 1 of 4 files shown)


app/views/cases/data_requests/show.html.slim
```diff
@@ -29,7 +29,8 @@ span.visually-hidden
       p.data_request_request_type
         = t("helpers.label.data_request.request_type.#{@data_request.request_type}")
         - if @data_request.request_type_note?
-          = ":<br>#{@data_request.request_type_note}"
+          =":"
+          <div>"#{@data_request.request_type_note}"
```
vertism, line 33: Try to use slim syntax in a slim file
```suggestion
          div= "#{@data_request.request_type_note}"
```

MeenaModhvadia: I've used div without the = due to using adding an id to the div if I used = it would print the id

app/views/cases/data_requests/show.html.slim
```diff
@@ -26,8 +26,11 @@ span.visually-hidden
     .form-group
       label.form-label
         = t('helpers.fieldset.data_request.request_type')
-      p.data_request_request_type
+      p.data_request_request_type style="margin-bottom: 0;"
```
vertism, line 29: It's best to avoid inline styles like this. 
I understand why you have done it, but as an alternative you could use a br within the p tag to add the extra data.

app/views/cases/data_requests/show.html.slim
```diff
@@ -26,8 +26,11 @@ span.visually-hidden
...
         = t("helpers.label.data_request.request_type.#{@data_request.request_type}")
+        - if @data_request.request_type_note?
+          =":"
+          div id="request_type_note_other" "#{@data_request.request_type_note}"
```
vertism, line 33:
```suggestion
          =":"
          br
          = @data_request.request_type_note
```

Your test can then check for the br being included with the other text.

app/views/cases/offender_sar/_data_request_request_type.html.slim
```diff
@@ -1,6 +1,7 @@
                 = t :"helpers.label.data_request.request_type.#{data_request.request_type}"
                 - if data_request.display_request_type_note?
-                  = ": #{data_request.request_type_note}"
+                  = ":"
+                  div "#{data_request.request_type_note}"
```
vertism, line 4: You can use `=` to render ruby. e.g.
```suggestion
                  div= data_request.request_type_note
```

vertism (approved):
LGTM

PR #2418, opened 2024-07-15
Speed up CI tests

vertism:
## Description
Increase speed of CI tests by running in parallel across multiple machines
Code coverage results are uploaded as artifacts, then collated and full results created in a separate step
Test timing stats have been updated so tests will be organised most efficiently

Results:
CircleCI: 12-15 minutes
Github Actions (GA) on 1 machine: 20-25 minutes
New GA solution: <5 minutes

---
Also, spec/features/cases/filters/external_deadline_spec.rb has been fixed as it was not previously running.


lucas-shaw (approved):
LGTM

PR #2419, opened 2024-07-16
Reduce unnecessary db calls

vertism:
## Description
The case history block on the case page was causing large amounts of DB calls. This has been improved by eager loading associations, and passing the user and team objects rather than an ID in places.


lucas-shaw (approved):
LGTM

PR #2428, opened 2024-08-01
Remove jquery-ui apart from dialog

vertism:
## Description
jquery-ui is only used for the dialog component, but we are loading all features.

Only loading the necessary component will reduce the size of the css and JS loaded on the client.


njpreddy (approved):
LGTM

PR #2430, opened 2024-08-01
Rewrite dockerfile

vertism:
## Description
Current docker image is 2.2GB. The new Dockerfile uses a builder image, and then the necessary files to a new image for deployment. This results in a much smaller image size.
Also removes various dependencies that are no longer required for the build.

Some shell scripts had the wrong permissions, and have been updated.


njpreddy (approved):
LGTM

PR #2440, opened 2024-08-07
CDPT-1922 Add zip dependency

vertism:
## Description
zip was removed in a recent cleanup of the dockerfile, but it is required for creating reports.


njpreddy (approved):
LGTM

PR #2441, opened 2024-08-07
Make running user owner of everything

vertism:
## Description
Change to the dockerfile means the app user could not make changes to temporary files.


wilson1000 (approved):
Looks good

PR #2443, opened 2024-08-08
Fix closed report bug

vertism:
## Description
The `exists?` check for whether the report data has been created in Redis was missing the `?`. This meant that if the background job had not finished processing, the report download would produce an empty file.

The README has also been updated to give details of how to start sidekiq locally.


njpreddy (approved):
LGTM

PR #2444, opened 2024-08-09
CDPT-1926 Add non-digested assets

vertism:
## Description
The error page is static and does not use fingerprinted assets.


njpreddy (approved):
LGTM

PR #2455, opened 2024-08-22
CDPT-1942 Allocate case to vetting team member

vertism:
## Description
When a Offender SAR case is "Ready for vetting" it can now optionally be assigned to a team member. This will make it appear on that user's "open cases" list, and will show up on the open cases report.

When the case is moved back or to "ready to copy", the assignment is removed.


MeenaModhvadia (approved):
LGTM

PR #2457, opened 2024-08-27
Add memory to pods

vertism:
## Description
Pods were running out of memory when we had issues with the database. 
Also updates the image name so it can be applied without going through CI.


njpreddy (approved):
LGTM

PR #2460, opened 2024-08-29
CDPT-1957 Add who made rejected case valid to warehouse

vertism:
## Description
Add the name of the person that validated a rejected case and made it valid to the warehouse case report.


MeenaModhvadia (approved):
LGTM

PR #2461, opened 2024-08-30
Make download test work locally

vertism:
## Description
This test was not working locally because the browser was using it's default download location.
The DownloadHelper helps manage the downloads and makes it consistent locally and in CI.


MeenaModhvadia (approved):
LGTM

PR #2465, opened 2024-09-05
CDPT-2004 Filter by assigned vetter

vertism:
## Description
Adds a filter to return Offender SAR cases by who is vetting them.


MeenaModhvadia (approved):
My understanding is that the relevant caseworker id’s are obtained from the defined statuses for Offender SAR’s.  A user’s scope is defined for these caseworker’s .
The scope takes an argument (user_id) and returns all  Offender SAR’s recorded with the user id.

vetter's id’s are obtained from the defined statuses of Offenders SAR’s and are added to an array.  This array is used to check if a user’s id is in the array. 

If it is then the case is added to the results and returned for the said vetter.

PR #2476, opened 2024-09-12
CDPT-2005 Reduce memory use when updating warehouse data

vertism:
## Description
Currently when an object with a lot of associated cases is updated, all related cases are loaded into memory at the same time and then the Warehouse::CaseReports are updated for each one.
When a user logs in, their database record is updated. This would then trigger a warehouse update which for a very active user could cause hundreds of cases to be loaded into memory, which could cause OOM issues.

This change only loads the IDs for each associated case, then creates update jobs for each one to be done separately. So only one case is ever loaded per job.


wilson1000 (approved):
Looks good 👍🏼

PR #2478, opened 2024-09-12
Limit which attributes cause warehouse update

vertism:
## Description
Currently when objects that include the Warehousable concern are updated, a job is triggered that will cause all associated cases to be updated in the warehouse table. This needs to happen because the case data is denormalised in that table.

However, what is being updated in the object is often not relevant to the case so the warehouse update is unnecessary. e.g. updating a user's last login data.

A new `warehousable_attributes` class method is now available to specify which updated fields should cause the warehouse sync to take place. If the method is not defined on a model then any update will still cause a sync.


njpreddy (approved):
LGTM

PR #2479, opened 2024-09-13
Allow multiple warehouseable attributes

vertism:
## Description
Extends warehousable_attributes so it can handle multiple attributes. Also stores the value as a string.


wilson1000 (approved):
LGTM

PR #2481, opened 2024-09-16
Extend the auto close time period from 1 month to 3 months

MeenaModhvadia:
## Self-review checklist

* [ ] (3) Tests passing

### DIFF
--- a/app/controllers/cases/offender_sar_controller.rb
+++ b/app/controllers/cases/offender_sar_controller.rb
@@ -46,7 +46,9 @@ module Cases
       @case.current_step = params[:current_step]
       get_rejected
       load_optional_flags_from_params
-
+      if @rejected == "true"
+        set_deadlines
+      end
       if steps_are_completed?
         if @case.valid_attributes?(create_params) && @case.valid?
           create_case
--- a/app/controllers/cases/offender_sar_controller.rb
+++ b/app/controllers/cases/offender_sar_controller.rb
@@ -46,9 +46,6 @@ module Cases
       @case.current_step = params[:current_step]
       get_rejected
       load_optional_flags_from_params
-      if @rejected == "true"
-        set_deadlines
-      end
       if steps_are_completed?
         if @case.valid_attributes?(create_params) && @case.valid?
           create_case
--- a/app/models/case/sar/offender.rb
+++ b/app/models/case/sar/offender.rb
@@ -559,6 +559,8 @@ private
 
   def set_deadlines
     super
-    self.external_deadline = @deadline_calculator.days_after(90, received_date)
+    if rejected?
+      self.external_deadline = @deadline_calculator.days_after(90, received_date)
+    end
   end
 end
--- a/app/models/case/sar/offender.rb
+++ b/app/models/case/sar/offender.rb
@@ -556,4 +556,9 @@ private
                     next_number
                   end
   end
+
+  def set_deadlines
+    super
+    self.external_deadline = @deadline_calculator.days_after(90, received_date)
+  end
 end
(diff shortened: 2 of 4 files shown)


app/controllers/cases/offender_sar_controller.rb
```diff
@@ -46,7 +46,9 @@ def create
       @case.current_step = params[:current_step]
       get_rejected
       load_optional_flags_from_params
-
+      if @rejected == "true"
+        set_deadlines
```
vertism, line 50: This method is already called on creation - https://github.com/ministryofjustice/correspondence_tool_staff/blob/ea807f31e2a372b15838eec958c8ac127fcf6633/app/models/case/base.rb#L416

MeenaModhvadia: amended

app/models/case/sar/offender.rb
```diff
@@ -556,4 +556,11 @@ def set_number
                     next_number
                   end
   end
+
+  def set_deadlines
+    super
+    if rejected?
+      self.external_deadline = @deadline_calculator.days_after(90, received_date)
```
vertism, line 563: Try to avoid [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)).

You could create a constant in the class for this value.

app/models/case/sar/offender.rb
```diff
@@ -64,6 +64,8 @@ def close_expired_rejected
         CaseClosureService.new(kase, User.system_admin, {}).call
       end
     end
+
+    AUTOCLOSUREDEADLINE = 90
```
vertism, line 68: We usually use underscores on constants for clarity.
Maybe the following?
```suggestion
    REJECTED_AUTO_CLOSURE_DEADLINE = 90
```

vertism: Also, it should be with the other constants in the class.

spec/models/case/sar/offender_spec.rb
```diff
@@ -76,6 +76,15 @@
     end
   end
 
+  describe "#external_deadline for rejected" do
+    let(:rejected_offender_sar_case) { create(:offender_sar_case, :rejected) }
+    let(:received_date) { Time.zone.today }
+
+    it "sets external deadline to 3 calendar months from received date" do
```
vertism, line 83: It sets it to 90 days which won't always be 3 months, so the test description should reflect that.

spec/models/case/sar/offender_spec.rb
```diff
@@ -76,6 +76,15 @@
     end
   end
 
+  describe "#external_deadline for rejected" do
+    let(:rejected_offender_sar_case) { create(:offender_sar_case, :rejected) }
+    let(:received_date) { Time.zone.today }
+
+    it "sets external deadline using a calculation of 90 days which is extended to next working day" do
```
vertism, line 83: In what way is it extended to the next working day?

MeenaModhvadia: It doesn't extend the date to the next working day therefore removed from description.

spec/models/case/sar/offender_spec.rb
```diff
@@ -76,6 +76,15 @@
     end
   end
 
+  describe "#external_deadline for rejected" do
+    let(:rejected_offender_sar_case) { create(:offender_sar_case, :rejected) }
+    let(:received_date) { Time.zone.today }
```
vertism, line 81: What's the purpose of `received_date`?

vertism (approved):
LGTM

PR #2493, opened 2024-10-15
Add missing event "assign_to_team_member"

vertism:
## Description
The event name was changed recently, but the old event name needs to be retained as it is used in multiple CaseTransitions in the database.


MeenaModhvadia (approved):
LGTM

PR #2510, opened 2024-11-08
Enable anon jobs

vertism:
## Description
Enables anon-jobs container on staging and production.
Also saves files locally to /tmp to avoid permissions issue.


lucas-shaw (approved):

njpreddy (approved):
LGTM

PR #2512, opened 2024-11-11
Fix db restore task

vertism:
## Description
- Uses /tmp directory as the user has permissions to write to it.
- Enables anon job in QA


njpreddy (approved):
LGTM

PR #2522, opened 2024-12-04
Allow for no original case

vertism:
## Description
Deleted ICO cases may not have been linked to an original case, which causes in error in the deleted cases report.
This stops an exception if there is no original case.


MeenaModhvadia (approved):
LGTM :)
My understanding is you are passing subject and name to the original case but if the original case is not found then the nil object error is gracefully ignored

PR #2532, opened 2024-12-19
CDPT-2377, CDPT-2378 Add alias and PNC to Standard commissioning document

vertism:
## Description
Adding new information to the standard commissioning document that gets sent when requesting data.


njpreddy (approved):
LGTM

PR #2534, opened 2025-01-08
CDPT-2390 CT: Hide Send Commissioning Email button for Offender SAR Complaints

MeenaModhvadia:
## Description
The functionality to send a commissioning email should not be available for the case type of  Offender SAR Complaints data requests.  The button 'send commissioning email' needs to be hidden.

## Self-review checklist

* [x] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-2390

Create offender sar and link to complaint types Standard, ICO & Litigation.


app/views/cases/data_request_areas/show.html.slim
```diff
@@ -44,7 +44,7 @@ div class="case"
             td
               span = @commissioning_document&.download_link
 
-      - if @data_request_area.data_request_emails.blank? && policy(@case).can_record_data_request?
+      - if @data_request_area.data_request_emails.blank? && policy(@case).can_record_data_request? && !@case.offender_sar_complaint?
```
vertism, line 47: Maybe we can replace `policy(@case).can_record_data_request?` with a policy method that relates to sending the email and that will differ depending in whether the case is a complaint or not.

vertism: There is an action of 'send_day_1_email' that is not enabled for complaints which is what causes the error, so that is what should be used to determine if sending is allowed.

app/policies/case/sar/offender_policy.rb
```diff
@@ -45,6 +45,11 @@ def can_record_data_request?
     check_can_trigger_event(:add_data_received)
   end
 
+  def can_send_day_1_email
```
vertism, line 48: All the other methods use a `?`, and you are using the question mark when you call the method.

spec/policies/case/sar/offender_policy_spec.rb
```diff
@@ -4,9 +4,15 @@
   subject { described_class }
 
   let(:user) { create :branston_user }
+  let(:offender_sar_case_b) { create(:offender_sar_case) }
```
vertism, line 7: Why has this got the `_b` suffix?

vertism (approved):
LGTM

PR #2552, opened 2025-01-31
Add cached num pages to warehouse report

lucas-shaw:
## Description
Capture the total number of pages for an Offender SAR case.

## Self-review checklist

* [ ] (3) Tests passing

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-2453

### DIFF
--- a/app/models/warehouse/case_report.rb
+++ b/app/models/warehouse/case_report.rb
@@ -285,7 +285,7 @@ module Warehouse
         case_report.third_party_company_name = kase.third_party_company_name
         case_report.number_of_exempt_pages = kase.number_exempt_pages
         case_report.number_of_final_pages = kase.number_final_pages
-        case_report.cached_num_pages = kase.page_count
+        case_report.pages_received = kase.page_count
         case_report.number_of_days_for_vetting = kase.number_of_days_for_vetting
         case_report.user_dealing_with_vetting = kase.user_dealing_with_vetting&.full_name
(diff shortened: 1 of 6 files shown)


app/models/warehouse/case_report.rb
```diff
@@ -281,10 +281,15 @@ def process_class_related_process(kase, case_report)
         end
       end
 
+      def total_cached_num_pages(kase)
+        kase.data_request_areas&.decorate&.sum(&:cached_num_pages) || 0
```
vertism, line 285: It doesn't seem right that this would require a decorator which suggests to me that the `cached_num_pages` method should be available on the object.

app/models/warehouse/case_report.rb
```diff
@@ -281,10 +281,15 @@ def process_class_related_process(kase, case_report)
...
+      end
+
       def process_offender_sar(kase, case_report)
         case_report.third_party_company_name = kase.third_party_company_name
         case_report.number_of_exempt_pages = kase.number_exempt_pages
         case_report.number_of_final_pages = kase.number_final_pages
+        case_report.cached_num_pages = total_cached_num_pages(kase)
```
vertism, line 292: "cached_num_pages" doesn't feel like the best name for this attribute. Maybe "pages_received" instead?

vertism:
I'm not sure if this will work as expected. We send an update to metabase when a case is updated, but I don't think changing the number of pages received for a data request will trigger an update.

PR #2553, opened 2025-01-31
CDPT-508 Display schedule and sent emails

vertism:
## Description
- Display info about next chase
- Add translations for email history table
- Display sent emails in reverse chronological order


njpreddy (approved):
LGTM

PR #2554, opened 2025-02-03
CDPT-2132 Add to case history when chase sent

vertism:
## Description
Adds case history message when automated chase email is sent.

Also adds method for checking if the chase is due today.


MeenaModhvadia (approved):
LGTM

PR #2557, opened 2025-02-05
Add area to chase case history message

vertism:
## Description
Distinguish case history messages for chase emails by adding data request area to the message.


MeenaModhvadia (approved):
LGTM

PR #2559, opened 2025-02-06
CDPT-2132 Fix case history message

vertism:
## Description
An email should always have been sent at this point. Re-use the message from the commissioning email.


MeenaModhvadia (approved):
LGTM

PR #2565, opened 2025-02-21
Check next date is present

vertism:
## Description
Method was not handling case where `next_chase_date` was nil.


njpreddy (approved):
LGTM

PR #2566, opened 2025-02-21
CDPT-2340 Remove empty scroll bar gutter from tables.

njpreddy:
## Description
If the width of tabulated data exceeds the width of the page the table becomes scrollable.  Currently empty scroll bar gutters appear when the table fits the page width.  To stop this behaviour the CSS overflow attribute should be changed to 'auto'.   

### Screenshots
Before:

before

After:

after

### Related JIRA tickets
https://dsdmoj.atlassian.net/browse/CDPT-2340


vertism (approved):
LGTM

PR #2568, opened 2025-02-24
CDPT-2513 Add chase number to email subject

vertism:
## Description
To help with auto-categorising of emails by the receiver, the chase number is being added to the email subject.


HettieS (approved):
LGTM

PR #2572, opened 2025-02-28
CDPT-2506 Case history pagination

vertism:
## Description
Recent additions have contributed to some cases having a large number of history logs which. Adding pagination means that older entries will be hidden by default, but still available to be seen if desired.


njpreddy (approved):
LGTM

PR #2574, opened 2025-03-06
CDPT-2445 Fix stub method that was not working

vertism:
## Description
The existing method was never calling the block passed to it. This meant tests were passing which weren't doing anything.
The method now causes the `set_case` method used in lots of controllers to return the case specified.


MeenaModhvadia (approved):
LGTM

PR #2576, opened 2025-03-07
CDPT-2564 Fix ability to upload request files

vertism:
## Description
Updated policy to stop uploading of request files for Offender SARs.
Also made the upload button visible on the case page before any documents have been uploaded where it is allowed.


MeenaModhvadia (approved):
LGTM

PR #2577, opened 2025-03-07
CDPT-2536 Fix file upload

vertism:
## Description
File upload does not work when editing an ICO case. Removing the option from the edit page as there is now a button on the main case page.

Also fixed a JS bug that happens locally when creating an ICO case.


MeenaModhvadia (approved):
LGTM

PR #2580, opened 2025-03-12
CDPT-2543 Change security data request area

vertism:
## Description
Changes "Security" data request area to "DPS sensitive" and adds "G1" data request type.
Also makes the headers and labels for data requests consistent.


MeenaModhvadia (approved):
LGTM

PR #2581, opened 2025-03-18
Remove dps_security request type

vertism:
## Description
This type has never been used and can be removed.
Note: It is not possible to remove individual enum values from the database


njpreddy (approved):
LGTM

PR #2582, opened 2025-03-20
CDPT-1425 Delete old RPI files

vertism:
## Description
ID documents are stored when received via the request personal information API.
These are not required once the request has been dealt with and should be deleted.

The cronjob has been created to run in production weekly on a Sunday at 3pm.


njpreddy (approved):
LGTM

